PR #19, opened 2020-03-17
Add setup.py to make library installable via pip

### LINKED ISSUE #7: Make pip installable

mschmidt87:
This PR adds a setup script to the repository so that the library can be installed via `pip install .` .
This fixes #7 .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ setup(
     version='0.1',
     author='Jakob Jordan, Maximilian Schmidt',
     author_email='jakobjordan@posteo.de',
-    description=('Cartesian genetic programming in Python.'),
+    description=('Cartesian Genetic Programming in Python.'),
     license='GPLv3',
     keywords='genetic programming',
     url='https://github.com/jakobj/python-gp',
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,28 @@
+# encoding: utf8
+from setuptools import setup
+
+setup(
+    name='cgprog',
+    version='0.1',
+    author='Jakob Jordan, Maximilian Schmidt',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -0,0 +1,28 @@
+# encoding: utf8
+from setuptools import setup
+
+setup(
+    name='python-gp',
+    version='0.1',
+    author='Jakob Jordan, Maximilian Schmidt',
+    author_email='j.jordan@fz-juelich.de',
```
jakobj:
```suggestion
    author_email='jakobjordan@posteo.de',
```

setup.py
```diff
@@ -0,0 +1,28 @@
...
+    description=('Cartesian genetic programming in Python.'),
+    license='GPLv3',
+    keywords='genetic programming',
+    url='https://github.com/jakobj/python-gp',
+    python_requires='>=3.6, <4',
+    install_requires=['sympy', 'torch', 'numpy', ],
```
jakobj: shouldn't we read this from requirements.txt?

setup.py
```diff
@@ -0,0 +1,28 @@
+# encoding: utf8
+from setuptools import setup
+
+setup(
+    name='python-gp',
+    version='0.1',
+    author='Jakob Jordan, Maximilian Schmidt',
+    author_email='j.jordan@fz-juelich.de',
+    description=('Cartesian genetic programming in Python.'),
```
jakobj:
```suggestion
    description=('Cartesian Genetic Programming in Python.'),
```

setup.py
mschmidt87: requirements file is introduced in #23 , so let's merge #23 first and I'll modify here then.

Issue #9, opened 2020-03-11
Review and fix examples


jakobj:

jakobj:
it looks like that example does not use an evolutionary method (no selection of individuals by fitness) but just randomly mutates expressions and performs gradient descent on the constants. how about adapting this to an example that essentially extends the other one by additionally performing gradient descent to find values for leaf nodes?

jakobj:
#20 is now merged. are you working on adapting the other script or should i have a look?

jakobj:
i'll work on the other example script and include the relevant information (we can backprop through the graph, yeah!) in the README

mschmidt87:
Great, thanks. I think it would be great to have that example, I was just too busy with the other stuff.

jakobj:
the original example was fixed. regarding the other example: closing this issue in favor of #59

PR #20, opened 2020-03-17
Update, clean up and refactor example_evo_regression script

mschmidt87:
This PR updates the `example_evo_regression` script. This addresses #9 partially.

@jakobj : For the "hard" example in the script, the evolution notoriously gets stuck with a constant expression. I tried a bunch of different parameters but couldn't get anywhere. Any ideas?

### DIFF
--- a/examples/example_evo_regression.py
+++ b/examples/example_evo_regression.py
@@ -1,122 +1,183 @@
 import numpy as np
 import matplotlib.pyplot as plt
+import os
 import scipy.constants
 import sympy
-from sympy.printing.dot import dotprint
+from functools import partial
 import sys
 import time
 import torch
 
+from sympy.printing.dot import dotprint
+
 sys.path.insert(0, '../')
 import gp
 
+"""
+Example script showing how to use Cartesian Genetic Programming for
+a simple regression task.
+"""
 
-def objective(individual):
 
+def f_target_easy(x):  # target function
+    return x[:, 0] ** 2 + 2 * x[:, 0] * x[:, 1] + x[:, 1] ** 2
+
+
+def f_target_hard(x):  # target function
+    return 1. / (1. + 1. / x[:, 0] ** 2) + 1. / (1. + 1. / x[:, 1] ** 2)
+
+
+def objective(individual, target_function):
+    """Objective function of the regression task.
+
+    Parameters
+    ----------
+    individual : gp.CGPIndividual
+        Individual of the Cartesian Genetic Programming Framework.
+    target_function : Callable
+        Target function.
+
+    Returns
+    -------
+    gp.CGPIndividual
+        Modified individual with updated fitness value.
+    """
     if individual.fitness is not None:
         return individual
 
     n_function_evaluations = 1000
 
     graph = gp.CGPGraph(individual.genome)
-    f_graph = graph.compile_torch_class()
-
-    def f_target(x):  # target function
-        # return 2.7182 + x[:, 0] - x[:, 1]
-        # return (1. + x[:, 0]) * (1. + x[:, 1])
-        return 1. / (1. + x[:, 0]) + 1. / (1. + x[:, 1])
-        # return 1. / (1. + 1. / x[:, 0]) + 1. / (1. + 1. / x[:, 1])
-        # return 1. / (1. + 1. / x[:, 0] ** 2) + 1.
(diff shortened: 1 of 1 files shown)


examples/example_evo_regression.py
```diff
@@ -1,122 +1,183 @@
...
 import sympy
-from sympy.printing.dot import dotprint
+from functools import partial
 import sys
 import time
 import torch
 
+from sympy.printing.dot import dotprint
+
 sys.path.insert(0, '../')
```
jakobj: i think we can remove this if the package is pip installable

examples/example_evo_regression.py
```diff
@@ -1,122 +1,183 @@
...
-        # return 1. / (1. + 1. / x[:, 0] ** 2) + 1.
-        # return 1. / (1. + 1. / x[:, 0] ** 2) + 1. / (1. + 1. / x[:, 1] ** 2)  # solved, ~50k
-        # return 1. / (1. + 1. / x[:, 0] ** 4) + 1. / (1. + 1. / x[:, 1] ** 4)
-        # return 1. / (1. + 1. / x[:, 0] ** 2)
-        # return x[:, 0] ** 2 + 2 * x[:, 0] * x[:, 1] + x[:, 1] ** 2  # solved, ~3k
-
+    f_graph = graph.to_torch()
     x = torch.Tensor(n_function_evaluations, 2).uniform_(-5, 5)
-    x[torch.abs(x) < 1e-5] = 1.  # avoid infinities due to zero division
+    # x[torch.abs(x) < 1e-5] = 1.  # avoid infinities due to zero division
```
jakobj: remove if not necessary

examples/example_evo_regression.py
```diff
@@ -1,122 +1,183 @@
...
+    
+    Parameters
+    ----------
+    f_target : Callable
+        Target function
+
+    Returns
+    -------
+    dict
+        Dictionary containing the history of the evoluation
```
jakobj:
```suggestion
        Dictionary containing the history of the evolution
```

examples/example_evo_regression.py
```diff
@@ -1,122 +1,183 @@
...
-    fig = plt.figure(figsize=(width, width / scipy.constants.golden))
-    ax_fitness = fig.add_subplot(311)
-    ax_fitness.set_xlabel('Evolution step')
-    ax_fitness.set_ylabel('Fitness')
+    pop = gp.CGPPopulation(**params['population_params'],
+                           seed=params['seed'], genome_params=params['genome_params'])
 
-    ax_phenotype_distance = fig.add_subplot(312)
-    ax_genotype_distance = fig.add_subplot(313)
+    # create instance of EA
```
jakobj:
```suggestion
    # create instance of evolutionary algorithm
```

examples/example_evo_regression.py
```diff
@@ -1,122 +1,183 @@
...
-    # sympy.plot(sympy_expr[0])
-
-    # export computational graphs; create pdf with
-    # $ dot -Tpdf {filename} -O
-    with open('example_evo_regression-graph.dot', 'w') as f:
-        f.write(dotprint(sympy_expr[0]))
-    with open('example_evo_regression-graph_simplified.dot', 'w') as f:
-        f.write(dotprint(sympy_expr[0].simplify()))
+    obj = partial(objective, target_function=f_target)
+    # Perform evoluation
```
jakobj:
```suggestion
    # Perform evolution
```

examples/example_evo_regression.py
```diff
@@ -1,122 +1,183 @@
 import numpy as np
 import matplotlib.pyplot as plt
+import os
 import scipy.constants
 import sympy
-from sympy.printing.dot import dotprint
+from functools import partial
```
jakobj: i would `import functools` to not loose the namespace on `partial`, but maybe it's common enough that it's clear where `partial` comes from

examples/example_evo_regression.py
mschmidt87: I will remove all of these calls from the scripts, incl. the test suite, once we have merged #19 . Let's do this in a separate PR.

mschmidt87: Added a ticket for this: #26

examples/example_evo_regression.py
mschmidt87: done.

mschmidt87:
I addressed all comments.

PR #23, opened 2020-03-18
Add requirements file and update travis file

mschmidt87:


mschmidt87:
One test is failing at the moment: The parallel speedup test doesn't show the desired speed up. I think this will be hard to investigate and fix in Travis. Should we just remove this test? I find it a bit peculiar anyway. @jakobj

.travis.yml
```diff
@@ -14,8 +16,10 @@ before_install:
...
 install:
-  - pip install --upgrade matplotlib numpy scipy sympy
-  - pip install https://download.pytorch.org/whl/cpu/torch-1.0.1.post2-cp37-cp37m-linux_x86_64.whl
-  - pip install torchvision
+  - pip install -r requirements.txt
+  - pip install pytest pytest-cov coveralls
 script:
-- cd test && pytest
\ No newline at end of file
+  - cd test/ && pytest --cov=dicthash
```
jakobj: what does `--cov=dicthash` do?

.travis.yml
```diff
@@ -14,8 +16,10 @@ before_install:
   - source activate myenv
   - pip install --upgrade pip
 install:
-  - pip install --upgrade matplotlib numpy scipy sympy
-  - pip install https://download.pytorch.org/whl/cpu/torch-1.0.1.post2-cp37-cp37m-linux_x86_64.whl
```
jakobj: aren't we missing pytorch?

requirements.txt
```diff
@@ -0,0 +1,4 @@
+numpy
+sympy
+torch
```
jakobj: does this work now? no need to pick a specific package?

test/test_cgp_evolve.py
```diff
@@ -217,25 +217,34 @@ def test_speedup_parallel_evolve():
...
+                                population_params['n_breeding'],
+                                population_params['tournament_size'],
+                                n_processes=n_processes)
+
+        t0 = time.time()
+        gp.evolve(pop, objective_speedup_parallel_evolve, ea,
+                  population_params['max_generations'],
+                  population_params['min_fitness'])
+        T = time.time() - t0
+        print(T)
```
jakobj: pls remove prints

test/test_cgp_evolve.py
```diff
@@ -217,25 +217,34 @@ def test_speedup_parallel_evolve():
...
+        if n_processes == 1:
+            T_baseline = T
+            assert T == pytest.approx(n_calls_objective * 0.1, rel=0.25)
```
jakobj: i don't understand this assert, maybe add a comment?

requirements.txt
mschmidt87: yep that works. See pytorch.org

.travis.yml
mschmidt87: is included in the requirements file

test/test_cgp_evolve.py
mschmidt87: This assert tests whether the total time is roughly equal to number of calls to the objective x execution time of one objective call.

Same as before, just refactored to make it more readable.

test/test_cgp_evolve.py
mschmidt87: done.

.travis.yml
mschmidt87: removed the `dicthash`.

mschmidt87:
> thanks! tada could you also implement in this PR that setup.py reads the requirements.txt?
> 
the `setup.py` is only introduced in #19 . I can't make both PRs depend on each other. So let's this one first and then I'll modify #19 .

> regarding the speedup test, can we keep it but disable it by default, such that one needs to explicitly enable it when calling pytest?

Okay, marked the test as skip.

mschmidt87:
Addressed all comments.

test/test_cgp_evolve.py
jakobj: oh wow, the `0.1` comes from the `sleep(0.1)`, i didn't see that, ok

PR #24, opened 2020-03-18
Bugfix/torch constant output shape

### LINKED ISSUE #22: CPGGraph.to_torch() generated class does not return correct shape for constant functions
When a genome represents a constant function, e.g. `f(x)=1`, the torch class generated by CPGGraph does not return the correct shape. I created a test for this on the `bugfix/torch_constant_output_shape` branch: https://github.com/jakobj/python-gp/blob/5a50c07876adf9e8a263b4e10c0e01c5b582fb68/test/test_cgp_graph.py#L161

This can probably be fixed by modifying the `CGPConstantFloat.format_output_str_torch()` method.
https://github.com/jakobj/python-gp/blob/3bdff24be10ef503652b44b0be1606a639555627/gp/cgp_node.py#L191

mschmidt87:
This PR fixes #22 .
It changes the format of the output string representing the `ConstantFloat` node. It also adds a test to verify that the changes work.

### DIFF
--- a/gp/cgp_node.py
+++ b/gp/cgp_node.py
@@ -227,7 +227,7 @@ class CGPConstantFloat(CGPNode):
         self._output_str = '{}'.format(self._output)
 
     def format_output_str_torch(self, graph):
-        self._output_str = 'self._p{}'.format(self._idx)
+        self._output_str = 'torch.cat(x.shape[0] * [self._p{}])'.format(self._idx)
 
     def format_parameter_str(self):
         self._parameter_str = 'self._p{} = torch.nn.Parameter(torch.Tensor([{}]))\n'.format(self._idx, self._output)
--- a/gp/cgp_node.py
+++ b/gp/cgp_node.py
@@ -227,7 +227,7 @@ class CGPConstantFloat(CGPNode):
         self._output_str = '{}'.format(self._output)
 
     def format_output_str_torch(self, graph):
-        self._output_str = 'torch.cat(x.shape[0] * [self._p{}])'.format(self._idx)
+        self._output_str = 'self._p{}.expand(x.shape[0])'.format(self._idx)
 
     def format_parameter_str(self):
         self._parameter_str = 'self._p{} = torch.nn.Parameter(torch.Tensor([{}]))\n'.format(self._idx, self._output)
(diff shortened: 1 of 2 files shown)


gp/cgp_node.py
```diff
@@ -188,7 +188,7 @@ def format_output_str(self, graph):
         self._output_str = '{}'.format(self._output)
 
     def format_output_str_torch(self, graph):
-        self._output_str = 'self._p{}'.format(self._idx)
+        self._output_str = 'torch.cat(x.shape[0] * [self._p{}])'.format(self._idx)
```
jakobj: i don't like this use of `x`, it's not apparent here where it comes from. isn't it enough to have as output value 2d tensor that torch can resize as necessary, i.e., `format_parameter_str` should be adjusted, rather than this function?

mschmidt87: I don't know how to do that in the way you're implying.
If the parameter is a 2d tensor, this doesn't work in combination with an operation that uses the input.

Also, x is used in CPGInputNode in a similar manner, such that it is not clear where x comes from.

test/test_cgp_graph.py
```diff
@@ -147,6 +148,34 @@ def test_compile_torch_and_backprop():
...
+genomes[3].dna = [-1, None, None, 1, None, None, 1, None, None, 0, 1, 1, 0, 0, 1, -2, 1, None, -2, 3, None]
+
+
+@pytest.mark.parametrize("genome, batch_size", product(genomes, batch_sizes))
+def test_compile_torch_output_shape(genome, batch_size):
+    graph = gp.CGPGraph(genome)
+    c = graph.compile_torch_class()
+    x = torch.Tensor(batch_size, 1).normal_()
+    y = c(x)
+    # if isinstance(y, tuple):
```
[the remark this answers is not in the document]
mschmidt87: Done.

gp/cgp_node.py
jakobj: wow, this took me some time to process. i understand the problem now, but suggest a slightly simpler version (using `expand` instead of the list constructor)

jakobj:
```suggestion
        self._output_str = 'self._p{}.expand(x.shape[0])'.format(self._idx)
```

mschmidt87: Yep, that's better, I committed the change.

PR #34, opened 2020-03-20
Replace all formatted strings by f-strings

mschmidt87:
As the PR title says: it simply replaces all formatted strings by f-strings, which is the more modern and readable syntax.

Fixes #21 .

Issue #26, opened 2020-03-18
Remove explicit system path manipulation


mschmidt87:
Once #19 has been merged, we can remove the lines that add the location of the package to the system path explicitly from all example and test scripts.

jakobj:
afterwards E402 should be enabled again for flake8

mschmidt87:
and we should edit the travis file. It will no longer be necessary to `cd` into the test folder.

mschmidt87:
One question: I think there are two possible approaches:
1. Keep the test suite at the top-level of the repo and then just do  `import gp` everywhere. That implies that the library needs to be installed before testing.
2. Move `test/` into `gp/` and then do `from .. import gp` everywhere. This is what is done in `python-dicthash` for instance. We can still exclude the test suite from being installed in the `setup.py`.
What do you think, @jakobj ?

jakobj:
i would strongly be in favor of 1. since one can just install in dev mode so that one does not have to reinstall after every change (`pip install -e`, see, e.g., https://stackoverflow.com/questions/35064426/when-would-the-e-editable-option-be-useful-with-pip-install), and keeping the tests separately seems like common practice

PR #35, opened 2020-03-20
Code formatting

### LINKED ISSUE #32: Use code formatter
As discussed in #3 we should use a formatter (black) to ensure consistent code style

mschmidt87:
This PR formats all python files in the library, tests, and examples using the black formatter and flake8.

The line length is set to 99 characters.
Flake8 is configured to ignore three rules:
- E203: This is currently broken and complains about whitespaces in list indices although that is encouraged by PEP8.  https://www.flake8rules.com/rules/E203.html (https://github.com/psf/black/issues/315)
- W503: This is outdated. https://www.flake8rules.com/rules/W503.html
- E402: Currently deactivated, because we have the manual import of the package in the test files which will be removed in a separate PR (see ticket: #26 ). https://www.flake8rules.com/rules/E402.html

The PR also adds the checks to CI so that Travis automatically checks for this in future.

Note: You can use pre-commit hooks to make your life easier. Just add the config file like this to your local repo (I could also add this to this PR? @jakobj ):
```
repos:
-   repo: https://github.com/ambv/black
    rev: 19.3b0
    hooks:
    - id: black
      language_version: python3.8
      args: [--config=pyproject.toml]
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v1.2.3
    hooks:
    - id: flake8
      args: [--config=.flake8]
```
Here are the steps to install the pre-commit hooks:

    Install pre-commit: pip install pre-commit
    Define .pre-commit-config.yaml with the hooks you want to include. (see above)
    Execute pre-commit install to install git hooks in your .git/ directory.

Note: This PR is branched off from #34 , so #34 should be merged first.

Fixes #32


.flake8
```diff
@@ -0,0 +1,6 @@
+[flake8]
+ignore = E203, W503, E402
+max-line-length = 99
+max-complexity = 18
+select = B,C,E,F,W,T4,B9
+exclude = __init__.py
```
jakobj: why exclude `__init__.py`?

jakobj:
regarding pre-commit hooks: looks really cool, but i don't think we need the file in the repository, do we? it seems boiler plate black and flake8

.flake8
mschmidt87: > regarding pre-commit hooks: looks really cool, but i don't think we need the file in the repository, do we? it seems boiler plate black and flake8

Yes, agree.

mschmidt87: I guess because flake8 then always complains about unused imports.

mschmidt87:
> regarding pre-commit hooks: looks really cool, but i don't think we need the file in the repository, do we? it seems boiler plate black and flake8

Okay.

jakobj:
for some reason travis fails, i've restarted the jobs now

mschmidt87:
Travis passes now. The construction in travis is not elegant because we first `cd` into tests and then `cd` out of it after executing the testsuite, but I'll keep it like this for now, and we can revise this.
The reason why we need to `cd` into the test folder is the construction in the test scripts where we manually add the path to the library to the system path. This will be changed soon anyway (#26 ) and so we can tackle this in a separate PR.

PR #42, opened 2020-03-23
Move ConstantFloat class factory to separate file

### LINKED ISSUE #29: Remove `cgp_constant_float` factory
The factory seems pretty pointless (https://github.com/jakobj/python-gp/blob/master/gp/cgp_node.py#L236) and should be removed: it takes only three lines to define your own constant float node (and avoiding pickling issues):
```python
class CustomCGPConstantFloat(CGPConstantFloat):
    def __init__(self, idx, inputs):
      super().__init__(idx, inputs)
      self._output = val
```
isn't that short enough? @mschmidt87

jakobj:
closes #29

### DIFF
--- a/test/test_cgp_node.py
+++ b/test/test_cgp_node.py
@@ -111,12 +111,10 @@ def test_pow():
     assert abs((x[0] ** x[1]) - y[0]) < 1e-15
 
 
-def test_const_float():
+def test_constant_float():
     params = {"n_inputs": 2, "n_outputs": 1, "n_columns": 1, "n_rows": 1, "levels_back": 1}
 
-    val = 1.678
-
-    primitives = [gp.custom_cgp_constant_float(val)]
+    primitives = [gp.CGPConstantFloat]
     genome = gp.CGPGenome(
         params["n_inputs"],
         params["n_outputs"],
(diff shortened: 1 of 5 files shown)


test/test_cgp_node.py
```diff
@@ -71,12 +71,10 @@ def test_mul():
...
 
 
-def test_const_float():
+def test_constant_float():
     params = {"n_inputs": 2, "n_outputs": 1, "n_columns": 1, "n_rows": 1, "levels_back": 1}
 
-    val = 1.678
-
-    primitives = [gp.custom_cgp_constant_float(val)]
+    primitives = [gp.CGPConstantFloat]
```
mschmidt87: Minor: Add a small comment explaining that this constant node has value 1.

test/test_cgp_node.py
```diff
@@ -88,19 +86,7 @@ def test_const_float():
     genome.dna = [-1, None, -1, None, 0, None, -2, 2]
     graph = gp.CGPGraph(genome)
 
-    x = [None, None]
+    x = [5.0, 1.5]
```
mschmidt87: Why not keep using `None` here?

jakobj: ok

jakobj:
github <-> travis comm == bork, merging

PR #55, opened 2020-03-27
Maint/remove binary population

### LINKED ISSUE #5: Remove `BinaryPopulation` class?
What to do with this class? Remove? Rename?

### LINKED ISSUE #31: Remove "CGP/cgp" prefixes where possible
We are using a lot of unnecessary prefixes, e.g., "CGPGraph", "CGPInputNode" etc. which was i guess motivated by not only having a `CGPPopulation` but also the `BinaryPopulation` class. Since we decided to remove the latter we should also go ahead and remove these prefixes to make the code more readable and avoid typing

jakobj:
- removes `BinaryIndividual`/`BinaryPopulation`
- removes now obsolete `AbstractIndividual`/`AbstractPopulation`
- removes the `cgp_`, `CGP` prefix everywhere
- renames `Graph` -> `CartesianGraph` to reduce confusion between genotype and phenotype

closes #5
closes #31


examples/example_evo_regression.py
```diff
@@ -54,7 +54,7 @@ def objective(individual, target_function):
 
 
 def evolution(f_target):
-    """Execute CGP on a regression task for a given target function.
+    """Execute  on a regression task for a given target function.
```
mschmidt87, line 57: Too much substitution here?

jakobj: yup, thx

jakobj:
thanks for the comments, sed was a bit to eager ;)

actually i did rename `cgp_individual.py` to `individual.py` and didn't delete it, so git should be able to realize that it was just renamed, thereby keeping the history intact. my gitk manages to detect that:
```
------------------------------- gp/individual.py -------------------------------
similarity index 81%
rename from gp/cgp_individual.py
rename to gp/individual.py
index 0edf8e4..9ab80df 100644
``` 

doesn't yours?

PR #60, opened 2020-04-02
Provide an example using the caching decorator

### LINKED ISSUE #50: Include caching example
we should add an example of an expensive fitness evaluation, e.g., just using `sleep` where we demonstrate how to use the caching decorator

jakobj:
closes #50

### DIFF
--- a/examples/example_caching.py
+++ b/examples/example_caching.py
@@ -1,0 +1,76 @@
+import numpy as np
+
+import gp
+
+""" Example demonstrating the use of the caching decorator.
+
+Caches the results of fitness evaluations in a pickle file
+('example_caching_cache.pkl'). To illustrate its practical use,
+compare the runtime of this script when you first call it vs. the
+second time and when you comment out the decorator on
+`inner_objective`."""
+
+
+def f_target(x):
+    return x ** 2 + x + 1.0
+
+
+def objective(individual):
+    if individual.fitness is not None:
+        return individual
+
+    @gp.utils.disk_cache("example_caching_cache.pkl")
+    def inner_objective(expr):
+        """The caching decorator uses the function parameter to identify
+        identical function calls. As many different genotypes produce
+        the same simplified SymPy expression we can use such
+        expressions as an argument to the decorated function to skip
+        reevaluating functionally identical individuals. Note that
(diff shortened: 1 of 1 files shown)


examples/example_caching.py
```diff
@@ -0,0 +1,76 @@
+import numpy as np
+
+import gp
+
+""" Example demonstrating the use of the caching decorator.
+
+Caches the results of fitness evaluations in a pickle file
+('example_caching_cache.pkl'). To illustrate its practical use,
+compare the runtime of this script when you first call it vs. the
+second time and when you comment out the decorator on
```
mschmidt87, line 10: I actually don't understand how to do this? If I just run the script twice, the cache will be emptied and the runtime will be the same, right? If I import the main function into an ipython terminal, and run it twice, I also observe the same runtime for both calls. 
If even if I modify the script like so:
```python
if __name__ == "__main__":
    import time
    t0 = time.time()
    champion = evolution()
    print(f"Time taken: {time.time() - t0}")

    t0 = time.time()
    champion = evolution()
    print(f"Time taken: {time.time() - t0}")
```
I am getting:
```
[64/100] max fitness: -4.930380657631324e-34
Time taken: 7.7174553871154785
[64/100] max fitness: -4.930380657631324e-34
Time taken: 7.535767555236816
evolved function: [x_0**2 + x_0 + 1.0]
```
I do observe a signifcant slowdown when I comment out the decorator, though, so that seems to work.

jakobj:
we can't use the LRU cache since it's not persistent: it only caches within a single session. if use we use multiple processes, for example, this wouldn't work. also this provides a simple version of checkpointing: even if your script crashes/job is killed you don't loose information about the already evaluated functions.

examples/example_caching.py
jakobj: i might be misunderstanding the question, but the decorator creates a file on disk that is never deleted by the decorator, i.e., the cache is never emptied. so if you don't manually delete the file before starting the interpreter you won't observe a speedup when running twice in the same session. maybe we should print a warning when using an existing file since that might lead to confusion on the users side?

mschmidt87: Okay, understood. I think it the documentation of the example is fine.

PR #62, opened 2020-04-02
Add checks for input dimension in python function and pytorch class

jakobj:
Without these checks one can produce quite strange error messages:
```bash
  File "<string>", line 1, in _f
IndexError: list index out of range
```


jakobj:
thanks, fixed them now.

PR #63, opened 2020-04-02
Split requirements into necessary and optional

jakobj:
This includes making compilation to SymPy expressions or PyTorch classes only available if these packages are installed. Optional dependencies are declared in a separate file (optional-requirement.txt).

Unfortunately this involves rewriting the example without sympy, removing the nice printing of the expression to pdf. :( We could think about adding that again in the caching example (#60), which anyway relies on sympy.


setup.py
```diff
@@ -11,7 +18,7 @@
     keywords="genetic programming",
     url="https://github.com/jakobj/python-gp",
     python_requires=">=3.6, <4",
-    install_requires=["sympy", "torch", "numpy"],
+    install_requires=read_requirements(),
```
mschmidt87, line 21: You can specify the extra requirements in a separate entry, like so:

```python
    extras_require={
        'sympy': ["sympy"]
        'torch': ["torch"],
        'numpy': ["numpy"]
    },
```

and then one can install them via pip by saying:
`pip install python-gp[sympy, torch]` or locally
`pip install .[sympy, torch]`

(Note that if you're using zsh as your shell you need to escape the whole thing to make it work because square brackets are protected in zsh: e.g. `pip install '.[sympy]'`

gp/cartesian_graph.py
```diff
@@ -292,6 +295,22 @@ def to_sympy(self, simplify=True):
         List[sympy.core.Expr]
             List of sympy expressions.
         """
+        import sympy
```
mschmidt87, line 298: You seem to use two different strategies for `sympy` and `torch`: For `torch`, you try to import it and then define the function regardless of whether the import worked or not. For `sympy`, you import the package inside the function definition. Is either of these ways safe?

It seems that even if importing `torch` failed, the user can call the `to_torch()` method and receive an error? For `sympy`, the user can also call `to_sympy()` and receive an `ImportError` then?

I am wondering whether if might be better to set a global variable if the import failed and then replace the function definition in that case by a simple `NotImplementedError`? Something like:

```python
try:
    import torch  # noqa: F401
    TORCH_FOUND = True
except ModuleNotFoundError:
    TORCH_FOUND = False

 ...

def to_torch(self):
    if not TORCH_FOUND:
        raise NotImplementedError("No torch installation found. Method not supported. Consider installing the extra requirements via `pip install python-gp[torch]`")
    else:
         # Define function
```

setup.py
jakobj: thanks, that's a good idea. however, i think `pip install -r optional-requirements.txt` is even simpler and a separate file with a list of modules easier to maintain. what do you think?

gp/cartesian_graph.py
jakobj: you are right, this was born out of not understanding python imports :( i would prefer to only import the modules where needed. for sympy this works, but for torch it creates problems with the dynamic creation of torch classes -- for some reason i don't understand.
i will change everything to a global import with an exception at the respective points where the module would be used. thanks for catching this! :+1:

setup.py
mschmidt87: The nice thing about what I suggested is that the user can switch off the extra requirements in a targeted manner. For instance, you might want torch but not sympy.

jakobj: :+1:

jakobj:
implemented your suggestion @mschmidt87 regarding extra requirements and fixed merge conflicts, please check again

Issue #37, opened 2020-03-23
Check whether silent mutations in nodes inputs are allowed


jakobj:
In principle each node has the same number of input genes, independent of its currently selected operation, e.g., a node with arity=1 still has 2 input genes if the max arity in the graph is 2. Does the current implementation allow the mutation of the silent input genes or are these excluded? Since silent mutations seem to be beneficial generally, we should make sure this is the case.

mschmidt87:
Currently, the only node with arity=1 is the OutputNode. Its silent input is the second entry which is always `None`, as far as I can see. Thus, it seems that this cannot be mutated? However, I don't quite understand the `cartesian_gaph._determine_active_nodes()` function which appears to be crucial for mutations. Can you explain this a bit more and ideally this should be written in a docstring of the function.

jakobj:
Yes, you are right. Unused genes are currently not mutated, there are checks in `mutate` such as `if self._dna[gene_idx] is not self._non_coding_allele` that prevent this. I would suggest to remove these checks and also allow mutations in dna parts which are currently non-coding. maybe this would allow us to remove the `self._non_coding_allele` completely. @mschmidt87 what do you think?

also, i've improved the implementation of `_determine_active_nodes` this should be more clear now (#82)

mschmidt87:
As far as I understand, the argument in favour of silent mutations is that these mutations can silently accumulate and then come into effect in a later generation, leading to higher exploration.
This should also hold for the unused input genes of nodes, thus it would make sense to allow that, in my opinion.

Issue #33, opened 2020-03-20
Add abstract class for algorithms


mschmidt87:
In order to simplify extending the library by more evolutionary algorithms, we should create a base class for algorithms that particular algorithms inherit from.

mschmidt87:
Do we still want to do this, @jakobj ?

jakobj:
let's add it only when the need arises. #KISS

PR #64, opened 2020-04-03
Add type checks

mschmidt87:
This PR adds type checks to the library and implements changes requested by the static type checker `mypy`. Eventually, it will also add `mypy` to the CI.

Note: This is a draft PR and was created now to raise awareness of this. To run `mypy`, execute:

```
pip install mypy
cd gp/
mypy *.py
```

This will currently raise a bunch of errors that need to be addressed before this PR can get merged. Also, the current state of the code introduces a circular import which will make importing the library fail.

### DIFF
--- a/gp/cartesian_graph.py
+++ b/gp/cartesian_graph.py
@@ -4,16 +4,23 @@ import re
 
 try:
     import sympy
+    from sympy.core import expr as sympy_expr
+
+    sympy_available = True
 except ModuleNotFoundError:
-    sympy = None
+    sympy_available = False
 
 try:
     import torch  # noqa: F401
+
+    torch_available = True
 except ModuleNotFoundError:
-    torch = None
+    torch_available = False
 
+from typing import Callable, Dict, List, Optional, Set
 
-from .node import InputNode, OutputNode, Parameter
+from .node import Node, InputNode, OutputNode, Parameter
+from .genome import Genome
 
 
 class CartesianGraph:
@@ -21,7 +28,7 @@ class CartesianGraph:
     Genome.
     """
 
-    def __init__(self, genome):
+    def __init__(self, genome: Genome) -> None:
         """Init function.
 
         Parameters
@@ -29,35 +36,34 @@ class CartesianGraph:
         genome: Genome
             Genome defining graph connectivity and node operations.
         """
-        self._n_outputs = None
-        self._n_inputs = None
-        self._n_columns = None
-        self._n_rows = None
-        self._nodes = None
-        self._gnome = None
+        self._n_outputs: int
+        self._n_inputs: int
+        self._n_columns: int
+        self._n_rows: int
+        self._nodes: List
 
         self.parse_genome(genome)
         self._genome = genome
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return "CartesianGraph(" + str(self._nodes) + ")"
 
-    def print_active_nodes(self):
+    def print_active_nodes(self) -> str:
         """Print a representation of all active nodes in the graph.
         """
         return "CartesianGraph(" + str([node for node in self._nodes if node._active]) + ")"
 
-    def pretty_str(self):
+    def pretty_str(self) -> str:
         """Print a pretty representation of the Cartesian graph.
         """
         n_characters = 24
 
-        def pretty_node_str(node):
+        def pretty_node_str(node: Node) -> str:
             s = node.pretty_str(n_characters)
             assert len(s) == n_characters
             return s
 
-        def empty_node_str():
+        def empty_node_str() -> str:
             return " " * n_characters
 
         s = "\n"
@@ -89,7 +95,7 @@ class CartesianGraph:
 
         return s
 
-    def parse_genome(self, genome):
+    def parse_genome(self, genome: Genome) -> None:
         if genome.dna is None:
             raise RuntimeError("dna not initialized")
 
@@ -117,22 +123,22 @@ class CartesianGraph:
 
         self._determine_active_nodes()
 
-    def _hidden_column_idx(self, idx):
+    def _hidden_column_idx(self, idx: int) -> int:
         return (idx - self._n_inputs) // self._n_rows
 
     @property
-    def input_nodes(self):
+    def input_nodes(self) -> List[Node]:
         return self._nodes[: self._n_inputs]
 
     @property
-    def hidden_nodes(self):
+    def hidden_nodes(self) -> List[Node]:
         return self._nodes[self._n_inputs : -self._n_outputs]
 
     @property
-    def output_nodes(self):
+    def output_nodes(self) -> List[Node]:
         return self._nodes[-self._n_outputs :]
 
-    def _determine_active_nodes(self):
+    def _determine_active_nodes(self) -> Dict[int, Set[Node]]:
         """Determine the active nodes in the graph.
 
         Starting from the output nodes, we work backward through the
@@ -145,7 +151,7 @@ class CartesianGraph:
 
         Returns
         -------
-        Dict[Set]
+        Dict[int, Set[Node]]
             Dictionary mapping colum indices to sets of active nodes.
 
         """
@@ -167,7 +173,7 @@ class CartesianGraph:
 
         return active_nodes_by_hidden_column_idx
 
-    def determine_active_regions(self):
+    def determine_active_regions(self) -> List[int]:
         """Determine the active regions in the computational graph.
 
         Returns
@@ -183,7 +189,7 @@ class CartesianGraph:
 
         return active_regions
 
-    def __call__(self, x):
+    def __call__(self, x: List[float]) -> List[float]:
         # store values of x in input nodes
         for i, xi in enumerate(x):
             assert isinstance(self._nodes[i], InputNode)
@@ -197,16 +203,16 @@ class CartesianGraph:
 
         return [node._output for node in self.output_nodes]
 
-    def __getitem__(self, key):
+    def __getitem__(self, key: int) -> Node:
         return self._nodes[key]
 
-    def to_str(self):
+    def to_str(self) -> str:
 
         self._format_output_str_of_all_nodes()
         out_str = ", ".join(node.output_str for node in self.output_nodes)
         return f"[{out_str}]"
 
-    def _format_output_str_of_all_nodes(self):
+    def _format_output_str_of_all_nodes(self) -> None:
 
         for i, node in enumerate(self.input_nodes):
             node.format_output_str(self)
@@ -216,17 +222,23 @@ class CartesianGraph:
             for node in active_nodes[hidden_column_idx]:
                 node.format_output_str(self)
 
-    def _fill_parameter_values(self, func_str, parameter_names_to_values):
+    def _fill_parameter_values(
+        self, func_str: str, parameter_names_to_values: Optional[Dict[str, float]] = None
+    ) -> str:
         g = re.findall("<p[0-9]+>", func_str)
-        if g and parameter_names_to_values is None:
-            raise ValueError("parameter node found but no value dict provided")
-        for parameter_name in g:
-            func_str = func_str.replace(
-                parameter_name, str(parameter_names_to_values[parameter_name])
-            )
+        if len(g) != 0:
+            if parameter_names_to_values is None:
+                raise ValueError("parameter node found but no value dict provided")
+            elif parameter_names_to_values is not None:
+                for parameter_name in g:
+                    func_str = func_str.replace(
+                        parameter_name, str(parameter_names_to_values[parameter_name])
+                    )
         return func_str
 
-    def to_func(self, parameter_names_to_values=None):
+    def to_func(
+        self, parameter_names_to_values: Optional[Dict[str, float]] = None
+    ) -> Callable[[List[float]], List[float]]:
         """Compile the function(s) represented by the graph.
 
         Generates a definition of the function in Python code and
@@ -259,7 +271,9 @@ def _f(x):
             for node in active_nodes[hidden_column_idx]:
                 node.format_output_str_numpy(self)
 
-    def to_numpy(self, parameter_names_to_values=None):
+    def to_numpy(
+        self, parameter_names_to_values: Optional[Dict[str, float]] = None
+    ) -> Callable[[np.ndarray], np.ndarray]:
         """Compile the function(s) represented by the graph to NumPy
         expression(s).
 
@@ -293,7 +307,9 @@ def _f(x):
 
         return locals()["_f"]
 
-    def to_torch(self, parameter_names_to_values=None):
+    def to_torch(
+        self, parameter_names_to_values: Optional[Dict[str, float]] = None
+    ) -> torch.nn.Module:
         """Compile the function(s) represented by the graph to a Torch class.
 
         Generates a definition of the Torch class in Python code and
@@ -304,7 +320,7 @@ def _f(x):
         torch.nn.Module
             Instance of the PyTorch class.
         """
-        if torch is None:
+        if not torch_available:
             raise ModuleNotFoundError("No module named 'torch' (extra requirement)")
 
         for i, node in enumerate(self.input_nodes):
@@ -350,7 +366,11 @@ class _C(torch.nn.Module):
 
         return locals()["_c"]
 
-    def to_sympy(self, simplify=True, parameter_names_to_values=None):
+    def to_sympy(
+        self,
+        simplify: Optional[bool] = True,
+        parameter_names_to_values: Optional[Dict[str, float]] = None,
+    ) -> List[sympy_expr.Expr]:
         """Compile the function(s) represented by the graph to a SymPy expression.
 
         Generates one SymPy expression for each output node.
@@ -363,13 +383,13 @@ class _C(torch.nn.Module):
 
         Returns
         ----------
-        List[sympy.core.Expr]
+        List[sympy.core.expr.Expr]
             List of SymPy expressions.
         """
-        if sympy is None:
+        if not sympy_available:
             raise ModuleNotFoundError("No module named 'sympy' (extra requirement)")
 
-        def _validate_sympy_expr(expr):
+        def _validate_sympy_expr(expr: sympy_expr.Expr) -> sympy_expr.Expr:
             """Helper function that raises an exception upon encountering a SymPy
             expression that can not be evaluated.
 
--- a/gp/genome.py
+++ b/gp/genome.py
@@ -1,11 +1,29 @@
+import numpy as np
+
+from typing import Generator, List, Optional, Tuple, Type
+
+from .node import Node
 from .primitives import Primitives
+
+
+ID_INPUT_NODE: int = -1
+ID_OUTPUT_NODE: int = -2
+ID_NON_CODING_GENE: int = -3
 
 
 class Genome:
     """Genome class for  individuals.
     """
 
-    def __init__(self, n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives):
+    def __init__(
+        self,
+        n_inputs: int,
+        n_outputs: int,
+        n_columns: int,
+        n_rows: int,
+        levels_back: int,
+        primitives: List[Type[Node]],
+    ) -> None:
         """Init function.
 
         Parameters
@@ -21,7 +39,7 @@ class Genome:
         levels_back : int
             Number of previous columns that an entry in the genome can be
             connected with.
-        primitives : List[gp.CPGNode]
+        primitives : List[Type[Node]]
            List of primitives that the genome can refer to.
         """
         if n_inputs <= 0:
@@ -53,45 +71,46 @@ class Genome:
             1 + self._primitives.max_arity
         )  # one function gene + multiple input genes
 
-        self._dna = None  # stores dna as list of alleles for all regions
+        self._dna: List[int] = []  # stores dna as list of alleles for all regions
 
         # constants used as identifiers for input and output nodes
-        self._id_input_node = -1
-        self._id_output_node = -2
+        self._id_input_node: int = ID_INPUT_NODE
+        self._id_output_node: int = ID_OUTPUT_NODE
+        self._id_unused_gene: int = ID_NON_CODING_GENE
 
-    def __getitem__(self, key):
+    def __getitem__(self, key: int) -> int:
         if self._dna is None:
             raise RuntimeError("dna not initialized")
         return self._dna[key]
 
-    def __setitem__(self, key, value):
+    def __setitem__(self, key: int, value: int) -> None:
         dna = list(self._dna)
         dna[key] = value
         self._validate_dna(dna)
         self._dna = dna
 
     @property
-    def dna(self):
+    def dna(self) -> List[int]:
         return self._dna
 
     @dna.setter
-    def dna(self, value):
+    def dna(self, value: List[int]) -> None:
         self._validate_dna(value)
         self._dna = value
 
     @property
-    def _n_hidden(self):
+    def _n_hidden(self) -> int:
         return self._n_columns * self._n_rows
 
     @property
-    def _n_regions(self):
+    def _n_regions(self) -> int:
         return self._n_inputs + self._n_hidden + self._n_outputs
 
     @property
-    def _n_genes(self):
+    def _n_genes(self) -> int:
         return self._n_regions * self._length_per_region
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         s = self.__class__.__name__ + "("
         for region_idx, input_region in self.iter_input_regions():
             s += str(region_idx) + ": " + str(input_region) + " | "
@@ -103,16 +122,18 @@ class Genome:
         s += ")"
         return s
 
-    def _create_input_region(self):
+    def _create_input_region(self) -> List[int]:
 
         region = []
         region.append(self._id_input_node)
-        # empty input genes (None) since input nodes do not receive
+        # non-coding input genes since input nodes do not receive
         # inputs themselves
-        region += [None] * self._primitives.max_arity
+        region += [self._id_unused_gene] * self._primitives.max_arity
         return region
 
-    def _create_random_hidden_region(self, rng, permissable_inputs):
+    def _create_random_hidden_region(
+        self, rng: np.random.RandomState, permissable_inputs: List[int]
+    ) -> List[int]:
 
         region = []
         node_id = self._primitives.sample(rng)
@@ -121,18 +142,19 @@ class Genome:
 
         return region
 
-    def _create_random_output_region(self, rng, permissable_inputs):
+    def _create_random_output_region(
+        self, rng: np.random.RandomState, permissable_inputs: List[int]
+    ) -> List[int]:
 
         region = []
         region.append(self._id_output_node)
         region.append(rng.choice(permissable_inputs))
-        # output nodes have only on input, other genes are hence left
-        # empty
-        region += [None] * (self._primitives.max_arity - 1)
+        # output nodes have only one input, other genes are hence non-coding
+        region += [self._id_unused_gene] * (self._primitives.max_arity - 1)
 
         return region
 
-    def randomize(self, rng):
+    def randomize(self, rng: np.random.RandomState) -> None:
         """Randomize the genome.
 
         Parameters
@@ -144,7 +166,7 @@ class Genome:
         ----------
         None
         """
-        dna = []
+        dna: List[int] = []
 
         # add input nodes
         for i in range(self._n_inputs):
@@ -167,7 +189,7 @@ class Genome:
         self._validate_dna(dna)
         self._dna = dna
 
-    def _permissable_inputs(self, region_idx):
+    def _permissable_inputs(self, region_idx: int) -> List[int]:
 
         assert not self._is_input_region(region_idx)
 
@@ -190,10 +212,10 @@ class Genome:
 
         return permissable_inputs
 
-    def _permissable_inputs_for_output_region(self):
+    def _permissable_inputs_for_output_region(self) -> List[int]:
         return self._permissable_inputs(self._n_inputs + self._n_rows * self._n_columns)
 
-    def _validate_dna(self, dna):
+    def _validate_dna(self, dna: List[int]) -> None:
 
         if len(dna) != self._n_genes:
             raise ValueError("dna length mismatch")
@@ -205,7 +227,7 @@ class Genome:
                     "function genes for input nodes need to be identical to input node identifiers"
                 )
 
-            if input_region[1:] != ([None] * self._primitives.max_arity):
+            if input_region[1:] != ([self._id_unused_gene] * self._primitives.max_arity):
                 raise ValueError("input genes for input nodes need to be empty")
 
         for region_idx, hidden_region in self.iter_hidden_regions(dna):
@@ -230,10 +252,10 @@ class Genome:
             if output_region[1] not in self._permissable_inputs_for_output_region():
                 raise ValueError("input gene for output nodes has invalid value")
 
-            if output_region[2:] != [None] * (self._primitives.max_arity - 1):
+            if output_region[2:] != [self._id_unused_gene] * (self._primitives.max_arity - 1):
                 raise ValueError("inactive input genes for output nodes need to be empty")
 
-    def _hidden_column_idx(self, region_idx):
+    def _hidden_column_idx(self, region_idx: int) -> int:
         assert self._n_inputs <= region_idx
         assert region_idx < self._n_inputs + self._n_rows * self._n_columns
         hidden_column_idx = (region_idx - self._n_inputs) // self._n_rows
@@ -241,7 +263,9 @@ class Genome:
         assert hidden_column_idx < self._n_columns
         return hidden_column_idx
 
-    def iter_input_regions(self, dna=None):
+    def iter_input_regions(
+        self, dna: Optional[List[int]] = None
+    ) -> Generator[Tuple[int, list], None, None]:
         if dna is None:
             dna = self.dna
         for i in range(self._n_inputs):
@@ -251,7 +275,9 @@ class Genome:
             ]
             yield region_idx, region
 
-    def iter_hidden_regions(self, dna=None):
+    def iter_hidden_regions(
+        self, dna: Optional[List[int]] = None
+    ) -> Generator[Tuple[int, List[int]], None, None]:
         if dna is None:
             dna = self.dna
         for i in range(self._n_hidden):
@@ -261,7 +287,9 @@ class Genome:
             ]
             yield region_idx, region
 
-    def iter_output_regions(self, dna=None):
+    def iter_output_regions(
+        self, dna: Optional[List[int]] = None
+    ) -> Generator[Tuple[int, List[int]], None, None]:
         if dna is None:
             dna = self.dna
         for i in range(self._n_outputs):
@@ -271,30 +299,30 @@ class Genome:
             ]
             yield region_idx, region
 
-    def _is_gene_in_input_region(self, gene_idx):
+    def _is_gene_in_input_region(self, gene_idx: int) -> bool:
         return gene_idx < (self._n_inputs * self._length_per_region)
 
-    def _is_gene_in_hidden_region(self, gene_idx):
+    def _is_gene_in_hidden_region(self, gene_idx: int) -> bool:
         return ((self._n_inputs * self._length_per_region) <= gene_idx) & (
             gene_idx < ((self._n_inputs + self._n_hidden) * self._length_per_region)
         )
 
-    def _is_gene_in_output_region(self, gene_idx):
+    def _is_gene_in_output_region(self, gene_idx: int) -> bool:
         return ((self._n_inputs + self._n_hidden) * self._length_per_region) <= gene_idx
 
-    def _is_input_region(self, region_idx):
+    def _is_input_region(self, region_idx: int) -> bool:
         return region_idx < self._n_inputs
 
-    def _is_hidden_region(self, region_idx):
+    def _is_hidden_region(self, region_idx: int) -> bool:
         return (self._n_inputs <= region_idx) & (region_idx < self._n_inputs + self._n_hidden)
 
-    def _is_output_region(self, region_idx):
+    def _is_output_region(self, region_idx: int) -> bool:
         return self._n_inputs + self._n_hidden <= region_idx
 
-    def _is_function_gene(self, gene_idx):
+    def _is_function_gene(self, gene_idx: int) -> bool:
         return (gene_idx % self._length_per_region) == 0
 
-    def _is_active_input_gene(self, gene_idx):
+    def _is_active_input_gene(self, gene_idx: int) -> bool:
         input_index = gene_idx % self._length_per_region
         assert input_index > 0
         region_idx = gene_idx // self._length_per_region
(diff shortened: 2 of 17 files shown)


gp/cartesian_graph.py
```diff
@@ -1,53 +1,53 @@
 import collections
+import gp
```
jakobj, line 2: should not use global import within the package

gp/cartesian_graph.py
```diff
@@ -1,53 +1,53 @@
...
         genome: Genome
             Genome defining graph connectivity and node operations.
         """
-        self._n_outputs = None
-        self._n_inputs = None
-        self._n_columns = None
-        self._n_rows = None
-        self._nodes = None
-        self._gnome = None
+        self._n_outputs: int = 0
```
jakobj, line 23: is there a `nan` type for ints that could signal uninitialized values?

gp/cartesian_graph.py
```diff
@@ -105,22 +105,22 @@ def parse_genome(self, genome):
...
+        return self._nodes[self._n_inputs:-self._n_outputs]
 
     @property
-    def output_nodes(self):
-        return self._nodes[-self._n_outputs :]
+    def output_nodes(self) -> List[gp.Node]:
+        return self._nodes[-self._n_outputs:]
 
-    def _determine_active_nodes(self):
+    def _determine_active_nodes(self) -> dict:
```
jakobj, line 123: is it possible to indicate dict of sets like `Dict[Set]`?

gp/genome.py
```diff
@@ -1,11 +1,23 @@
-from .primitives import Primitives
+import gp
```
jakobj, line 1: no absolute imports

gp/genome.py
```diff
@@ -1,11 +1,23 @@
...
 
-    def __init__(self, n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives):
+    def __init__(
+        self,
+        n_inputs: int,
+        n_outputs: int,
+        n_columns: int,
+        n_rows: int,
+        levels_back: int,
+        primitives: List[gp.CPGNode],
```
jakobj, line 19:
```suggestion
        primitives: List[gp.Node],
```

gp/genome.py
```diff
@@ -53,46 +65,46 @@ def __init__(self, n_inputs, n_outputs, n_columns, n_rows, levels_back, primitiv
...
         )  # one function gene + multiple input genes
 
-        self._dna = None  # stores dna as list of alleles for all regions
+        self._dna: List[int] = []  # stores dna as list of alleles for all regions
 
         # constants used as identifiers for input and output nodes
         self._id_input_node = -1
         self._id_output_node = -2
-        self._non_coding_allele = None
+        self._non_coding_allele: int = 0
```
jakobj, line 73: this is not good as `0` is a valid gene. maybe we can remove this "non-coding allele" completely as suggested in #37

gp/genome.py
```diff
@@ -257,15 +273,17 @@ def _validate_dna(self, dna):
...
         assert region_idx < self._n_inputs + self._n_rows * self._n_columns
         hidden_column_idx = (region_idx - self._n_inputs) // self._n_rows
         assert 0 <= hidden_column_idx
         assert hidden_column_idx < self._n_columns
         return hidden_column_idx
 
-    def iter_input_regions(self, dna=None):
+    def iter_input_regions(
+        self, dna: Optional[List[int]] = None
+    ) -> Generator[Tuple[int, list], None, None]:
```
jakobj, line 286: what do the last `None, None` indicate?

gp/hl_api.py
```diff
@@ -1,18 +1,21 @@
 import numpy as np
+from typing import Optional, Callable
+from .abstract_individual import AbstractIndividual
```
jakobj, line 3: i think these `Abstract...` classes are outdated

gp/hl_api.py
```diff
@@ -1,18 +1,21 @@
...
-    max_generations,
-    min_fitness,
-    print_progress=False,
-    *,
-    callback=None,
-    label=None,
-    n_processes=1,
-):
+    pop: AbstractPopulation,
+    objective: Callable[[AbstractIndividual], AbstractIndividual],
```
jakobj, line 10: shouldn't the objective just take one individual?

gp/hl_api.py
```diff
@@ -1,18 +1,21 @@
...
+    ea: MuPlusLambda,
```
jakobj, line 11: now this would be a reason to create an abstract ea class ;)

gp/hl_api.py
```diff
@@ -1,18 +1,21 @@
...
+    max_generations: int,
+    min_fitness: float,
+    print_progress: Optional[bool] = False,
+    callback: Optional[Callable[[AbstractPopulation], None]] = None,
```
jakobj, line 15: callback just receives the population

gp/node.py
```diff
@@ -1,7 +1,13 @@
+from __future__ import annotations
+
+from typing import List, Union
+
+from cartesian_graph import CartesianGraph
```
jakobj, line 5:
```suggestion
from .cartesian_graph import CartesianGraph
```

gp/node.py
```diff
@@ -22,14 +28,17 @@ class Node:
...
 
-    _arity = None
-    _active = False
-    _inputs = None
-    _output = None
-    _idx = None
-    _is_parameter = False
-
-    def __init__(self, idx, inputs):
+    __name__ = "Node"
```
jakobj, line 31: why do we need this?

gp/node.py
```diff
@@ -22,14 +28,17 @@ class Node:
...
+    _arity: int = 0
```
jakobj, line 32: again, this is suboptimal since the intention of setting this to `None` is to cause failures if it's not set. if it's set to `0` it could just generate nonsense results

gp/node.py
```diff
@@ -44,33 +53,36 @@ def __init__(self, idx, inputs):
 
         assert idx not in inputs
 
-    def __init_subclass__(cls, **kwargs):
-        super().__init_subclass__(**kwargs)
-        register(cls)
+    # def __init_subclass__(cls: Node, **kwargs: Any) -> None:
```
jakobj, line 56: should be commented in before merge

gp/population.py
```diff
@@ -1,14 +1,18 @@
+import gp
```
jakobj, line 1: no absolute import

gp/cartesian_graph.py
mschmidt87: The imports are a bit tricky with mypy and I haven't figured it out yet. Essentially, mypy tries to check all packages that are imported, but some of them don't have proper type annotations yet (e.g. numpy). 

In addition, I haven't figured out how to import the package's own files properly. I agree that we shouldn't have global imports, but with relative imports I am getting errors with mypy all the time. Let's discuss a little bit in person. There is also documentation of this in mypy, but to be honest, I don't understand their documentation or at least my conclusions seem to be wrong. This is their documentation: https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports

gp/cartesian_graph.py
mschmidt87: I now changed this and don't initalize the values at all.

gp/cartesian_graph.py
mschmidt87: Yep, changed this.

gp/genome.py
mschmidt87: see above

gp/genome.py
mschmidt87: fixed.

gp/genome.py
mschmidt87: Fixed this for now by not initializing this variable.

gp/genome.py
mschmidt87: The `SendType` and `ReturnType`. See https://docs.python.org/3/library/typing.html#typing.Generator

gp/hl_api.py
mschmidt87: yes, fixed.

gp/hl_api.py
mschmidt87: This defines the arguments and return value: https://docs.python.org/3/library/typing.html#typing.Callable

gp/hl_api.py
mschmidt87: Yes, that's what I've been saying in #33

gp/hl_api.py
mschmidt87: yes, the `None` defines the return type.

gp/node.py
mschmidt87: Don't remember now, but I added it because of some mypy complaint. Will look into it.

gp/node.py
mschmidt87: yes, fixed.

gp/node.py
mschmidt87: there was some problem with this construction, `mypy` complained about it. Will look into this to sort it out.

gp/population.py
mschmidt87: see above.

mschmidt87:
Thank you for your comments, I fixed most of them, the most important point and obstacle towards getting `mypy` passing is the problem with the imports, which is still mysterious to me.

gp/cartesian_graph.py
jakobj: my C++ alarms are ringing reading this statement ;) how does it work in python? to which value are they set in that case? like in C++ whatever resides at that memory location? defaults of the annotations module?

gp/genome.py
jakobj: see above regarding uninitialized values

gp/genome.py
jakobj: thanks!

gp/hl_api.py
jakobj: thanks!

gp/hl_api.py
jakobj: right :D ok, let's stick to our decision and keep it like it is for now

jakobj:
thanks for the update! let's discuss the remaining issues per video.

gp/cartesian_graph.py
jakobj: it's strange because in other parts we do use the relative import of (sub)modules, don't we? does mypy complain about those?

mschmidt87: Yes, when I use relative imports, mypy throws an error. Using absolute imports solves this, but it's not a good way, so I opened a ticket in their repo to ask about that: https://github.com/python/mypy/issues/8726

gp/cartesian_graph.py
mschmidt87: Yeah, maybe that is not a good thing, Python will throw a `NameError` when you try to access the member. The alternative would be to write it like this:
`self._n_outputs: Optional[int] = None` (i.e. to make the type Optional which means it can be None or an integer).

mschmidt87:
After getting a reply in the mypy-issue, I now fixed all the imports to be relative instead of absolute.
Note that with this PR, the library is currently not usable due to some cyclic imports. My plan is to first make mypy pass and then deal with that. We will likely need to introduce some stub type definitions to avoid the cyclic imports.

To run `mypy`, do the following:

```bash
cd gp
cd ..
mypy gp
```

I am currently getting these errors:

```
gp/genome.py:127: error: List item 0 has incompatible type "None"; expected "int"
gp/genome.py:128: error: Incompatible return value type (got "List[int]", expected "List[Optional[int]]")
gp/genome.py:128: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
gp/genome.py:128: note: Consider using "Sequence" instead, which is covariant
gp/genome.py:128: note: Perhaps you need a type annotation for "region"? Suggestion: "List[Optional[int]]"
gp/genome.py:150: error: List item 0 has incompatible type "None"; expected "int"
gp/genome.py:170: error: Argument 1 to "__iadd__" of "list" has incompatible type "List[Optional[int]]"; expected "Iterable[int]"
gp/genome.py:431: error: Argument 6 to "Genome" has incompatible type "Primitives"; expected "List[Type[Node]]"
gp/cartesian_graph.py:12: error: Incompatible types in assignment (expression has type "None", variable has type Module)
gp/cartesian_graph.py:109: error: Argument 2 to "__call__" of "Node" has incompatible type "List[int]"; expected "CartesianGraph"
gp/cartesian_graph.py:161: error: Argument 1 to "_hidden_column_idx" of "CartesianGraph" has incompatible type "Optional[int]"; expected "int"
Found 8 errors in 2 files (checked 12 source files)
```

There are two main problems that need to be resolved:
- This error: `gp/genome.py:431: error: Argument 6 to "Genome" has incompatible type "Primitives"; expected "List[Type[Node]]"` is because the `clone()` function of the `Genome` class puts its `_primitives` members as an input variable to its `__init__` function to generate the cloned class. However, these types are incompatible since `_primitives` is an instance of `Primitives`, but in the `__init__` functions, the instance of `Primitives` is generated from the input variable, thus this is somewhat circular. This should also fail at the moment, as far as I can see.

- mypy complains about the use of lists for indexing regions in the genome. The reason is that we use either integers of `None` and `mypy` demands lists to be invariant, i.e., to contain only one type. See this documentation for some explanation: https://mypy.readthedocs.io/en/latest/common_issues.html#invariance-vs-covariance
Thus, we can either use something else than `None` for non-used genes (e.g. -1) or we refrain from using lists and replace them by an immutable object like e.g. a `tuple`.

mschmidt87:
One more point:
The `__call__` method of the `Node` class has this signature:
https://github.com/Happy-Algorithms-League/python-gp/blob/c98cf32ca8b645c35ae6a6b74a370b19efef4456/gp/node.py#L168
```
def __call__(self, x, graph)
```
However, the `x` appears to be unused in all child class of `Node`. Can it be removed?
Also, this `__call__` method sets the `output_` member of the Node class and seems to be used only in the `CartesianGraph.parse_genome()` method:
https://github.com/Happy-Algorithms-League/python-gp/blob/c98cf32ca8b645c35ae6a6b74a370b19efef4456/gp/cartesian_graph.py#L108

and in the `__call__` method of the CartesianGraph.
The more relevant member for actually computing the function of the graph is `_output_str` if I'm not mistaken.

jakobj:
>     * This error: `gp/genome.py:431: error: Argument 6 to "Genome" has incompatible type "Primitives"; expected "List[Type[Node]]"` is because the `clone()` function of the `Genome` class puts its `_primitives` members as an input variable to its `__init__` function to generate the cloned class. However, these types are incompatible since `_primitives` is an instance of `Primitives`, but in the `__init__` functions, the instance of `Primitives` is generated from the input variable, thus this is somewhat circular. This should also fail at the moment, as far as I can see.

how about supporting the `list()` constructor so we can call `list(self._primitives)` in `Genome::clone`?

>     * mypy doesn't like the construction about checking the availability of torch and sympy. I think the best way to solve this is to rather use a `torch_found` boolean variable and check for this than to directly check `if torch`.

I find it less pretty than the current implementation that does not introduce an additional variable, but if mypy is unhappy otherwise, feel free to change it. i guess a comment would help there.

>     * mypy complains about the use of lists for indexing regions in the genome. The reason is that we use either integers of `None` and `mypy` demands lists to be invariant, i.e., to contain only one type. See this documentation for some explanation: https://mypy.readthedocs.io/en/latest/common_issues.html#invariance-vs-covariance
>       Thus, we can either use something else than `None` for non-used genes (e.g. -1) or we refrain from using lists and replace them by an immutable object like e.g. a `tuple`.

we can use -3 for unused genes in the `InpuNode/OutputNode` sections. we've already introduced -1 and -2 as special values so why not an additional one ;)

jakobj:
> One more point:
> The `__call__` method of the `Node` class has this signature:
> [...snip...]

unfortunately i don't remember why we need the `x` there. input values seem to be retrieved by accessing input nodes via the graph instance. in the beginning we provided the option of calling the graph directly without calling any `to...`method. i am now wondering how input values are passed into the graph if `x` is never used? we can either add tests to make sure this functionality works as expected, or we force the user to always call one of the `to...` functions. i am inclined to follow the latter path, just because it reduces complexity without any significant drawback (that i'm currently aware of).

note that `__call__` is _not_ used in https://github.com/Happy-Algorithms-League/python-gp/blob/c98cf32ca8b645c35ae6a6b74a370b19efef4456/gp/cartesian_graph.py#L108 as `primitives` does not store instances, but classes, so this calls the constructor, not the `__call__` function!

gp/cartesian_graph.py
```diff
@@ -4,60 +4,66 @@
...
         ----------
         genome: Genome
             Genome defining graph connectivity and node operations.
         """
-        self._n_outputs = None
-        self._n_inputs = None
-        self._n_columns = None
-        self._n_rows = None
-        self._nodes = None
-        self._gnome = None
```
jakobj, line 37: isn't the declaration of `_genome : Genome` missing here?

gp/cartesian_graph.py
```diff
@@ -216,17 +222,21 @@ def _format_output_str_of_all_nodes(self):
             for node in active_nodes[hidden_column_idx]:
                 node.format_output_str(self)
 
-    def _fill_parameter_values(self, func_str, parameter_names_to_values):
+    def _fill_parameter_values(
+        self, func_str: str, parameter_names_to_values: Optional[Dict[str, float]] = None
+    ) -> str:
         g = re.findall("<p[0-9]+>", func_str)
-        if g and parameter_names_to_values is None:
+        if len(g) != 0 and parameter_names_to_values is None:
```
jakobj, line 229: why can't we use `if g`?

gp/cartesian_graph.py
```diff
@@ -216,17 +222,21 @@ def _format_output_str_of_all_nodes(self):
...
             raise ValueError("parameter node found but no value dict provided")
-        for parameter_name in g:
-            func_str = func_str.replace(
-                parameter_name, str(parameter_names_to_values[parameter_name])
-            )
+        # explicit elif branch is necessary because mypy does not pass otherwise
```
jakobj, line 231: this is a very strange construction. what's wrong with this according to mypy?

gp/cartesian_graph.py
```diff
@@ -259,7 +269,7 @@ def _format_output_str_numpy_of_all_nodes(self):
             for node in active_nodes[hidden_column_idx]:
                 node.format_output_str_numpy(self)
 
-    def to_numpy(self, parameter_names_to_values=None):
+    def to_numpy(self, parameter_names_to_values: Optional[Dict[str, float]] = None):
```
jakobj, line 272: missing return value annotation?

gp/cartesian_graph.py
```diff
@@ -293,7 +303,9 @@ def _f(x):
 
         return locals()["_f"]
 
-    def to_torch(self, parameter_names_to_values=None):
+    def to_torch(
```
jakobj, line 306: does this return value annotation cause any problems if torch is not available?

gp/cartesian_graph.py
```diff
@@ -216,17 +222,21 @@ def _format_output_str_of_all_nodes(self):
...
+        # explicit elif branch is necessary because mypy does not pass otherwise
+        elif len(g) != 0 and parameter_names_to_values is not None:
+            for parameter_name in g:
+                func_str = func_str.replace(
+                    parameter_name, str(parameter_names_to_values[parameter_name])
+                )
         return func_str
 
-    def to_func(self, parameter_names_to_values=None):
+    def to_func(self, parameter_names_to_values: Optional[Dict[str, float]] = None) -> Callable:
```
jakobj, line 239: this returns a list of callable, doesn't it?

gp/cartesian_graph.py
```diff
@@ -350,7 +362,11 @@ def forward(self, x):
 
         return locals()["_c"]
 
-    def to_sympy(self, simplify=True, parameter_names_to_values=None):
+    def to_sympy(
+        self,
+        simplify: Optional[bool] = True,
+        parameter_names_to_values: Optional[Dict[str, float]] = None,
+    ) -> List[sympy_expr.Expr]:
```
jakobj, line 369: does this return value annotation cause any problems if sympy is not available?

gp/cartesian_graph.py
mschmidt87: This is the old code.

gp/cartesian_graph.py
mschmidt87: I think it's because `if g` is ambiguous, it does not distinguish between `g = None`, `g=False`, `g=[]`, (here the latter is the case we want to catch).

gp/cartesian_graph.py
mschmidt87: `mypy` doesn't recognize the fact that we never the for-loop if `len(g) == 0` and then complains that the for-loop will break if `parameter_names_to_values = None`.

gp/cartesian_graph.py
mschmidt87: Thanks.

gp/cartesian_graph.py
mschmidt87: probably yes, we currently don't test this in travis, but we should.

gp/cartesian_graph.py
mschmidt87: see above.

gp/cartesian_graph.py
mschmidt87: I see, the type of `self._genome` is clear because `genome` is specified and annotated as an input variable of the `__init__`

gp/cartesian_graph.py
mschmidt87, line 239: No, it returns a Callable that returns a list.

gp/cartesian_graph.py
mschmidt87:
```python
if torch_available:
    def to_torch(...):
         # full definition
else:
    def to_torch(*args, **kwargs):
        raise NotImplementedError(...)
```

mschmidt87: But I will investigate if this throws an error.

gp/cartesian_graph.py
```diff
@@ -117,22 +123,22 @@ def parse_genome(self, genome):
...
+    def hidden_nodes(self) -> List[Node]:
         return self._nodes[self._n_inputs : -self._n_outputs]
 
     @property
-    def output_nodes(self):
+    def output_nodes(self) -> List[Node]:
         return self._nodes[-self._n_outputs :]
 
-    def _determine_active_nodes(self):
+    def _determine_active_nodes(self) -> Dict[int, set]:
```
jakobj, line 141: could this be annotated with even more detail, like `Dict[int, Set[Node]]`

gp/cartesian_graph.py
```diff
@@ -216,17 +222,21 @@ def _format_output_str_of_all_nodes(self):
...
-        for parameter_name in g:
-            func_str = func_str.replace(
-                parameter_name, str(parameter_names_to_values[parameter_name])
-            )
+        # explicit elif branch is necessary because mypy does not pass otherwise
+        elif len(g) != 0 and parameter_names_to_values is not None:
+            for parameter_name in g:
+                func_str = func_str.replace(
+                    parameter_name, str(parameter_names_to_values[parameter_name])
+                )
```
jakobj, line 236: maybe add an explicit `else: pass` branch to indicate that this is not forgotten but intentional to do nothing in that case?

gp/cartesian_graph.py
```diff
@@ -363,13 +379,13 @@ def to_sympy(self, simplify=True, parameter_names_to_values=None):
...
         Returns
         ----------
-        List[sympy.core.Expr]
+        List[sympy.core.expr.Expr]
             List of SymPy expressions.
         """
-        if sympy is None:
-            raise ModuleNotFoundError("No module named 'sympy' (extra requirement)")
+        if not sympy_available:
+            raise ModuleNotFoundError("No module named 'sympy' (extra requirement)")  #
```
jakobj, line 386:
```suggestion
            raise ModuleNotFoundError("No module named 'sympy' (extra requirement)")
```

gp/cartesian_graph.py
```diff
@@ -4,60 +4,66 @@
 
 try:
     import sympy
+    from sympy.core import expr as sympy_expr
```
jakobj, line 7: why not `import sympy.core` and then use `sympy.core.expr`? it's not much longer and it's much clearer that you accessing a member of the sympy package. e.g., below i was wondering whether `sympy_expr` is something we introduced ;)

gp/genome.py
```diff
@@ -53,45 +73,46 @@ def __init__(self, n_inputs, n_outputs, n_columns, n_rows, levels_back, primitiv
             1 + self._primitives.max_arity
         )  # one function gene + multiple input genes
 
-        self._dna = None  # stores dna as list of alleles for all regions
+        self._dna: List[int] = []  # stores dna as list of alleles for all regions
 
         # constants used as identifiers for input and output nodes
-        self._id_input_node = -1
-        self._id_output_node = -2
+        self._id_input_node: int = ID_INPUT_NODE
```
jakobj, line 79: omg this is soo much better with globals :heart_eyes: but now i'm starting to think. shouldn't we just introduce `InputNode` and `OutputNode` as always-present members of `Primitives` that would hence have function id 0 and 1 respectively? The numbering for all other nodes would then start at 2. this seems more natural, doesn't it? and we would need only one magic number of unused gene (which i would prefer to call "non-coding gene" btw ;) )

gp/genome.py
```diff
@@ -103,16 +124,18 @@ def __repr__(self):
         s += ")"
         return s
 
-    def _create_input_region(self):
+    def _create_input_region(self) -> List[int]:
 
         region = []
         region.append(self._id_input_node)
         # empty input genes (None) since input nodes do not receive
```
jakobj, line 131:
```suggestion
        # non-coding input genes since input nodes do not receive
```

gp/genome.py
jakobj: also: do we need these members if we have the globals?

gp/genome.py
```diff
@@ -103,16 +124,18 @@ def __repr__(self):
...
         # inputs themselves
-        region += [None] * self._primitives.max_arity
+        region += [self._id_unused_gene] * self._primitives.max_arity
         return region
 
-    def _create_random_hidden_region(self, rng, permissable_inputs):
+    def _create_random_hidden_region(
+        self, rng: np.random.RandomState, permissable_inputs: List[int]
+    ) -> list:
```
jakobj, line 138:
```suggestion
    ) -> List[int]:
```

gp/genome.py
```diff
@@ -103,16 +124,18 @@ def __repr__(self):
...
         region.append(self._id_input_node)
         # empty input genes (None) since input nodes do not receive
         # inputs themselves
-        region += [None] * self._primitives.max_arity
+        region += [self._id_unused_gene] * self._primitives.max_arity
         return region
 
-    def _create_random_hidden_region(self, rng, permissable_inputs):
+    def _create_random_hidden_region(
+        self, rng: np.random.RandomState, permissable_inputs: List[int]
```
jakobj, line 137:
```suggestion
        self, rng: np.random.RandomState, permissible_inputs: List[int]
```

typo, whoops O:)

gp/genome.py
```diff
@@ -121,18 +144,20 @@ def _create_random_hidden_region(self, rng, permissable_inputs):
 
         return region
 
-    def _create_random_output_region(self, rng, permissable_inputs):
+    def _create_random_output_region(
+        self, rng: np.random.RandomState, permissable_inputs: List[int]
+    ) -> list:
```
jakobj, line 149:
```suggestion
    ) -> List[int]:
```

gp/genome.py
```diff
@@ -121,18 +144,20 @@ def _create_random_hidden_region(self, rng, permissable_inputs):
...
 
         region = []
         region.append(self._id_output_node)
         region.append(rng.choice(permissable_inputs))
-        # output nodes have only on input, other genes are hence left
+        # output nodes have only one input, other genes are hence left
         # empty
```
jakobj, line 155:
```suggestion
        # output nodes have only one input, other genes are hence
        # non-coding
```

gp/genome.py
```diff
@@ -190,10 +215,10 @@ def _permissable_inputs(self, region_idx):
 
         return permissable_inputs
 
-    def _permissable_inputs_for_output_region(self):
+    def _permissable_inputs_for_output_region(self) -> list:
```
jakobj, line 218:
```suggestion
    def _permissable_inputs_for_output_region(self) -> List[int]:
```

gp/hl_api.py
```diff
@@ -1,18 +1,21 @@
 import numpy as np
+from typing import Optional, Callable
+from .population import Population
```
jakobj, line 3:
```suggestion

from .population import Population
```

gp/cartesian_graph.py
jakobj, line 239:
```suggestion
    def to_func(self, parameter_names_to_values: Optional[Dict[str, float]] = None) -> Callable[[List[float]], List[float]]:
```

can we specify in even more detail what kind of callable this returns? btw same for `to_numpy` except it would accept/return lists but np.arrays?

gp/local_search/gradient_based.py
```diff
@@ -1,12 +1,28 @@
...
+    import torch  # noqa: F401
+
+    torch_available = True
 except ModuleNotFoundError:
-    torch = None
+    torch_available = False
+
+from typing import Callable, List, Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
```
jakobj, line 12: why do we need this?

gp/local_search/gradient_based.py
```diff
@@ -1,12 +1,28 @@
...
+    import torch.optim.Optimizer
+
+from ..individual import Individual  # noqa: F401
 
 
-def gradient_based(individuals, objective, lr, gradient_steps, optimizer=None, clip_value=None):
+def gradient_based(
+    individuals: List[Individual],
+    objective: Callable,
```
jakobj, line 20:
```suggestion
    objective: Callable[[Individual], Individual],
```

gp/node.py
```diff
@@ -43,33 +53,36 @@ def __init__(self, idx, inputs):
 
         assert idx not in inputs
 
-    def __init_subclass__(cls, **kwargs):
-        super().__init_subclass__(**kwargs)
+    def __init_subclass__(cls: Type[Node]) -> None:
+        super().__init_subclass__()
         register(cls)
 
+    def __call__(self, x: float, graph: Union[CartesianGraph, List[Node]]) -> None:
```
jakobj, line 60: why do we need the `Union` here? shouldn't it always be `CartesianGraph`?

gp/node.py
```diff
@@ -43,33 +53,36 @@ def __init__(self, idx, inputs):
...
+        raise NotImplementedError
+
     @property
-    def arity(self):
+    def arity(self) -> Union[None, int]:
```
jakobj, line 64:
```suggestion
    def arity(self) -> int:
```

gp/node.py
```diff
@@ -157,13 +170,13 @@ class Add(Node):
 
     _arity = 2
 
-    def __init__(self, idx, inputs):
+    def __init__(self, idx: int, inputs: List[int]) -> None:
         super().__init__(idx, inputs)
 
-    def __call__(self, x, graph):
+    def __call__(self, x: float, graph: Union[CartesianGraph, List[Node]]) -> None:
```
jakobj, line 176: (see above)

gp/primitives.py
```diff
@@ -34,7 +37,10 @@ def __init__(self, primitives):
 
         self._determine_max_arity()
 
-    def _determine_max_arity(self):
+    def __iter__(self) -> Iterator:
+        return iter([self[i] for i in range(len(self._primitives))])
```
jakobj, line 41: this is a potentially dangerous statement: it's important that the order of primitives is exactly conserved from how this object was initialized (ll29) otherwise the mapping from function id to function changes when an individual is cloned (where this function is used). maybe we should use an `OrderedDict` and just return `self._primitives.values()` here? at minimum this line should have a "WARNING" comment that reminds us of the importance of the correct order

gp/primitives.py
```diff
@@ -44,7 +50,7 @@ def _determine_max_arity(self):
 
         self._max_arity = arity
 
-    def sample(self, rng):
+    def sample(self, rng: np.random.RandomState) -> int:
```
jakobj, line 53:
```suggestion
    def sample_index(self, rng: np.random.RandomState) -> int:
```

or something similar, this name is confusing! although maybe not part of this PR, strictly speaking ;)

test/test_cartesian_graph.py
```diff
@@ -31,7 +32,20 @@ def test_to_func_simple():
     primitives = [gp.Add]
     genome = gp.Genome(2, 1, 1, 1, 1, primitives)
 
-    genome.dna = [-1, None, None, -1, None, None, 0, 0, 1, -2, 2, None]
+    genome.dna = [
+        ID_INPUT_NODE,
```
jakobj, line 36: omg, so clear :heart_eyes:

test/test_genome.py
```diff
@@ -212,7 +344,7 @@ def objective(ind):
 
 
 def test_is_gene_in_input_region(rng_seed):
-    genome = gp.Genome(2, 1, 2, 1, None, [gp.Add])
+    genome = gp.Genome(2, 1, 2, 1, ID_UNUSED_GENE, [gp.Add])
```
jakobj, line 347:
```suggestion
    genome = gp.Genome(2, 1, 2, 1, None, [gp.Add])
```

a bit too much search & replace? ;) this is the value for `levels_back`, also applies below!

gp/cartesian_graph.py
mschmidt87: yes. done.

gp/cartesian_graph.py
mschmidt87: Yes, done.

gp/cartesian_graph.py
mschmidt87: Thanks, removed.

gp/cartesian_graph.py
mschmidt87: yes, thanks, done.

gp/genome.py
mschmidt87: Out of scope of this PR. I created a separate ticket. https://github.com/Happy-Algorithms-League/python-gp/issues/105

mschmidt87: I did rename the `unused_gene` to `non_coding_gene`.

gp/node.py
mschmidt87: what is above?

gp/primitives.py
mschmidt87: `self._primitives` is created with this code:
```
for i in range(len(primitives)):
            self._primitives[i] = primitives[i]
```
Thus the elements are with increasing index `0, 1, 2,...` into the dict, which exactly matches the way they are retrieved here (with a `for i in range(len(...)`. So I think this is safe here.

mschmidt87: It could be written in a better way, but that is out of scope of this PR. Please create an issue if you think it's necessary.

gp/primitives.py
mschmidt87: Yes, out of scope. Create an issue: #106

test/test_genome.py
mschmidt87: Yes, fixed.

mschmidt87:
I addressed all your comments, @jakobj . Please take another look.

gp/genome.py
mschmidt87: out of scope.

gp/genome.py
mschmidt87: done.

gp/genome.py
mschmidt87: done.

gp/genome.py
mschmidt87: done.

gp/genome.py
mschmidt87: done.

gp/hl_api.py
mschmidt87: done

gp/node.py
mschmidt87: removed.

gp/node.py
mschmidt87: fixed.

gp/local_search/gradient_based.py
mschmidt87: Should be this Callable[[torch.nn.Module], torch.Tensor]. Done.

gp/cartesian_graph.py
```diff
@@ -4,60 +4,66 @@
...
+    torch_available = True
 except ModuleNotFoundError:
-    torch = None
+    torch_available = False
 
 
-from .node import InputNode, OutputNode, Parameter
+from .node import Node, InputNode, OutputNode, Parameter
+from .genome import Genome
+from typing import Callable, Dict, List, Optional, Set
```
jakobj, line 23:
```suggestion
from typing import Callable, Dict, List, Optional, Set


from .node import Node, InputNode, OutputNode, Parameter
from .genome import Genome
```

gp/population.py
```diff
@@ -1,14 +1,17 @@
 import numpy as np
 
 from .individual import Individual, IndividualMultiGenome
+from typing import List
```
jakobj, line 4:
```suggestion
import numpy as np
from typing import List


from .individual import Individual, IndividualMultiGenome
```

gp/primitives.py
```diff
@@ -1,19 +1,22 @@
 from .node import Node
 
+from typing import Iterator, List, Tuple, Type
+import numpy as np  # type: ignore
+
```
jakobj, line 5:
```suggestion
from typing import Iterator, List, Tuple, Type
import numpy as np  # type: ignore


from .node import Node
```

gp/primitives.py
jakobj: it's not that your operation is unsafe, but someone who wants to do "smart" refactoring might change this for example to iterating over the entries of `_primitives`. created #107 to address this

gp/genome.py
```diff
@@ -1,11 +1,29 @@
+import numpy as np
+
+from typing import Generator, List, Optional, Tuple, Type
+
+from .node import Node
```
jakobj, line 5:
```suggestion
import numpy as np
from typing import Generator, List, Optional, Tuple, Type


from .node import Node
```

gp/cartesian_graph.py
```diff
@@ -216,17 +222,23 @@ def _format_output_str_of_all_nodes(self):
...
+        if len(g) != 0:
+            if parameter_names_to_values is None:
+                raise ValueError("parameter node found but no value dict provided")
+            elif parameter_names_to_values is not None:
+                for parameter_name in g:
+                    func_str = func_str.replace(
+                        parameter_name, str(parameter_names_to_values[parameter_name])
+                    )
+        else:
+            pass
```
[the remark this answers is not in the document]
mschmidt87: done.

gp/cartesian_graph.py
mschmidt87: done.

gp/population.py
mschmidt87: done.

gp/primitives.py
mschmidt87: done.

gp/genome.py
mschmidt87: I prefer leaving one row in between.

PR #65, opened 2020-04-10
Replace assert with approximate limits by pytest.approx constructions

### LINKED ISSUE #41: Use `pytest.approx` functions in tests
Some of the tests are using constructions like `assert abs(1.0 - y[0]) < 1e-15`. We should probably use `pytest.approx` in these for consistent error reporting.

mschmidt87:
This PR fixes #41 .

### DIFF
--- a/test/test_cartesian_graph.py
+++ b/test/test_cartesian_graph.py
@@ -29,7 +29,7 @@ def test_direct_input_output():
     x = [2.14159]
     y = graph(x)
 
-    assert abs(x[0] - y[0]) < 1e-15
+    assert pytest.approx(x[0] == y[0])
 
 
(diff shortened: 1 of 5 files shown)


test/test_cartesian_graph.py
```diff
@@ -29,7 +29,7 @@ def test_direct_input_output():
     x = [2.14159]
     y = graph(x)
 
-    assert abs(x[0] - y[0]) < 1e-15
+    assert pytest.approx(x[0] == y[0])
```
jakobj, line 32: interesting, i thought one has to write `x[0] == pytest.approx(y[0])`, but of course this is much nicer! :+1:

PR #71, opened 2020-04-15
Add consistency checks to disk cache decorator

### LINKED ISSUE #67: Add consistency check to caching decorator
Currently, the caching decorator does not check whether the function to be cached is actually the same as the one that the cache was created for. Thus, it can happen that the decorator mistakes a different function for the cached function. 

This is a minimal reproducer:

```python
import gp


@gp.utils.disk_cache("example_caching_cache.pkl")
def f1(x):
    return x


@gp.utils.disk_cache("example_caching_cache.pkl")
def f2(x):
    return x**2

for i in range(10):
    print(f"x = {f1(i)}, x^2 = {f2(i)}")
```
yields
```
x = 0, x^2 = 0
x = 1, x^2 = 1
x = 2, x^2 = 2
x = 3, x^2 = 3
x = 4, x^2 = 4
x = 5, x^2 = 5
x = 6, x^2 = 6
x = 7, x^2 = 7
x = 8, x^2 = 8
x = 9, x^2 = 9
```

Proposed solution is: Let the decorator verify for a cached function that the function called with  the first cached argument yields the same result as the cached result.

jakobj:
As proposed by @mschmidt87 this PR implements a consistency check in the disk cache decorator. It checks for consistency of arguments, keyword arguments and return values to avoid reusing the disk cache for different functions.

Unfortunately flake8 complains about a "too complex" function (`disk_cache`). Any idea how to address this @mschmidt87? Move the whole decorator to a separate module? I like that the required functions are only defined within the decorator function, hence not visible outside.

fixes #67

### DIFF
--- a/gp/utils.py
+++ b/gp/utils.py
@@ -80,13 +80,23 @@ def disk_cache(fn):
     decorated function is called with the same arguments it returns the stored
     values from disk instead of executing the function.
 
-    Warning: this implementation is neither optimized for speed nor storage
+    Consistency of the cache is checked upon the first call of the
+    decorated function by making sure the decorated function returns
+    the same value as the first argument from the cache.
+
+    WARNING: this implementation is neither optimized for speed nor storage
     space and does not limit the size of the cache file.
+
+    WARNING: the consistency check may pass incorrectly if the
+    decorated function happens to return a consistent value for the
+    first argument from the cache although it returns different values
+    for other arguments.
 
     Parameters
     ----------
     fn : Callable
         Function to be cached.
+
     """
     first_function_call = True
 
--- a/gp/utils.py
+++ b/gp/utils.py
@@ -1,9 +1,76 @@
 import functools
 import hashlib
+import numpy as np
 import os
 import pickle
 
 from .node import primitives_dict
+
+
+def __check_cache_consistency(fn, func, args, kwargs):
+    if os.path.isfile(fn):
+        with open(fn, "rb") as f:
+            try:
+                cursor = pickle.load(f)
+            except EOFError:
+                return  # no entry yet, so not possible to check
+
+            cached_item = list(cursor.values())[0]
+
+            if len(args) != len(cached_item["args"]):
+                raise RuntimeError(
+                    "inconsistent arguments"
+                    " -- are different functions using the same cache file?"
+                    " please use different cache files for different functions"
+                )
+
+            if kwargs.keys() != cached_item["kwargs"].keys():
+                raise RuntimeError(
+                    "inconsistent keyword arguments"
+                    " -- are different functions using the same cache file?"
+                    " please use different cache files for different functions"
+                )
+
+            return_value = func(*cached_item["args"], **cached_item["kwargs"])
+
+            if not np.isclose(return_value, cached_item["return_value"]):
+                raise RuntimeError(
+                    "inconsistent return values"
+                    " -- are different functions using the same cache file?"
+                    " please use different cache files for different functions"
+                )
+
+
+def __compute_key_from_args(*args, **kwargs):
+    s = str(args) + str(kwargs)
+    return hashlib.sha1(s.encode("utf-8")).hexdigest()
+
+
+def __find_result_in_cache_file(fn, key):
+    if os.path.isfile(fn):
+        with open(fn, "rb") as f:
+            while True:
+                try:
+                    cursor = pickle.load(f)
+                except EOFError:
+                    break
+
+                if key in cursor:
+                    return cursor[key]
+
+    return None
+
+
+def __store_new_cache_entry(fn, key, return_value, args, kwargs):
+    with open(fn, "ab") as f:
+
+        result = {
+            "args": args,
+            "kwargs": kwargs,
+            "return_value": return_value,
+        }
+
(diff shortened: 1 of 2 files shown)


test/test_utils.py
```diff
@@ -29,6 +29,55 @@ def objective(label):
...
+    @gp.utils.disk_cache(cache_fn)
+    def objective_f_with_kwargs(x, test=None):
+        return x
+
+    x = 2
+
+    assert objective_f(x) == pytest.approx(x)
+    assert objective_h(x) == pytest.approx(x)
+
+    # calling a different function decorated with same name but
```
mschmidt87, line 61: with "same name", you mean "same filename", right? Can you make this more clear then. Also for the other comments.

jakobj: good point, fixed

jakobj:
you are right, with the current implementation it's not possible to cache two different functions in the same cache file. i guess this is a design decision: do we want to
1. raise an error if a user is trying to use the same name
2. create a hierarchy in the file with separate entries for each function
3. modify the filename automatically if a user is trying to cache two different functions in the same file, so that they are actually using different cache files?

i would argue in favor of 1. since it's easy to fix on the user side. 2. is difficult because we are appending pickle streams, not modifying the existing ones. 3. is non-trivial since the modification needs to be determinstic, i.e., identical functions should lead to identical modification, excluding random strings, strings that are based on object identity etc. maybe you have an idea?

regarding the complexity: one of the nice things of the current implementation is that we don't need to pass around the filename since the helper function are defined within the scope of the decorator and can just use the nonlocal variable `fn`. we would need to change this if we move them out. i really don't know what the best practice is here.

mschmidt87:
> regarding the complexity: one of the nice things of the current implementation is that we don't need to pass around the filename since the helper function are defined within the scope of the decorator and can just use the nonlocal variable `fn`. we would need to change this if we move them out. i really don't know what the best practice is here.

I don't see a problem in passing one more variable, i.e., the filename around. I think it makes sense to restrict the complexity of functions via `flake8` and the current construction with so many functions defined inside a function looks like bad practice to me. I'd prefer to define them outside, and use the `__...` function names.

mschmidt87:
> you are right, with the current implementation it's not possible to cache two different functions in the same cache file. i guess this is a design decision: do we want to
> 
>     1. raise an error if a user is trying to use the same name
> 
>     2. create a hierarchy in the file with separate entries for each function
> 
>     3. modify the filename automatically if a user is trying to cache two different functions in the same file, so that they are actually using different cache files?
> 
> 
> i would argue in favor of 1. since it's easy to fix on the user side. 2. is difficult because we are appending pickle streams, not modifying the existing ones. 3. is non-trivial since the modification needs to be determinstic, i.e., identical functions should lead to identical modification, excluding random strings, strings that are based on object identity etc. maybe you have an idea?

Okay, makes sense to me, option 1 is fine for me. However, I would prefer to adapt the error message. Rather than asking a question, let's provide a concrete suggestion to the user, I made a suggestion.

gp/utils.py
```diff
@@ -21,36 +22,86 @@ def disk_cache(fn):
...
+                    cursor = pickle.load(f)
+                except EOFError:
+                    return  # no entry yet, so not possible to check
+
+                cached_item = list(cursor.values())[0]
+
+                if len(args) != len(cached_item["args"]):
+                    raise RuntimeError(
+                        "inconsistent arguments"
+                        "-- are different functions using the same cache file?"
```
mschmidt87, line 40:
```suggestion
                        "-- are different functions using the same cache file?"
                        "In this case, use different cache files for the different functions."
```

gp/utils.py
```diff
@@ -21,36 +22,86 @@ def disk_cache(fn):
...
+                    )
+
+                if kwargs.keys() != cached_item["kwargs"].keys():
+                    raise RuntimeError(
+                        "inconsistent keyword arguments"
+                        "-- are different functions using the same cache file?"
```
mschmidt87, line 46:
```suggestion
                        "-- are different functions using the same cache file?"
                        "In this case, use different cache files for the different functions."
```

gp/utils.py
```diff
@@ -21,36 +22,86 @@ def disk_cache(fn):
...
+                    )
+
+                return_value = func(*cached_item["args"], **cached_item["kwargs"])
+
+                if not np.isclose(return_value, cached_item["return_value"]):
+                    raise RuntimeError(
+                        "inconsistent return values"
+                        "-- are different functions using the same cache file?"
```
mschmidt87, line 54:
```suggestion
                        "-- are different functions using the same cache file?"
                        "In this case, use different cache files for the different functions."
```

jakobj:
thanks for your comments! i've reduced the complexity of the decorator function as suggested and adapted the error message, please have another look

jakobj:
that's a good point. if the return value on the first function call is identical for two different functions, then no error is raised although subsequent function evaluations return wrong values. we could introduce a parameter that controls how many times consistency should be checked, or alternatively, as you suggest, document this scenario. i would opt for the latter.

jakobj:
i've updated the doc, please have another look

jakobj:
restarting the job solved the issue.

PR #72, opened 2020-04-16
Fix pytest.approx usage in all tests

### LINKED ISSUE #41: Use `pytest.approx` functions in tests

mschmidt87:
This PR fixes the wrong usage of `pytest.approx` introduced by #65 (which intended to fix #41 ).


test/test_cartesian_graph.py
```diff
@@ -157,13 +157,13 @@ def test_to_torch_and_backprop():
...
-    assert pytest.approx(float(loss.detach()) == 0.0)
+    assert float(loss.detach()) == pytest.approx(0.0)
 
     x = [3.0]
     x_torch = torch.Tensor(x).view(1, 1)
-    assert pytest.approx(c(x_torch)[0].detach().numpy() != graph(x))
+    assert c(x_torch)[0].detach().numpy() != pytest.approx(graph(x))
     graph.update_parameters_from_torch_class(c)
-    assert pytest.approx(c(x_torch)[0].detach().numpy() == graph(x))
+    assert c(x_torch)[0][0].detach().numpy() == pytest.approx(graph(x)[0].detach().numpy())
```
jakobj, line 166: why the extra zero here? shouldn't it work just like two lines above?

jakobj, line 166:
```suggestion
    assert c(x_torch)[0].detach().numpy() == pytest.approx(graph(x))
```

test/test_node.py
```diff
@@ -67,7 +67,7 @@ def test_mul():
     x = [5.0, 1.5]
     y = graph(x)
 
-    assert pytest.approx((x[0] * x[1]) == y[0])
+    assert (x[0] * x[1]) == pytest.approx(y[0])
```
jakobj, line 70:
```suggestion
    assert x[0] * x[1] == pytest.approx(y[0])
```

test/test_node.py
```diff
@@ -88,7 +88,7 @@ def test_div():
     x = [5.0, 1.5]
     y = graph(x)
 
-    assert pytest.approx((x[0] / x[1]) == y[0])
+    assert (x[0] / x[1]) == pytest.approx(y[0])
```
jakobj, line 91:
```suggestion
    assert x[0] / x[1] == pytest.approx(y[0])
```

test/test_node.py
```diff
@@ -109,7 +109,7 @@ def test_pow():
     x = [5.0, 1.5]
     y = graph(x)
 
-    assert pytest.approx((x[0] ** x[1]) == y[0])
+    assert (x[0] ** x[1]) == pytest.approx(y[0])
```
jakobj, line 112:
```suggestion
    assert x[0] ** x[1] == pytest.approx(y[0])
```

test/test_cartesian_graph.py
```diff
@@ -131,7 +131,7 @@ def test_compile_addsubmul():
     x = [5.0, 2.0]
     y = f(x)
 
-    assert pytest.approx(((x[0] * x[1]) - (x[0] - x[1])) == y[0])
+    assert ((x[0] * x[1]) - (x[0] - x[1])) == pytest.approx(y[0])
```
jakobj, line 134:
```suggestion
    assert (x[0] * x[1]) - (x[0] - x[1]) == pytest.approx(y[0])
```

test/test_cartesian_graph.py
mschmidt87: This does not work with `pytest.approx`, it will give 
`*** TypeError: cannot make approximate comparisons to non-numeric values: [tensor(-6.4248, grad_fn=<MulBackward0>)]  at index 0`.

Doesn't this happen on your machine?

mschmidt87: See below, your suggested approach does not work on my machine, so I need to transform the right hand side. To be able to call `detach().numpy()` I need to do `[0]` first, because it is a `list`.

When I do that, I need to take the extra `[0]` on the left hand side.

mschmidt87:
I applied the suggested changes and added a reply to the question. Please check again @jakobj

test/test_cartesian_graph.py
jakobj: ah interesting, this is a bug! the `update_parameters_from_torch_class` changes the output of the corresponding node to a torch parameter (https://github.com/jakobj/python-gp/blob/master/gp/cartesian_graph.py#L290). the line in `cartesian_graph` should read `n._output = eval(f"torch_cls._p{n._idx}[0].item()")` instead to extract the value and avoid setting the node output to a torch parameter. could you fix that quickly in this PR so you can use the suggested comparison without the detach etc?

jakobj: (see reply above)

mschmidt87: Okay, done in #74 . Note that #74 now has to be merged first to make CI pass on this PR.

mschmidt87:
Addressed remaining comment, @jakobj

jakobj:
merged #74. could you please rebase to also fix the merge conflicts?

PR #77, opened 2020-04-20
Enh/to numpy

### LINKED ISSUE #56: Provide `to_numpy` compilation
As far as I can see the `to_func` produces a python function that can only operate on scalar values. We should provide a `to_numpy` function that returns a python function able to digest vectors. What do you think @mschmidt87?

jakobj:
support `to_numpy` compilation by replacing `x[0]` with `x[:, 0]` etc in regular output string of compiled expression. assumes that inputs are numpy arrays with structure `(batch_size, input_size)`. what do you think of this torch-style convention with regards to numpy arrays @mschmidt87 ?

fixes #56

### DIFF
--- a/gp/cartesian_graph.py
+++ b/gp/cartesian_graph.py
@@ -1,4 +1,5 @@
 import collections
+import numpy as np  # noqa: F401
 
 try:
     import sympy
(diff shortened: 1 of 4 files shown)


jakobj:
test fails due to constant not being expanded to the correct shape, will fix

jakobj:
ok, fixed, reused the parametrized test for torch output shape, should be ready for review now @mschmidt87

PR #82, opened 2020-04-24
Simplify `_determine_active_nodes` and add doc string

jakobj:


mschmidt87:
A principal question: According to the docstring, an "active" node is a node that is used in the computational graph described by the cartesian graph. The mutation method only uses active nodes. Does this mean that an inactive node is never mutated? Is this the intended behavior of CGP?

jakobj:
the `mutate` method should not only be applied to active nodes, i agree. where do you see that? or is just the docstring misleading? related: why is the `determine_active_regions` methods called in `Individual._mutate` instead of just locally in `mutate`?

mschmidt87:
> the `mutate` method should not only be applied to active nodes, i agree. where do you see that? or is just the docstring misleading? related: why is the `determine_active_regions` methods called in `Individual._mutate` instead of just locally in `Graph.mutate`?

Yes, you're right, I had misunderstood the the `mutate` function. So, s'all good.

Issue #75, opened 2020-04-20
Investigate random timeouts from Travis


jakobj:
like this https://travis-ci.org/github/jakobj/python-gp/jobs/675940992

jakobj:
maybe related to `test_evolve_two_expressions` sometimes failing like in https://travis-ci.org/github/jakobj/python-gp/jobs/677775573

jakobj:
first step: fix the seed for all tests

jakobj:
seems to have been fixed by #84. closing. we can reopen if the need arises.

PR #84, opened 2020-04-24
Use fixed rng seed for tests

jakobj:
as the title suggests, this PR replaces the stochastic random seed with a fixed random seed from `conftest` in all tests. it also uses a few more fixtures where appropriate.

this should avoid the random failures of tests possibly responsible for #75

### DIFF
--- a/test/test_hl_api.py
+++ b/test/test_hl_api.py
@@ -31,29 +31,11 @@ def _objective_test_population(individual, rng_seed):
     return individual
 
 
-def _test_population(rng_seed, n_processes):
-
-    population_params = {"n_parents": 5, "mutation_rate": 0.05, "seed": rng_seed}
-
-    genome_params = {
-        "n_inputs": 2,
-        "n_outputs": 1,
-        "n_columns": 3,
-        "n_rows": 3,
-        "levels_back": 2,
-        "primitives": [gp.Add, gp.Sub, gp.ConstantFloat],
-    }
-
-    ea_params = {
-        "n_offsprings": 5,
-        "n_breeding": 5,
-        "tournament_size": 2,
-        "n_processes": n_processes,
-    }
+def _test_population(population_params, genome_params, ea_params):
 
     evolve_params = {"max_generations": 2000, "min_fitness": -1e-12}
 
-    np.random.seed(rng_seed)
+    np.random.seed(population_params["seed"])
 
     pop = gp.Population(**population_params, genome_params=genome_params)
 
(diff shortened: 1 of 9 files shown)


test/test_hl_api.py
```diff
@@ -33,9 +31,9 @@ def f_target(x):  # target function
     return individual
 
 
-def _test_population(n_processes):
+def _test_population(rng_seed, n_processes):
 
-    population_params = {"n_parents": 5, "mutation_rate": 0.05, "seed": SEED}
+    population_params = {"n_parents": 5, "mutation_rate": 0.05, "seed": rng_seed}
```
mschmidt87, line 36: Can't this be replaced by the fixture?

test/test_hl_api.py
```diff
@@ -67,30 +65,31 @@ def _test_population(n_processes):
...
     assert fitness_per_n_processes[0] == pytest.approx(fitness_per_n_processes[2])
 
 
-def test_pop_uses_own_rng():
+def test_pop_uses_own_rng(rng_seed):
     """Test independence of Population on global numpy rng.
     """
 
-    population_params = {"n_parents": 5, "mutation_rate": 0.05, "seed": SEED}
+    population_params = {"n_parents": 5, "mutation_rate": 0.05, "seed": rng_seed}
```
mschmidt87, line 92: Same here, replace with the fixture?

test/test_hl_api.py
```diff
@@ -150,7 +149,7 @@ def f1(x):
 
         return individual
 
-    population_params = {"n_parents": 5, "mutation_rate": 0.05, "seed": SEED}
+    population_params = {"n_parents": 5, "mutation_rate": 0.05, "seed": rng_seed}
```
mschmidt87, line 152: same here, replace with fixture?

test/test_hl_api.py
```diff
@@ -198,9 +197,9 @@ def _objective_speedup_parallel_evolve(individual):
 
 
 @pytest.mark.skip(reason="Test is not robust against execution in CI.")
-def test_speedup_parallel_evolve():
+def test_speedup_parallel_evolve(rng_seed):
 
-    population_params = {"n_parents": 4, "mutation_rate": 0.05, "seed": SEED}
+    population_params = {"n_parents": 4, "mutation_rate": 0.05, "seed": rng_seed}
```
mschmidt87, line 202: and here.

test/test_hl_api.py
jakobj: good point!

test/test_hl_api.py
jakobj: yes!

test/test_hl_api.py
jakobj: yes!

test/test_hl_api.py
jakobj: yes!

jakobj:
thanks for the suggestions @mschmidt87, i've included them now. please have another look

PR #86, opened 2020-04-27
Fix caching example and update decorator docstring

jakobj:
This PR avoids using a nested function in the caching example thus following the recommendation from the updated docstring ;) 

[edit]
for completeness, quoted from the discussion below:
> it's a scope issue: nested functions are defined and hence decorated each time the outer function is called. as the consistency check is performed on the first execution after decoration it is hence executed on every outer function call which should not be the case for efficient caching

[/edit]

### DIFF
--- a/examples/example_caching.py
+++ b/examples/example_caching.py
@@ -1,4 +1,5 @@
 import numpy as np
+import time
 
 import gp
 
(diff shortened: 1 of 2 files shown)


jakobj:
this does not fix the issue that multiprocessing pickles and unpickles the functions, causing the decoration to happen on every function evaluation. do you have any idea of how to track whether this is the first function call across multiple processes?

jakobj:
ok, how about this suggestion: the user can provide an optional keyword argument, e.g., `check_consistency` to the decorator which by default is `True`. if `check_consistency` we use the currently implemented checks that work for single processes. in addition we provide a function `check_cache_consistency(<fn>, <objective>)` that the user can call manually. So
- single process: decorate with `disk_cache(<fn>)`
- mulitple processes: decorate with `disk_cache(<fn>, check_consistency=False)` and call `check_cache_consistency(<fn>, <objective>)` manually once at the beginning, e.g., before `evolve`

this way we end up the issue of caching actually _increasing_ the runtime in the multi process case. i currently can't come up with a more user friendly solution. any ideas @mschmidt87 ?

mschmidt87:
Your suggestions makes sense to me, but we need to take care to properly document and explain this.

This PR which is unrelated to this, looks fine to me. However, in order to have this properly documented, can you 
1. Describe the problem that led to this PR in a bit more detail
2. Create an issue for the multiprocessing problem.

jakobj:
ok, 2) is covered by #91. I'm not sure about 1), what exactly do you mean? should i improve the warning in the docstring? it's a scope issue: nested functions are defined and hence decorated each time the outer function is called. as the consistency check is performed on the first execution after decoration it is hence executed _on every outer function call_ which should not be the case for efficient caching ;)

mschmidt87:
Yes, thanks, I was just missing that explanation. This PR came out of nowhere for me, or was this problem documented somewhere before?

Anyway, perhaps you can move the explanation into the PR and then merge.

jakobj:
no, i discovered this problem while working on reproducing our simulations. hence it's not documented anywhere else. i've quoted my explanation above at the top now for documentation.

thanks for the review :tada:

PR #95, opened 2020-05-04
Wrap objective function before fitness evaluation to return early if fitness of an individual is not None

jakobj:
Previously the user needed to implement this early return manually to avoid reevaluating individuals for which the (deterministic) fitness was already calculated. This PR introduces a new parameter for the evolutionary algorithm where the user can decide whether to reevaluate individuals with fitness not None.

After implementing this it's not clear to me whether this is really useful or should just be left to the user and properly documented. It increases the complexity of the module code while (typically) saving only two lines in user code (e.g., https://github.com/Happy-Algorithms-League/python-gp/compare/master...jakobj:enh/only-evaluate-none?expand=1#diff-e26da432872da9fa28a952944b363633L42). Also it can lead to unexpected results for non-deterministic fitness functions if the user overlooks this option. What do you think about this @mschmidt87?

### DIFF
--- a/examples/example_caching.py
+++ b/examples/example_caching.py
@@ -39,11 +39,7 @@ def inner_objective(expr):
 
 
 def objective(individual):
-    if individual.fitness is not None:
-        return individual
-
     individual.fitness = -inner_objective(individual.to_sympy())
-
     return individual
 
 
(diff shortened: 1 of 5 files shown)


mschmidt87:
Another argument against this is that we typically solve this problem with the caching decorator, don't we? If the decorator is used, then this problem is non-existent because the fitness function will not be reevaluated on the same individual (result already cached).

jakobj:
agreed. too many arguments against this, thus closing.

PR #99, opened 2020-05-05
Check cache consistency on first decoration, not on first call

### LINKED ISSUE #91: Caching decorator doubles runtime for n_processes > 1
As shown in  #86, the decorator should not be used on nested functions as the consistency check (one execution of the objective on cached input) is applied on the first call after decoration. In a multiprocessing scenario (`n_processes > 1`) the decorator is applied on _each evaluation of the decorated function_ even if the function is defined globally due to the pickling/unpickling within multiprocessing thus doubling the runtime.

One way to avoid this problem is suggested in the discussion of #86: 
> the user can provide an optional keyword argument, e.g., `check_consistency` to the decorator which by default is `True`. if `check_consistency` [is true] we use the currently implemented checks that work for single processes. in addition we provide a function `check_cache_consistency(<fn>, <objective>)` that the user can call manually. So
> - single process: decorate with `disk_cache(<fn>)`
> - mulitple processes: decorate with `disk_cache(<fn>, check_consistency=False)` and call `check_cache_consistency(<fn>, <objective>)` manually once at the beginning, e.g., before `evolve`

jakobj:
This PR makes the caching decorator naturally work with multiprocessing without the need for manual cache checking at the expense of removing an additional check for consistency of (keyword) arguments of decorated functions.

Note that this is a different, in my opinion cleaner and more user friendly, implementation than those discussed in #86 and #91.

closes #91

### DIFF
--- a/examples/example_caching.py
+++ b/examples/example_caching.py
@@ -18,14 +18,14 @@ def f_target(x):
 
 @gp.utils.disk_cache("example_caching_cache.pkl")
 def inner_objective(expr):
-    """The caching decorator uses the function parameter to identify
-    identical function calls. As many different genotypes produce
-    the same simplified SymPy expression we can use such
-    expressions as an argument to the decorated function to skip
-    reevaluating functionally identical individuals. Note that
-    caching only makes sense for deterministic objective
-    functions, as it assumes that expressions will always return
-    the same fitness values.
+    """The caching decorator uses the function parameters to identify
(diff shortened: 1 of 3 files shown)


jakobj:
ok, implemented a test for it. code became a bit more complex, have a look pls :)

jakobj:
rebased. and merging.

PR #100, opened 2020-05-05
Fix pretty_str to not cut of inputs/outputs if their count exceed n_rows

### LINKED ISSUE #81: `pretty_str` cuts off input nodes if n_inputs > n_rows
For example in test_node.py::test_add we use two inputs, but only one row. `pretty_str` assumes that we have more rows than inputs and hence does not print the second input node. this should be fixed.

jakobj:
fixes #81

### DIFF
--- a/gp/cartesian_graph.py
+++ b/gp/cartesian_graph.py
@@ -62,7 +62,7 @@ class CartesianGraph:
 
         s = "\n"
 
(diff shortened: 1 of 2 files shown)


jakobj:
wohoooo #100 :100: :tada: :tada:

PR #110, opened 2020-05-12
Replace dictionary in Primitives with tuple and turn Primitives into frozen dataclass

### LINKED ISSUE #106: Better name for `Primitives.sample()` function
The `sample()` method of `Primitives` has an ambiguous name and a misleading docstring because it doesn't sample a primitive, but rather an index that indexes a primitive within the `Primitives` class. It should be renamed.

For example to `sample_index` (@jakobj 's suggestion).

### LINKED ISSUE #107: Use `OrderedDict` to store primitives in consistent order
As mentioned in #64 with regards to `Primitives._primitives`:
> it's important that the order of primitives is exactly conserved from how this object was initialized otherwise the mapping from function id to function changes when an individual is cloned [...]. maybe we should use an OrderedDict [...]?

jakobj:
Instead of using an `OrderedDict` like suggested in #107 this PR replaces the dictionary used in `Primitives` with a list. Since function keys are just integers, they might as well be indices in an array. This automatically avoid a potential relabeling of functions when a `Primitives` object is converted to a list a reinitialized.

Also renames `Primitives.sample` -> `Primitives.sample_allele`.
Also removes some outdated comments.

closes #107 
closes #106

### DIFF
--- a/gp/primitives.py
+++ b/gp/primitives.py
@@ -1,55 +1,52 @@
+from dataclasses import dataclass, field
 import numpy as np
 
-from typing import Iterator, List, Tuple, Type
+from typing import Iterator, Tuple, Type
 
 from .node import Node
 
 
+@dataclass(frozen=True)
 class Primitives:
-    """Class collecting primitives of the Cartesian Genetic Programming framework.
+    """Convenience class to manage primitives, i.e., Node classes.
+
     """
 
-    _max_arity = 0
-    _primitives: dict = {}
+    _primitives: Tuple[Type[Node], ...]
+    _max_arity: int = field(init=False)
 
-    def __init__(self, primitives: List[Type[Node]]) -> None:
-        """Init function.
+    def __post_init__(self):
+        self._check_types()
+        self.__dict__[
+            "_max_arity"
+        ] = self._determine_max_arity()  # avoid using use __setattr_ since dataclass is frozen
 
-        Parameters
-        ----------
-        primitives : List[Type[Node]]
-            List of primitives.
-        """
-        for i in range(len(primitives)):
-            if not isinstance(primitives[i], type):
-                raise TypeError(f"expected class but received {type(primitives[i])}")
-            if not issubclass(primitives[i], Node):
-                raise TypeError(f"expected subclass of Node but received {primitives[i].__name__}")
+    def _check_types(self):
+        if not isinstance(self._primitives, tuple):
+            raise TypeError(f"expected tuple but received {type(self._primitives)}")
 
-        self._primitives = {}
-        for i in range(len(primitives)):
-            self._primitives[i] = primitives[i]
+        for i in range(len(self._primitives)):
+            if not isinstance(self._primitives[i], type):
+                raise TypeError(
+                    f"expected class but received instance of {type(self._primitives[i])}"
+                )
+            if not issubclass(self._primitives[i], Node):
+                raise TypeError(
+                    f"expected subclass of Node but received class {self._primitives[i].__name__}"
+                )
 
-        # hide primitives dict behind MappingProxyType to make sure it
-        # is not changed after construction
-        # unfortunately not supported by pickle, necessary for
-        # multiprocessing; another way to implement this?
-        # self._primitives = types.MappingProxyType(self._primitives)
-
-        self._determine_max_arity()
-
-    def __iter__(self) -> Iterator:
-        return iter([self[i] for i in range(len(self._primitives))])
-
-    def _determine_max_arity(self) -> None:
+    def _determine_max_arity(self) -> int:
 
         arity = 1  # minimal possible arity (output nodes need one input)
 
-        for idx, p in self._primitives.items():
+        for p in self._primitives:
             if arity < p._arity:
                 arity = p._arity
 
-        self._max_arity = arity
+        return arity
+
+    def __iter__(self) -> Iterator[Type[Node]]:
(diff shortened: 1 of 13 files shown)


gp/primitives.py
```diff
@@ -26,33 +23,24 @@ def __init__(self, primitives: List[Type[Node]]) -> None:
...
-        self._primitives = {}
-        for i in range(len(primitives)):
-            self._primitives[i] = primitives[i]
-
-        # hide primitives dict behind MappingProxyType to make sure it
-        # is not changed after construction
-        # unfortunately not supported by pickle, necessary for
-        # multiprocessing; another way to implement this?
-        # self._primitives = types.MappingProxyType(self._primitives)
+        self._primitives = list(primitives)
```
mschmidt87, line 26: If `primitives` is already a list, why calling `list()` here?

gp/primitives.py
```diff
@@ -26,33 +23,24 @@ def __init__(self, primitives: List[Type[Node]]) -> None:
...
+        self._max_arity: int = self._determine_max_arity()
 
-        self._determine_max_arity()
+    def __iter__(self) -> Iterator[Type[Node]]:
+        return iter(p for p in self._primitives)
```
mschmidt87, line 30: why not `iter(self._primitives)`?

gp/primitives.py
```diff
@@ -26,33 +23,24 @@ def __init__(self, primitives: List[Type[Node]]) -> None:
...
-        for i in range(len(primitives)):
-            self._primitives[i] = primitives[i]
-
-        # hide primitives dict behind MappingProxyType to make sure it
-        # is not changed after construction
-        # unfortunately not supported by pickle, necessary for
-        # multiprocessing; another way to implement this?
-        # self._primitives = types.MappingProxyType(self._primitives)
+        self._primitives = list(primitives)
+        self._max_arity: int = self._determine_max_arity()
```
mschmidt87, line 27: Wouldn't this be better written as a class property then?

gp/primitives.py
```diff
@@ -62,20 +50,18 @@ def sample(self, rng: np.random.RandomState) -> int:
...
         int
-            Index of the sample primitive
+            Index of the sampled primitive.
         """
-        return rng.choice(self.alleles)
+        return rng.randint(len(self._primitives))
 
     def __getitem__(self, key: int) -> Type[Node]:
+        if key < 0 or key >= len(self._primitives):
+            raise IndexError("primitives index out of bounds")
```
mschmidt87, line 59: "primitive index out of bounds"

gp/primitives.py
jakobj: to create a shallow copy of a list

gp/primitives.py
jakobj: good point. but even better: we don't need this function any more :)

gp/primitives.py
jakobj: what is "this"? and why?

gp/primitives.py
jakobj: yes!

jakobj:
thanks for the review :sparkles:  addressed all your comments, pls have another look

gp/primitives.py
mschmidt87: Got it.

gp/primitives.py
mschmidt87: `_max_arity`. This seems to me like the classic example where you want to use a class property: When accessing `_max_arity`, you want to call a routine that determines it on the fly. Or am I missing something?

gp/primitives.py
mschmidt87: Hmm, I think we do need it so that `list(Primitives)` works. That's also why `mypy` complains in CI.

gp/primitives.py
jakobj: once initialized, this class is/should be treated as constant, the `max_arity` hence does not change afterwards. to save runtime it hence makes sense to compute it once during initialization

gp/primitives.py
jakobj: it's not needed for `list()` (`__getitem__`) is enough, but i will look into the mypy issue

mschmidt87:
I thought about the problem that an instance of `Primitives` is still mutable, potentially causing a risk in user applications. Here is my suggestion to solve that problem:
1. Instead of using a `list`, use a `tuple` for the `self._primitives` member, because `tuple` is immutable.
2. Turn the `Primitives` class into a `dataclass` and make it frozen, like so:
```python
@dataclass(frozen=True)
class Primitives:
    primitives: tuple(Type[Node])
```
You will need to adapt the class code a little bit because the `__init__` is then created automatically, i.e., everything that is not in `__init__` has to go to a `__post_init__` function. But that should be all and pretty straightforward.

What this will do is it will prevent two ways to manipulate the `_primitives` after instance creation. You can no longer do
```python
prim = Primitives((gp.Add, gp.Sub))
prim._primitives[0] = gp.Mul
```
because tuples are immutable.
And you can no longer do
```python
prim = Primitives((gp.Add, gp.Sub))
prim._primitives = (gp.Mul, gp.Pow)
```
because the members of `Primitive` are frozen.

jakobj:
ok, i've implemented your suggestion for a frozen dataclass. replacing the list with a tuple requires an API change, which subsequently required many changes in the tests. i've also added the additional dependency on "dataclasses" for python 3.6. please have another look @mschmidt87

PR #119, opened 2020-05-21
Move parameter_names_to_values from Individual to Genome

### LINKED ISSUE #97: Move parameter value storage from `Individual` to `Genome`
The output values of `Parameter` nodes are currently stored in each individual (https://github.com/Happy-Algorithms-League/python-gp/blob/master/gp/individual.py#L33). I would argue this is rather surprising since they define the values of node encoded in the `Genome` and like it are passed on to the next generation.
I think it would hence be more natural to store these values within the `Genome`. This would also remove the need to pass around the dictionary from the `Individual` to all `to...` functions (e.g., https://github.com/Happy-Algorithms-League/python-gp/blob/master/gp/individual.py#L123).

jakobj:
As discussed in #97. This dictionary is used to store the values for ParameterNodes appearing in the graph. It is hence closely linked to the specific genome and should hence be managed there.

closes #97

### DIFF
--- a/cgp/cartesian_graph.py
+++ b/cgp/cartesian_graph.py
@@ -17,7 +17,7 @@ try:
 except ModuleNotFoundError:
     torch_available = False
 
-from typing import Callable, Dict, List, Optional, Set
+from typing import Callable, DefaultDict, Dict, List, Optional, Set
 
 from .node import Node, InputNode, OutputNode, Parameter
 from .genome import Genome
@@ -36,14 +36,14 @@ class CartesianGraph:
         genome: Genome
             Genome defining graph connectivity and node operations.
         """
-        self._n_outputs: int
         self._n_inputs: int
+        self._n_outputs: int
         self._n_columns: int
         self._n_rows: int
         self._nodes: List
+        self._parameter_names_to_values: DefaultDict
 
         self.parse_genome(genome)
-        self._genome = genome
 
     def __repr__(self) -> str:
         return "CartesianGraph(" + str(self._nodes) + ")"
@@ -99,12 +99,15 @@ class CartesianGraph:
         if genome.dna is None:
             raise RuntimeError("dna not initialized")
 
-        self._genome = genome
-
         self._n_inputs = genome._n_inputs
         self._n_outputs = genome._n_outputs
         self._n_columns = genome._n_columns
         self._n_rows = genome._n_rows
+
+        # WARNING: creating a reference, not a copy here is essential;
+        # accessing missing elements in this DefaultDict during graph
+        # construction needs to construct keys with default values
+        self._parameter_names_to_values = genome.parameter_names_to_values
 
         self._nodes = []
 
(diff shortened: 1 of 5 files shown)


gp/cartesian_graph.py
```diff
@@ -222,23 +224,19 @@ def _format_output_str_of_all_nodes(self) -> None:
...
                 node.format_output_str(self)
 
-    def _fill_parameter_values(
-        self, func_str: str, parameter_names_to_values: Optional[Dict[str, float]] = None
-    ) -> str:
+    def _fill_parameter_values(self, func_str: str) -> str:
         g = re.findall("<p[0-9]+>", func_str)
         if len(g) != 0:
-            if parameter_names_to_values is None:
+            if self.parameter_names_to_values is None:
```
jakobj, line 230: not sure whether this if branch is needed, will investigate tomorrow

gp/genome.py
```diff
@@ -434,4 +454,35 @@ def clone(self) -> "Genome":
...
+        """
+        for name, value in torch_cls.named_parameters():
+            name = f"<{name[1:]}>"
+            if name in self.parameter_names_to_values:
+                self.parameter_names_to_values[name] = value.item()
+                assert not np.isnan(self.parameter_names_to_values[name])
+
+                # since a parameter of this individual has changed its
+                # fitness needs to be reevaluated
+                self.fitness = None
```
jakobj, line 488: this should be done in the individual not the genome. scary that no test catches this, will investigate tomorrow

gp/genome.py
```diff
@@ -11,6 +19,14 @@
 ID_NON_CODING_GENE: int = -3
 
 
+def return_float_one():
```
mschmidt87, line 22: Needs a return type annotation.

gp/genome.py
mschmidt87: What's the outcome here?

gp/cartesian_graph.py
mschmidt87: what's the outcome here?

jakobj: not needed

gp/genome.py
jakobj: thanks

gp/genome.py
jakobj: this was indeed wrong, fixed it now and added a test

gp/individual.py
```diff
@@ -149,33 +129,14 @@ def to_sympy(self, simplify=True):
...
-            if name in self.parameter_names_to_values:
-                self.parameter_names_to_values[name] = value.item()
-                assert not np.isnan(self.parameter_names_to_values[name])
-
-                # since a parameter of this individual has changed its
-                # fitness needs to be reevaluated
-                self.fitness = None
+        # since a parameter of this individual has changed its
+        # fitness needs to be reevaluated
+        self.fitness = None
```
jakobj, line 139: this is probably suboptimal since fitness only needs to be reevaluated when the value of a parameter which is used in the computational graph has changed, will look into this

jakobj: fixed

jakobj:
thanks for your comments! :) addressed all of them and rebased, pls have another look

PR #123, opened 2020-05-27
Add type hints to Individual class and fix detected issues

mschmidt87:
This PR adds type hints to the `Individual` and `IndividualMultiGenome` classes and fixes the detected issues:

1. `Individual` and `IndividualMultiGenome` class
The relation between the two classes was not good: Having IndividualMultiGenome inherit from Individual causes problems. For instance, the self.genome member is a list of Genome instances in the former, but since it is initialized in the latter as a Genome instance, this does not make much sense. We could refrain from calling the super().__init__() function but then the inheritance does not make any sense. 

*Solution*: As discussed below, I removed the `IndividualMultiGenome` class and make the `self.genome` member of `Individual` a list. This triggered some changes in the `Population` class and the local search algorithms which now needs a list of `genome_params` and in the tests. As a side remark: Some functionality would not have worked for the multi genome version, e.g. the local search. This PR removes this problem.

2. The `Individual` class could be initialized with `genome=None` which is dangerous because all its methods rely on a genome being defined. I removed this option therefore and adapted the code in the `Population` class which had used this feature. 

3. `Individual.fitness` can be `None`
This is dangerous because some of the operations on the fitness relied on the fitness being a float without explicitly checking or catching exceptions. I remedied this in this PR.

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -1,9 +1,9 @@
 import concurrent.futures
 import numpy as np
 
-from typing import Callable, List
+from typing import Callable, List, Tuple
 
-from ..individual import Individual
+from ..individual import IndividualBase
 from ..population import Population
 
 
@@ -24,7 +24,7 @@ class MuPlusLambda:
         tournament_size: int,
         *,
         n_processes: int = 1,
-        local_search: Callable[[Individual], None] = lambda combined: None
+        local_search: Callable[[IndividualBase], None] = lambda combined: None,
     ):
         """Init function
 
@@ -59,7 +59,7 @@ class MuPlusLambda:
         self.local_search = local_search
 
     def initialize_fitness_parents(
-        self, pop: Population, objective: Callable[[Individual], Individual]
+        self, pop: Population, objective: Callable[[IndividualBase], IndividualBase]
     ) -> None:
         """Initialize the fitness of all parents in the given population.
 
@@ -67,25 +67,27 @@ class MuPlusLambda:
         ----------
         pop : Population
             Population instance.
-        objective : Callable[[gp.Individual], gp.Individual]
+        objective : Callable[[gp.IndividualBase], gp.IndividualBase]
             An objective function used for the evolution. Needs to take an
-            individual (Individual) as input parameter and return
+            individual (IndividualBase) as input parameter and return
             a modified individual (with updated fitness).
         """
         # TODO can we avoid this function? how should a population be
         # initialized?
         pop._parents = self._compute_fitness(pop.parents, objective)
 
-    def step(self, pop: Population, objective: Callable[[Individual], Individual]) -> Population:
+    def step(
+        self, pop: Population, objective: Callable[[IndividualBase], IndividualBase]
+    ) -> Population:
         """Perform one step in the evolution.
 
         Parameters
         ----------
         pop : Population
             Population instance.
-        objective : Callable[[gp.Individual], gp.Individual]
+        objective : Callable[[gp.IndividualBase], gp.IndividualBase]
             An objective function used for the evolution. Needs to take an
-            individual (Individual) as input parameter and return
+            individual (IndividualBase) as input parameter and return
             a modified individual (with updated fitness).
 
         Returns
@@ -108,7 +110,6 @@ class MuPlusLambda:
 
         for ind in combined:
             self.local_search(ind)
-
         combined = self._compute_fitness(combined, objective)
 
         combined = self._sort(combined)
@@ -117,10 +118,10 @@ class MuPlusLambda:
 
         return pop
 
-    def _create_new_offspring_generation(self, pop: Population) -> List[Individual]:
+    def _create_new_offspring_generation(self, pop: Population) -> List[IndividualBase]:
         # fill breeding pool via tournament selection from parent
         # population
-        breeding_pool: List[Individual] = []
+        breeding_pool: List[IndividualBase] = []
         while len(breeding_pool) < self.n_breeding:
             tournament_pool = pop.rng.permutation(pop.parents)[: self.tournament_size]
             best_in_tournament = sorted(tournament_pool, key=lambda x: -x.fitness)[0]
@@ -137,8 +138,8 @@ class MuPlusLambda:
         return offsprings
 
     def _compute_fitness(
-        self, combined: List[Individual], objective: Callable[[Individual], Individual],
-    ) -> List[Individual]:
+        self, combined: List[IndividualBase], objective: Callable[[IndividualBase], IndividualBase]
+    ) -> List[IndividualBase]:
         # computes fitness on all individuals, objective functions
         # should return immediately if fitness is not None
         if self.n_processes == 1:
@@ -149,7 +150,7 @@ class MuPlusLambda:
 
         return combined
 
-    def _sort(self, combined: List[Individual]) -> List[Individual]:
+    def _sort(self, combined: List[IndividualBase]) -> List[IndividualBase]:
         # create copy of population
         combined_copy = [ind.clone() for ind in combined]
 
@@ -159,20 +160,31 @@ class MuPlusLambda:
             if np.isnan(ind.fitness):
                 ind.fitness = -np.inf
 
-        # get list of indices that sorts combined_copy ("argsort")
+        def sort_func(zipped_ind: Tuple[int, IndividualBase]) -> float:
+            """Return fitness of an individual or raise error if it is None.
+            """
+            _, ind = zipped_ind
+            if isinstance(ind.fitness, float):
+                return ind.fitness
+            else:
+                raise ValueError(
+                    f"IndividualBase fitness value is of wrong type {type(ind.fitness)}."
+                )
+
+        # get list of indices that sorts combined_copy ("argsort") in descending order
         combined_sorted_indices = [
-            idx for (idx, _) in sorted(enumerate(combined_copy), key=lambda x: -x[1].fitness)
+            idx for (idx, _) in sorted(enumerate(combined_copy), key=sort_func, reverse=True)
         ]
 
-        # return sorted original list of individuals
+        # return original list of individuals sorted in descending order
         return [combined[idx] for idx in combined_sorted_indices]
 
     def _create_new_parent_population(
-        self, n_parents: int, combined: List[Individual]
-    ) -> List[Individual]:
+        self, n_parents: int, combined: List[IndividualBase]
+    ) -> List[IndividualBase]:
         # create new parent population by picking the `n_parents` individuals
         # with the highest fitness
-        parents: List[Individual] = []
+        parents: List[IndividualBase] = []
         for i in range(n_parents):
             # note: unclear whether clone() is needed here; using
             # clone() avoids accidentally sharing references across
--- a/cgp/hl_api.py
+++ b/cgp/hl_api.py
@@ -2,13 +2,13 @@ import numpy as np
 from typing import Optional, Callable
 
 from .population import Population
-from .individual import Individual
+from .individual import IndividualBase
 from .ea import MuPlusLambda
 
 
 def evolve(
     pop: Population,
-    objective: Callable[[Individual], Individual],
+    objective: Callable[[IndividualBase], IndividualBase],
     ea: MuPlusLambda,
     max_generations: int,
     min_fitness: float,
@@ -63,6 +63,7 @@ def evolve(
         # progress printing, recording, checking exit condition etc.; needs to
         # be done /after/ new parent population was populated from combined
         # population and /before/ new individuals are created as offsprings
+        assert isinstance(pop.champion.fitness, float)
         if pop.champion.fitness > max_fitness:
             max_fitness = pop.champion.fitness
 
--- a/cgp/individual.py
+++ b/cgp/individual.py
@@ -1,53 +1,80 @@
+import numpy as np
+from typing import Callable, List, Union
+
+try:
+    import sympy  # noqa: F401
+    from sympy.core import expr as sympy_expr  # noqa: F401
+
+    sympy_available = True
+except ModuleNotFoundError:
+    sympy_available = False
+
+try:
+    import torch  # noqa: F401
+
+    torch_available = True
+except ModuleNotFoundError:
+    torch_available = False
+
 from .genome import Genome
 from .cartesian_graph import CartesianGraph
 
 
-class Individual:
-    """An individual representing a particular computational graph.
+class IndividualBase:
+    """Base class for all individuals.
     """
 
-    def __init__(self, fitness, genome):
+    def __init__(self, fitness: Union[float, None]) -> None:
         """Init function.
 
         fitness : float
             Fitness of the individual.
-        genome: Genome instance
-            Genome of the individual.
         """
-        self.fitness = fitness
-        self.genome = genome
-        self.idx = None  # an identifier to keep track of all unique genomes
-
-    def __repr__(self):
-        return f"Individual(idx={self.idx}, fitness={self.fitness}, genome={self.genome}))"
+        self.fitness: Union[float, None] = fitness
+        self.idx: int
 
     def clone(self):
-        """Clone the individual.
+        raise NotImplementedError()
 
-        Returns
-        -------
-        Individual
-        """
-        return Individual(self.fitness, self.genome.clone())
+    def mutate(self, mutation_rate, rng):
+        raise NotImplementedError()
 
-    def crossover(self, other_parent, rng):
-        """Create a new individual by cross over with another individual.
+    def randomize_genome(self, rng):
+        raise NotImplementedError()
+
+    def to_func(self):
+        raise NotImplementedError()
+
+    def to_numpy(self):
+        raise NotImplementedError()
+
+    def to_torch(self):
+        raise NotImplementedError()
+
+    def to_sympy(self, simplify):
+        raise NotImplementedError()
+
+    def update_parameters_from_torch_class(self, torch_cls):
+        raise NotImplementedError()
+
+    @staticmethod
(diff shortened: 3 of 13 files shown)


mschmidt87:
As a result, I detected the following issues with the current code:

- Having `IndividualMultiGenome` inherit from `Individual` causes problems. For instance, the `self.genome` member is a list of `Genome` instances in the former, but since it is initialized in the latter as a `Genome` instance, this does not make much sense. We could refrain from calling the `super().__init__()` function but then the inheritance does not make any sense. There are two possible solutions:
-- Make both independent classes and implement the common functions as external functions which are used by both classes. I implemented this for now.
-- Write an `IndividualBase` class and let the other two classes inherit from it. This would probably the cleanest solution

- The `Individual` class can currently be initialized with `genome=None` (happens in the `Population` class, see the diff), but this does not make sense, because the methods of `Individual` are unusable. That is why whenever it is initialized with `genome=None`, the randomize method is immediately called. `mypy` complains about this pattern which makes sense because this is ugly and dangerous. I would therefore propose the implemented solution. However, then we need to make a decision which rng we pass to the `randomize_genome` method if it is called in the `__init__` of `Individual`. All of this also extends to `IndividualMultiGenome`.

- The fitness value of individuals is now initialized as `None` and only set once it is computed. While this makes sense in principle, it causes many problems because some parts of the code rely on the fitness being a float (or an int), e.g. when we order individuals by their fitness.

FYI, this is the current outcome of `mypy`:
```
cgp/individual.py:109: error: Too few arguments for "randomize_genome" of "Individual"
cgp/population.py:130: error: Unsupported operand type for unary - ("Optional[float]")
cgp/ea/mu_plus_lambda.py:164: error: Unsupported operand type for unary - ("Optional[float]")
cgp/hl_api.py:79: error: Unsupported operand types for + ("None" and "float")
cgp/hl_api.py:79: note: Left operand is of type "Optional[float]"
```
The first line is related to the 2nd issue. The other ones are related to the third issue.
I'd like your feedback on all 3 issues @jakobj .

Also note that many tests fail at the moment, I will adapt them once we settled the issues above.

jakobj:
thanks for adding this! regarding your points:
1. what about replacing the `Individual` class with the `IndividualMultiGenome` class + some specific handling of the case where the `genome` argument is a dict, not a list of dicts? this seems to me the most natural solution as `Individual` is just a specific form of `IndividualMultiGenome`. otherwise i think i'm leaning towards the idea of having an `AbstractIndividual` base class that implements the shared functionality. both concrete classes then call these functions as required.
2. what about generating a `Genome` in the population class, then passing this as an argument to `Individual.__init__`? this seems natural to me as I see `Individual` just as a shallow wrapper around `Genome` anyway -- something that we could make even more explicit if you agree with this view. we would hence avoid the need to support `genome=None` in `Individual.__init__`
3. we use `fitness=None` as a signal to trigger evaluation of the individual on the objective, for example after non-silent mutation the fitness is set to `None`. I'm not sure how to implement this differently, but I'm open for ideas!

who knew such issues are lurking beneath such an innocent-sounding PR ;) mypy FTW!!!11

mschmidt87:
> thanks for adding this! regarding your points:
> 
>     1. what about replacing the `Individual` class with the `IndividualMultiGenome` class + some specific handling of the case where the `genome` argument is a dict, not a list of dicts? this seems to me the most natural solution as `Individual` is just a specific form of `IndividualMultiGenome`. otherwise i think i'm leaning towards the idea of having an `AbstractIndividual` base class that implements the shared functionality. both concrete classes then call these functions as required.
> 
If we go for this, then I would not allow a single genome. Rather, an individual with a single genome has a list of genomes with only one entry. Then we could even call that class `Individual`. That means, in the current code we just make `Individual.genome` a list. 

>     2. what about generating a `Genome` in the population class, then passing this as an argument to `Individual.__init__`? this seems natural to me as I see `Individual` just as a shallow wrapper around `Genome` anyway -- something that we could make even more explicit if you agree with this view. we would hence avoid the need to support `genome=None` in `Individual.__init__`
> 
That would probably work, yes. I'll implement that.

>     3. we use `fitness=None` as a signal to trigger evaluation of the individual on the objective, for example after non-silent mutation the fitness is set to `None`. I'm not sure how to implement this differently, but I'm open for ideas!
> 

Yes, the `None` does make sense. But we need to handle the case that the fitness is `None` everywhere the fitness is actually used. `mypy` detected the four cases above. That will make the code a bit more verbose and lengthy. For instance, we could then check manually whether the fitness is `None` and throw an error if it is. What do you think?

jakobj:
> If we go for this, then I would not allow a single genome. Rather, an individual with a single genome has a list of genomes with only one entry. Then we could even call that class `Individual`. That means, in the current code we just make `Individual.genome` a list.

sounds good to me!

> Yes, the `None` does make sense. But we need to handle the case that the fitness is `None` everywhere the fitness is actually used. `mypy` detected the four cases above. That will make the code a bit more verbose and lengthy. For instance, we could then check manually whether the fitness is `None` and throw an error if it is. What do you think?

I think it's a good idea to add these checks as they provide an additional layer of safety against strange things happening without an error being raised, even if that means more loc

mschmidt87:
I addressed all the issues and documented the solutions in the PR description above. Please take another look @jakobj

cgp/individual.py
```diff
@@ -46,19 +89,7 @@ def crossover(self, other_parent, rng):
...
-        graph = CartesianGraph(genome)
-        active_regions = graph.determine_active_regions()
-        only_silent_mutations = genome.mutate(n_mutations, active_regions, rng)
-
-        if not only_silent_mutations:
-            self.fitness = None
-
-    def mutate(self, mutation_rate, rng):
+    def mutate(self, mutation_rate: float, rng: np.random.RandomState) -> None:
         """Mutate the individual in place.
```
jakobj, line 93:
```suggestion
        """Mutate the individual.
```

cgp/individual.py
```diff
@@ -90,89 +124,51 @@ def randomize_genome(self, genome_params, rng):
...
+    def to_func(self) -> List[Callable[[List[float]], List[float]]]:
         """Return the expression represented by the individual as Callable.
 
         Returns
         ----------
-        Callable
+        Callable[[List[float]], List[float]]
         """
-        return CartesianGraph(self.genome).to_func()
+        return [CartesianGraph(g).to_func() for g in self.genome]
```
jakobj, line 137: i feel like this will add a lot of unpacking for the user, won't it? it's already suboptimal that they will need to do `f([...])[0]` for a function with a single output, now it would read `f[0]([...])[0]`. this is not very pretty. should we add automatic unpacking for the special case of a single genome, even if this seems to contradict best practice? this holds essentially for all cases where we add an additional list in this new implementation

cgp/local_search/gradient_based.py
```diff
@@ -57,20 +57,25 @@ def gradient_based(
 
     f = individual.to_torch()
 
-    if len(list(f.parameters())) > 0:
-        optimizer = optimizer_class(f.parameters(), lr=lr)
+    params = []
+    for torch_mod in f:
+        params += list(torch_mod.parameters())
+    if len(params) > 0:
+        optimizer = optimizer_class(params, lr=lr)  #
```
jakobj, line 64:
```suggestion
        optimizer = optimizer_class(params, lr=lr)
```

cgp/population.py
```diff
@@ -1,8 +1,8 @@
 import numpy as np
 
-from typing import List
-
-from .individual import Individual, IndividualMultiGenome
+from typing import List, Union
+from .genome import Genome
```
jakobj, line 4:
```suggestion

from .genome import Genome
```

cgp/population.py
```diff
@@ -77,17 +77,11 @@ def _label_new_individuals(self, individuals: List[Individual]) -> None:
...
-            if isinstance(self._genome_params, dict):
-                individual = Individual(fitness=None, genome=None)
-            elif isinstance(self._genome_params, list) and isinstance(
-                self._genome_params[0], dict
-            ):
-                individual = IndividualMultiGenome(fitness=None, genome=None)
-            else:
-                raise NotImplementedError()
-            individual.randomize_genome(self._genome_params, self.rng)
+            genome = [Genome(**gd) for gd in self._genome_params]
```
jakobj, line 80: maybe use `genomes` as variable name whereever this is a list?

cgp/population.py
```diff
@@ -123,7 +117,14 @@ def crossover(self, breeding_pool: List[Individual], n_offsprings: int) -> List[
         # Conference on Genetic and Evolutionary Computation-Volume 2,
         # pages 1135–1142. Morgan Kaufmann Publishers Inc.
         assert len(breeding_pool) >= n_offsprings
-        return sorted(breeding_pool, key=lambda x: -x.fitness)[:n_offsprings]
+
+        def sort_func(ind: Individual) -> float:
+            if isinstance(ind.fitness, float):
+                return -ind.fitness
```
jakobj, line 123: same as above, use `reversed` instead

cgp/individual.py
mschmidt87: In place is correct, isn't it? The function does create a new individual.

jakobj: yes, technically it's correct, but isn't it a bit redundant to state for every method that/whether it operates on the corresponding class? in contrast, if this would be a regular function that receives an individual, this additional info would be very helpful!

cgp/individual.py
```diff
@@ -1,43 +1,84 @@
...
+
+    return only_silent_mutations
 
-    def __init__(self, fitness, genome):
+
+class IndividualBase:
+    """Base class for all individuals.
+    """
+
+    def __init__(self, fitness: Union[float, None], genome: Union[Genome, List[Genome]]) -> None:
```
jakobj, line 52: to avoid the unions for genome, wouldn't it be cleaner to only define this member in the derived classes?

cgp/individual.py
```diff
@@ -1,43 +1,84 @@
...
-    def clone(self):
+    def clone(self) -> "IndividualBase":
         """Clone the individual.
 
         Returns
         -------
         Individual
         """
-        return Individual(self.fitness, self.genome.clone())
+        return type(self)(self.fitness, self.genome)
```
jakobj, line 72:
```suggestion
        raise NotImplementedError()
```

jakobj: it does not make sense to use an instance of `IndividualBase`, does it? so i would just not provide any implementation for functions that are overridden in derived classes

cgp/individual.py
```diff
@@ -46,35 +87,23 @@ def crossover(self, other_parent, rng):
...
+            Proportion of genes to be mutated, between 0 and 1.
         rng : numpy.RandomState
             Random number generator instance to use for crossover.
 
         Returns
         -------
         None
         """
-        self._mutate(self.genome, mutation_rate, rng)
+        pass
```
jakobj, line 104:
```suggestion
        raise NotImplementedError()
```

cgp/individual.py
mschmidt87: the idea is that we don't need to define the `clone` function in the child classes.

cgp/individual.py
mschmidt87: this is needed by the `clone` function.

cgp/ea/mu_plus_lambda.py
```diff
@@ -159,20 +160,29 @@ def _sort(self, combined: List[Individual]) -> List[Individual]:
             if np.isnan(ind.fitness):
                 ind.fitness = -np.inf
 
-        # get list of indices that sorts combined_copy ("argsort")
+        # get list of indices that sorts combined_copy ("argsort") in descending order
```
jakobj, line 163: i would move this comment below. mentally i tend to skip functions until i see where there are used and that's where i would need the information that line 173 does argsort

cgp/local_search/gradient_based.py
```diff
@@ -57,20 +57,33 @@ def gradient_based(
 
     f = individual.to_torch()
 
-    if len(list(f.parameters())) > 0:
-        optimizer = optimizer_class(f.parameters(), lr=lr)
+    if isinstance(f, List):
```
jakobj, line 60:
```suggestion
    if isinstance(f, list):
```

cgp/population.py
```diff
@@ -48,61 +53,67 @@ def __init__(
...
-                individual = Individual(fitness=None, genome=None)
-            elif isinstance(self._genome_params, list) and isinstance(
-                self._genome_params[0], dict
-            ):
-                individual = IndividualMultiGenome(fitness=None, genome=None)
+                genome: Genome = Genome(**self._genome_params)
+                genome.randomize(self.rng)
+                individual_s = IndividualSingleGenome(
+                    fitness=None, genome=genome
+                )  # type: IndividualBase
```
jakobj, line 90:
```suggestion
                )
```

jakobj: or what is this needed for?

cgp/population.py
```diff
@@ -48,61 +53,67 @@ def __init__(
...
-                raise NotImplementedError()
-            individual.randomize_genome(self._genome_params, self.rng)
-            individuals.append(individual)
-
+                genomes: List[Genome] = [Genome(**gd) for gd in self._genome_params]
+                for gen in genomes:
+                    gen.randomize(self.rng)
+                individual_m = IndividualMultiGenome(
+                    fitness=None, genomes=genomes
+                )  # type: IndividualBase
```
jakobj, line 98:
```suggestion
                )
```

cgp/population.py
mschmidt87: Done. And added a comment.

cgp/population.py
mschmidt87: It is needed by mypy. I added an explanatory comment behind this.

cgp/population.py
mschmidt87: see above.

cgp/local_search/gradient_based.py
mschmidt87: Done.

mschmidt87:
Addressed all your comments, @jakobj . One more point: I would rename `IndividualMultiGenome.genomes` to `genome`. It might be orthographically more correct to say `genomes`, but it makes it more difficult to handle for the user if they want to switch from SingleGenome to MultiGenome. It also introduces more complex code to handle the different variable names in the library and in the tests.

cgp/ea/mu_plus_lambda.py
mschmidt87: Done

cgp/local_search/gradient_based.py
```diff
@@ -57,20 +57,33 @@ def gradient_based(
 
     f = individual.to_torch()
 
-    if len(list(f.parameters())) > 0:
-        optimizer = optimizer_class(f.parameters(), lr=lr)
+    if isinstance(f, list):
```
jakobj, line 60: this stuff makes me flinch every time i see it, but i don't think we have an easier solution. oh well :|

cgp/population.py
```diff
@@ -48,61 +53,71 @@ def __init__(
...
+                # IndividualBase
+                individuals.append(individual_s)
             else:
-                raise NotImplementedError()
-            individual.randomize_genome(self._genome_params, self.rng)
-            individuals.append(individual)
-
+                genomes: List[Genome] = [Genome(**gd) for gd in self._genome_params]
+                for gen in genomes:
+                    gen.randomize(self.rng)
```
jakobj, line 97:
```suggestion
                for g in genomes:
                    g.randomize(self.rng)
```

cgp/population.py
```diff
@@ -48,61 +53,71 @@ def __init__(
...
+                individual_m = IndividualMultiGenome(
+                    fitness=None, genomes=genomes
+                )  # type: IndividualBase # indicates to mypy that
+                # individual_s is an instance of a child class of
```
jakobj, line 101:
```suggestion
                # individual_m is an instance of a child class of
```

cgp/hl_api.py
```diff
@@ -63,6 +63,7 @@ def evolve(
         # progress printing, recording, checking exit condition etc.; needs to
         # be done /after/ new parent population was populated from combined
         # population and /before/ new individuals are created as offsprings
+        assert isinstance(pop.champion.fitness, float)
```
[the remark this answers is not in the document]
mschmidt87: Yes, necessary.

cgp/individual.py
```diff
@@ -1,178 +1,223 @@
...
+    @staticmethod
+    def _to_sympy(genome: Genome, simplify) -> "sympy_expr.Expr":
+        return CartesianGraph(genome).to_sympy(simplify)
 
-        Returns
-        ----------
-        Callable
+    @staticmethod
+    def _update_parameters_from_torch_class(genome: Genome, torch_cls: "torch.nn.Module") -> bool:
+        """Update parameters of the individual from a torch class.
```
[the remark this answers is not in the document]
mschmidt87: done.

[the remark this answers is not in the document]
mschmidt87: done.

cgp/hl_api.py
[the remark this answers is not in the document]
mschmidt87: necesary.

mschmidt87:
Addressed all comments. @jakobj

test/test_hl_api.py
```diff
@@ -168,7 +166,7 @@ def test_speedup_parallel_evolve(population_params, genome_params, ea_params):
     # Serial execution
 
     for n_processes in [1, 2, 4]:
-        pop = cgp.Population(**population_params, genome_params=genome_params)
+        pop = cgp.Population(**population_params, genome_params=[genome_params])
```
jakobj, line 169: why does this need to be a list of params?

mschmidt87: It doesn't. I reverted the change.

jakobj:
excellent work :sparkles: :tada: :sparkles:  and merging

PR #126, opened 2020-05-29
Move all shared functionality from IndividualSingle/MultiGenome to base class

jakobj:
as discussed on slack, this moves all functionality from the derived classes into the base class where static methods which operate on `Genome` are defined. might seem to be too many indirections, e.g., `to_func -> _to_func -> to_func` but this makes the derived classes trivial to understand as they only call the static methods on each of their `Genome` members. note that here `__repr__` is missing. `crossover` was intentionally deleted to remove dead code.

### DIFF
--- a/cgp/individual.py
+++ b/cgp/individual.py
@@ -20,98 +20,54 @@ from .genome import Genome
 from .cartesian_graph import CartesianGraph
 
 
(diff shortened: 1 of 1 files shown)


jakobj:
oops i really seem to have forgotten `clone`, will fix. we do have docstrings for all the `to_` functions in `CartesianGraph`, in general i don't see why we should reproduce them here. but we could maybe add a "see CartesianGraph.to_" to each of them?

PR #136, opened 2020-06-08
Fix examples

### LINKED ISSUE #121: Fix examples
Due to API changes the examples are currently broken and should be fixed.

jakobj:
i've adapted the mutation rate to more reasonable values (0.5 -> 0.05). also needed to increase the number of cols in `example_evo_regression.py` to make it converge in a reasonable number of generations.

fixes #121

### DIFF
--- a/examples/example_caching.py
+++ b/examples/example_caching.py
@@ -49,7 +49,7 @@ def objective(individual):
 
 def evolution():
     params = {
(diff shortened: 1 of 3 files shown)


jakobj:
hmm, this is strange, i can not reproduce this. have you tried deleting the cache file before executing the example?

PR #139, opened 2020-06-09
Avoid catching of SymPy expression

jakobj:
These kind of problems should not be caught but propagated to the user who will know how to deal with them appropriately. This specific case is even worse: since the `InvalidSympyExpression` is only defined locally in `_validate_sympy_expr`, we can't even catch that specific expression outside on the user level, forcing us to use a catch all.

### DIFF
--- a/test/test_cartesian_graph.py
+++ b/test/test_cartesian_graph.py
@@ -383,7 +383,7 @@ def test_to_sympy():
         )
 
 
-def test_catch_invalid_sympy_expr():
+def test_allow_sympy_expr_with_infinities():
     pytest.importorskip("sympy")
 
     primitives = (cgp.Sub, cgp.Div)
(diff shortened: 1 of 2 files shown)


test/test_cartesian_graph.py
```diff
@@ -383,7 +383,7 @@ def test_to_sympy():
         )
 
 
-def test_catch_invalid_sympy_expr():
+def test_handle_sympy_expr_with_infinities():
```
mschmidt87, line 386: Perhaps a more descriptive name. With your change, infitinies are *not* handled, right?

mschmidt87, line 386:
```suggestion
def test_let_sympy_expr_with_infinities_pass():
```

jakobj: good point. tried something new.

jakobj:
thanks for the quick review :1st_place_medal: and merging

PR #140, opened 2020-06-10
Fix bug in IndividualMultiGenome and add test

### LINKED ISSUE #127: Add tests for local search on individuals with multiple genomes
The work on #123 revealed that currently, local search does not work for individuals with multiple genomes. #123 will fix this, but we have to add explicit tests for this, which is outside the scope of #123 .

mschmidt87:
This PR adds a test for local search with `IndividualMultiGenome` and fixes but in the `update_parameters_from_torch_class` function.

The bug was tricky: We called `any()` on a generator. However, it turns out that if do that, then the iteration stops once the condition inside `any()` has been evaluated to `True` once. That makes sense since `any()` will return `True` if any of the conditions is `True`, i.e., there is no point in continuing after finding one evaluation to `True`. 
Obviously, this is not what we want here, because all the genomes should update their parameters. Thus, I turned the generator into a list comprehension to make sure that all genomes are updated.

The test extends the test for `IndividualSingleGenome` by using two genomes and defining the loss function such that the second genome is counted with a weighting factor of twice.

Fixes #127 .

### DIFF
--- a/cgp/individual.py
+++ b/cgp/individual.py
@@ -211,9 +211,10 @@ class IndividualMultiGenome(IndividualBase):
 
     def update_parameters_from_torch_class(self, torch_cls: List["torch.nn.Module"]) -> None:
         any_parameter_updated = any(
-            self._update_parameters_from_torch_class(g, tcls)
-            for g, tcls in zip(self.genome, torch_cls)
+            [
+                self._update_parameters_from_torch_class(g, tcls)
+                for g, tcls in zip(self.genome, torch_cls)
+            ]
         )
-
         if any_parameter_updated:
             self.fitness = None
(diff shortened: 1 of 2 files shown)


cgp/individual.py
```diff
@@ -211,9 +211,10 @@ def to_sympy(self, simplify: bool = True) -> List["sympy_expr.Expr"]:
 
     def update_parameters_from_torch_class(self, torch_cls: List["torch.nn.Module"]) -> None:
         any_parameter_updated = any(
-            self._update_parameters_from_torch_class(g, tcls)
-            for g, tcls in zip(self.genome, torch_cls)
+            [
```
jakobj, line 214: wow, good catch, this was a sneaky one!

jakobj: actually, maybe a comment would help?

PR #152, opened 2020-06-16
Change mode of installation in travis CI

### LINKED ISSUE #151: Use local checkout in travis CI
Instead of installing the library to the conda environment in Travis CI, I suggest to use the local source code, i.e. do `pip install -e .` instead of `pip install .`. 

If we do this, we can better check code coverage on coveralls. Currently, we don't get a visual overview of coverage because coveralls cannot find the source code files. See here for instance: https://coveralls.io/builds/31473347/source?filename=/home/travis/miniconda3/envs/myenv/lib/python3.8/site-packages/cgp/individual.py

mschmidt87:
Fixes #151

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,11 +27,11 @@ install:
   - pip install pytest pytest-cov coveralls
   - pip install flake8 black
   - pip install mypy
-  - pip install .$DEP
+  - pip install -e .$DEP
 script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
-  - cd test && pytest --cov=cgp
+  - pytest --cov=cgp
 after_success:
   - coveralls


.travis.yml
```diff
@@ -27,11 +27,11 @@ install:
...
   - pip install flake8 black
   - pip install mypy
-  - pip install .$DEP
+  - pip install -e .$DEP
 script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
-  - cd test && pytest --cov=cgp
+  - pytest test --cov=cgp
```
jakobj, line 35: do you need the `test` argument here? shouldn't pytest automatically discover the tests there?

mschmidt87: no, not necessary, removed, thanks.

PR #154, opened 2020-06-17
Add examples to CI

### LINKED ISSUE #122: Add examples to CI
To make sure examples are always up-to-date and not broken, for example, by API changes, we should add them to our continuous integration setup.

mschmidt87:
This PR adds the examples to the CI. Note that the examples require the extra requirements, therefore we cannot run the examples in all jobs.

Discussion point: Perhaps it's sufficient to run the examples in one Python version?

Fixes #122

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,5 +33,10 @@ script:
   - flake8 --config=.flake8 .
   - mypy cgp
   - pytest --cov=cgp
+  - if [ "$DEP" = "[all]" -a $TRAVIS_PYTHON_VERSION = 3.8 ]; then
+       python examples/example_evo_regression.py || exit 1 ;
+       python examples/example_differential_evo_regression.py || exit 1;
+       python examples/example_caching.py || exit 1;
+    fi
 after_success:
   - coveralls


.travis.yml
```diff
@@ -33,5 +33,11 @@ script:
   - flake8 --config=.flake8 .
   - mypy cgp
   - cd test && pytest --cov=cgp
+  - if [ "$DEP" = "[all]" ]; then
+       python examples/example_evo_regression.py;
+       python examples/example_differential_evo_regression.py;
+       python examples/example_caching.py
+    fi
+>>>>>>> Stashed changes
```
jakobj, line 41: this seems like it shouldn't be here ;)

mschmidt87: removed.

mschmidt87:
Thanks for the feedback, all done, @jakobj

mschmidt87:
Thanks for the feedback, I checked and had to add some extra bash code to each line, because otherwise the `if` statement would eat the exit 1 code.

Regarding splitting up the examples in the travis file: I would object against that because it makes the travis file messy, especially if we keep adding examples, and one can easily see the cause of the error by checking out the logs (which you anyway have to do in case of failure).

PR #155, opened 2020-06-18
Enh/custom parameter init

### LINKED ISSUE #102: Provide convenient way of choosing default values for `Parameter` nodes
Currently all `Parameter` nodes are initialized to the same value 1.0 (https://github.com/Happy-Algorithms-League/python-gp/blob/master/gp/individual.py#L33). We should provide a convenient way of choosing this default value.

jakobj:
As mentioned in #102 the initial values for `Parameter` nodes were hardcoded in the default function of the `DefaultDict` storing parameter values. This PR introduces a flexible initialization scheme: users can derive new classes from the `Parameter` node which override the `initial_value` function. This function can return arbitrary fixed values, but also random samples from some specified distribution.

This PR also changes access to `Genome.dna` to always use the getter/setter functions. This way we can make sure the dna is always in a valid state (via `validate_dna`) without having to perform the validation anywhere else.

closes #102

### DIFF
--- a/cgp/cartesian_graph.py
+++ b/cgp/cartesian_graph.py
@@ -1,4 +1,5 @@
 import collections
+import copy
 import numpy as np  # noqa: F401
 import re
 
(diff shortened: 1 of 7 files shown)


jakobj:
thanks for the review :sparkles: i think checking for the correct distribution is a bit overkill, and i agree that the changes to access of the dna should've been a separate PR. i've updated the PR description accordingly. and merging.

PR #157, opened 2020-06-24
Fix counting of mutations in genome.mutate

### LINKED ISSUE #156: Broken counter for successful mutations in `Genome.mutate()`
I believe there is a problem with counting the mutations in `Genome.mutate() `
The counter `successful_mutations` increases only for mutations in the output region (see genome.py L400, L395: https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/genome.py#L400). This causes the mutation_factor to be multiplied by a factor ≃ n_hidden_genes / n_output_genes (a bit lower in practice, since some hidden genes might be mutated several times)

Suggestion 1: add `successful_mutations += 1` on line 400. But this causes the new issue, that also mutations that result in the same gene value are counted as successful mutations.
Suggestion 2: add a second output to `_mutate_hidden_region` that indicates whether there was a real change in the value of the gene (success?) and count only if a real mutation happened. 

(PS. point 2 might also be necessary in `_mutate_output_region` as additional control for a "real" mutation)

HenrikMettler:
This PR fixes the mutation counter by explicitly comparing the number of differences in dna before and after mutations

fixes #156

### DIFF
--- a/cgp/genome.py
+++ b/cgp/genome.py
@@ -361,6 +361,10 @@ class Genome:
         bool
             True if only inactive regions of the genome were mutated, False otherwise.
         """
+
+        def count_dna_differences(dna0: List[int], dna1: List[int]) -> int:
+            return len([1 for gene0, gene1 in zip(dna0, dna1) if gene0 != gene1])
+
         n_mutations = int(mutation_rate * len(self.dna))
         assert n_mutations > 0
 
@@ -369,9 +373,8 @@ class Genome:
 
         dna = list(self._dna)
 
-        successful_mutations = 0
         only_silent_mutations = True
-        while successful_mutations < n_mutations:
+        while count_dna_differences(self.dna, dna) < n_mutations:
 
             gene_idx = rng.randint(0, self._n_genes)
             region_idx = gene_idx // self._length_per_region
@@ -380,31 +383,30 @@ class Genome:
                 continue  # nothing to do here
 
             elif self._is_output_region(region_idx):
-                success = self._mutate_output_region(dna, gene_idx, rng)
-                if success:
-                    silent = False
-                    only_silent_mutations = only_silent_mutations and silent
-                    successful_mutations += 1
+                silent = self._mutate_output_region(dna, gene_idx, rng)
 
             elif self._is_hidden_region(region_idx):
                 silent = self._mutate_hidden_region(dna, gene_idx, active_regions, rng)
-                only_silent_mutations = only_silent_mutations and silent
 
             else:
                 assert False  # should never be reached
 
+            only_silent_mutations = only_silent_mutations and silent
+
         self.dna = dna
         return only_silent_mutations
 
-    def _mutate_output_region(self, dna: List[int], gene_idx: int, rng: np.random.RandomState):
+    def _mutate_output_region(
+        self, dna: List[int], gene_idx: int, rng: np.random.RandomState
+    ) -> bool:
         assert self._is_gene_in_output_region(gene_idx)
 
         if not self._is_function_gene(gene_idx) and self._is_active_input_gene(gene_idx):
             permissible_inputs = self._permissible_inputs_for_output_region()
             dna[gene_idx] = rng.choice(permissible_inputs)
-            return True
-        else:
             return False
+        else:
+            return True
 
     def _mutate_hidden_region(
         self, dna: List[int], gene_idx: int, active_regions: List[int], rng: np.random.RandomState
--- a/test/test_genome.py
+++ b/test/test_genome.py
@@ -404,6 +404,31 @@ def test_mutate_output_region(rng_seed):
         ID_NON_CODING_GENE,
     ]
 
-    assert genome._mutate_output_region(list(dna), 9, rng) is False
-    assert genome._mutate_output_region(list(dna), 10, rng) is True
-    assert genome._mutate_output_region(list(dna), 11, rng) is False
+    assert genome._mutate_output_region(list(dna), 9, rng) is True
+    assert genome._mutate_output_region(list(dna), 10, rng) is False
+    assert genome._mutate_output_region(list(dna), 11, rng) is True
+
+
+@pytest.mark.parametrize("mutation_rate", [0.02, 0.05, 0.2])
+def test_correct_number_of_mutations(mutation_rate, rng_seed):
+
+    n_inputs = 2
+    n_outputs = 1
+    n_columns = 10
+    n_rows = 2
+    levels_back = 5
+    primitives = (cgp.Add, cgp.Sub, cgp.Mul, cgp.Div, cgp.ConstantFloat)
+
+    rng = np.random.RandomState(rng_seed)
+    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives)
+    genome.randomize(rng)
+
+    n_mutations = 0
+    genome_new = genome.clone()
+    genome_new.mutate(mutation_rate, rng)
+    for (gene_0, gene_1) in zip(genome.dna, genome_new.dna):
+        if gene_0 != gene_1:
+            n_mutations += 1
+
+    n_mutations_expected = int(mutation_rate * len(genome.dna))
+    assert n_mutations == n_mutations_expected
--- a/examples/example_evo_regression.py
+++ b/examples/example_evo_regression.py
@@ -141,7 +141,8 @@ if __name__ == "__main__":
 
         f_graph = champion.to_func()
         x_0_range = np.linspace(-5.0, 5.0, 20)
-        x_1_range = np.ones_like(x_0_range) * 2.0  # fix x_1 such than 1d plot makes sense
+        x_1_range = np.ones_like(x_0_range) * 2.0
+        # fix x_1 such than 1d plot makes sense
         y = [f_graph([x_0, x_1_range[0]]) for x_0 in x_0_range]
         y_target = target_function(np.hstack([x_0_range.reshape(-1, 1), x_1_range.reshape(-1, 1)]))
 
--- a/test/test_population.py
+++ b/test/test_population.py
@@ -57,7 +57,7 @@ def test_crossover_two_offspring(population_simple_fitness):
 
 
 def test_mutate(population_params, genome_params):
-    population_params["mutation_rate"] = 0.999999
+    population_params["mutation_rate"] = 0.5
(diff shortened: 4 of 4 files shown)


test/test_genome.py
```diff
@@ -404,6 +404,29 @@ def test_mutate_output_region(rng_seed):
...
+    primitives = (cgp.Add, cgp.Sub, cgp.Mul, cgp.Div, cgp.ConstantFloat)
+
+    rng = np.random.RandomState(rng_seed)
+    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives)
+    genome.randomize(rng)
+
+    dna_diff = 0
+    genome_new = genome.clone()
+    genome_new.mutate(mutation_rate, rng)
+    for idx_dna in range(len(genome.dna)):
```
jakobj, line 428: here you could also `zip` the dnas of the two genomes instead of using the C-style iteration

test/test_genome.py
```diff
@@ -404,6 +404,29 @@ def test_mutate_output_region(rng_seed):
...
+        if genome.dna[idx_dna] != genome_new.dna[idx_dna]:
+            dna_diff += 1
+
+    assert dna_diff == int(mutation_rate * len(genome.dna))
```
jakobj, line 432:
```suggestion
    n_mutations_expected = int(mutation_rate * len(genome.dna))
    assert dna_diff == n_mutations_expected
```
just to make this bit a bit easier to read

test/test_genome.py
```diff
@@ -404,6 +404,29 @@ def test_mutate_output_region(rng_seed):
...
+    n_columns = 10
+    n_rows = 2
+    levels_back = 5
+    primitives = (cgp.Add, cgp.Sub, cgp.Mul, cgp.Div, cgp.ConstantFloat)
+
+    rng = np.random.RandomState(rng_seed)
+    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives)
+    genome.randomize(rng)
+
+    dna_diff = 0
```
jakobj, line 425: i would prefer to call this variable `n_mutations` because it counts the number of mutations

jakobj:
thanks for the update :sparkles: unfortunately travis is unhappy with the new changes, could you please have a look and fix the issues?

cgp/genome.py
```diff
@@ -361,6 +361,14 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
         bool
             True if only inactive regions of the genome were mutated, False otherwise.
         """
+
+        def count_dna_differences(dna_0: List[int], dna_1: List[int]) -> int:
+            dna_diff = 0
+            for gene_0, gene_1 in zip(dna_0, dna_1):
+                if gene_0 != gene_1:
+                    dna_diff += 1
+            return dna_diff
```
mschmidt87, line 370: This can be written as a one-liner:

mschmidt87:
```suggestion
        def count_dna_differences(dna0: List[int], dna1: List[int]) -> int:
            return len([1 for gene0, gene1 in zip(dna0, dna1) if gene0 != gene1])
```

mschmidt87: This is just a personal preference (I think), but I'd prefer appending numbers without a `_` to variable names, i.e. to write `dna0, dna1, gene0, gene1` in this case.

test/test_genome.py
```diff
@@ -404,6 +404,30 @@ def test_mutate_output_region(rng_seed):
...
+    rng = np.random.RandomState(rng_seed)
+    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives)
+    genome.randomize(rng)
+
+    n_mutations = 0
+    genome_new = genome.clone()
+    genome_new.mutate(mutation_rate, rng)
+    for (gene_0, gene_1) in zip(genome.dna, genome_new.dna):
+        if gene_0 != gene_1:
+            n_mutations += 1
```
mschmidt87, line 430:
```suggestion
    n_mutations = len([1 for gene0, gene1 in zip(genome.dna, genome_new.dna) if gene0 != gene1])
```

mschmidt87: You could also import `cgp.genome.count_dna_differences` here.

test/test_genome.py
```diff
@@ -404,6 +404,30 @@ def test_mutate_output_region(rng_seed):
...
 
-    assert genome._mutate_output_region(list(dna), 9, rng) is False
-    assert genome._mutate_output_region(list(dna), 10, rng) is True
-    assert genome._mutate_output_region(list(dna), 11, rng) is False
+    assert genome._mutate_output_region(list(dna), 9, rng) is True
+    assert genome._mutate_output_region(list(dna), 10, rng) is False
+    assert genome._mutate_output_region(list(dna), 11, rng) is True
+
+
+def test_correct_number_of_mutations(mutation_rate, rng_seed):
```
mschmidt87, line 412: I think it would be better to test this with e.g. three mutation rates, instead of just one, using `parametrize` fixture of `pytest`.

cgp/genome.py
jakobj: regarding the one liner: please show both to your colleague for 5sec and ask them if the understand what's happening ;) i do like compact code but it feels like this change would be sacrificing readability

test/test_genome.py
jakobj: i would not reuse the function from the library, the probability of implementing it wrong two times is trivially smaller than implementing it wrong once ;)

jakobj:
regarding the example: the mutation rate was chosen accordingly to the CGP literature, assuming the implementation is correct, so it should be fine. seems like they all still work, don't they?

test/test_genome.py
HenrikMettler: In the newest commit I added: `@pytest.mark.parametrize("mutation_rate", [0.02, 0.05, 0.2,])` just before this test. Is that the right way to do it?

cgp/genome.py
mschmidt87: Hmm, I understand your point in general, but in this case I think the list comprehension is pretty readable and more readable than the original code.

test/test_genome.py
mschmidt87: > regarding the example: the mutation rate was chosen accordingly to the CGP literature, assuming the implementation is correct, so it should be fine. seems like they all still work, don't they?

I don't know, I haven't run them. The CI only checks whether they crash, but not if they convert to good fitness values.

mschmidt87:
> regarding the example: the mutation rate was chosen accordingly to the CGP literature, assuming the implementation is correct, so it should be fine. seems like they all still work, don't they?

I haven't tested the examples. The CI only checks if they crash, not if they convert.

test/test_genome.py
mschmidt87: Looks good to me, although I am not sure whether this might interfer with the fixture from `conftest.py` that has the same name.

test/test_genome.py
mschmidt87: Okay for me.

cgp/genome.py
jakobj: ok

mschmidt87:
I think that CI is failing because of `flake8` issues. I would recommend to install pre-commit hooks for git to catch such errors locally, @HenrikMettler .

test/test_genome.py
```diff
@@ -404,6 +404,29 @@ def test_mutate_output_region(rng_seed):
...
+    primitives = (cgp.Add, cgp.Sub, cgp.Mul, cgp.Div, cgp.ConstantFloat)
+
+    rng = np.random.RandomState(rng_seed)
+    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives)
+    genome.randomize(rng)
+
+    n_mutations = 0
+    genome_new = genome.clone()
+    genome_new.mutate(mutation_rate, rng)
+    n_mutations = len([1 for gene0, gene1 in zip(genome.dna, genome_new.dna) if gene0 != gene1])
```
jakobj, line 429: even under the assumption of the one liner being acceptable in the main module, i would strongly suggest to use the long, slow, but super easy to write and read version in the tests. test should always be very verbose, cool tricks can be applied in the main module.

jakobj:
> > regarding the example: the mutation rate was chosen accordingly to the CGP literature, assuming the implementation is correct, so it should be fine. seems like they all still work, don't they?
> 
> I haven't tested the examples. The CI only checks if they crash, not if they convert.

good point! @HenrikMettler please also manually test the examples and check whether they converge properly.

test/test_genome.py
HenrikMettler: my bad I didn't notice your reply to the above comments, sorry

jakobj: no worries, you can just "unfix" it again in the test ;)

HenrikMettler:
For the examples: 
- all of them stop prior to the max number of generations with values ~ -1e-15 (to -1e-30) -> that this is convergence I assume? 
- `example_evo_regression.py`: The first only converges after 950 of (max) 1000 generations. Should I increase the max number of allowed generations?

jakobj:
no, i think it's fine, no need to change it as long as it converges within the given number of generations. thanks for checking this :)

HenrikMettler:
the latest commit should clear all the issues. I can't squash though (I think) since i don't have the rights to squash and merge (if I understand this: [https://cloudfour.com/thinks/squashing-your-pull-requests/](url) correctly)

examples/example_differential_evo_regression.py
```diff
@@ -126,7 +126,7 @@ def recording_callback(pop):
 
     history, champion = evolution()
 
-    print(f"Final expression {champion.to_sympy()[0]} with fitness {champion.fitness}")
+    print(f"Final expression {champion.to_sympy()[0]} " f"with fitness {champion.fitness}")
```
jakobj, line 129: why this change?

HenrikMettler: hmm... i had to do a linebreak to make `flake` happy... and then I guess `black` removed it again?

jakobj: that's strange. could you try removing the `" f"` again and see what travis thinks? maybe after all you need to run flake with `flake8 --config=.flake8 .` in contrast to what i might've said :P

mschmidt87:
Before we merge, one more (minor) thing: could you make the commit message a bit nicer, i.e., remove the "some more changes" lines etc., and stick to imperative tense and capitalized beginning?

PR #159, opened 2020-06-29
Add new nodes with multiplicative and additive parameters

### LINKED ISSUE #98: Introduce multiplicative and additive parameters for each node
Currently the only way to introduce constants in the expressions is via the `ConstantFloat` or `Parameter` nodes. The values of the parameter nodes can be tuned by local search to specific values in order to increase the fitness of individuals.
However, there is a fundamental problem. As nodes can be used in more than one subsequent node as inputs, it can happen that a particular `Parameter` node feeds into two subsequent nodes. If one of them would require the parameter to be large, the other to be small, the gradient would point in opposite directions, thereby bringing the local search for this particular parameter to a halt, or at least slowing it down significantly. These can lead to difficult-to-escape local minima in the search landscape.

This can be avoided by additionally introducing parameters private to each node (e.g., Izzo et al., 2017). These private parameters are by definition not used by any other node and hence are never subject to "opposing" gradient information. For simplicity a first implementation would introduce two parameters, one that scales the output of each node and one that shifts it. By default these would have the values 1.0 and 0.0, respectively and hence not change the behavior of the node from the current implementation. I guess it could be helpful to introduce a mechanic that allows users to choose between updating only values of `Parameter` nodes, only private node parameters or both. As a side remark, this would allow the evolution of simple ANNs by introducing nodes that only apply an non-linearity to their input (e.g., CGPANNs; Khan et al., 2013).

> Izzo, D., Biscani, F., & Mereta, A. (2017). Differentiable Genetic Programming. In European Conference on Genetic Programming (pp. 35-51). Springer, Cham.
> Khan, M. M., Ahmad, A. M., Khan, G. M., & Miller, J. F. (2013). Fast Learning Neural Networks using Cartesian Genetic Programming. Neurocomputing, 121, 274-289.

jakobj:
This PR introduces new nodes for addition, subtraction and multiplication operations that are parametrized with multiplicative and additive parameters, such that the operations looks, for example, like this `w * (input[0] + input[1]) + b`. These parameters (`w,b`) can be tuned via local search.

This is a first draft. @mschmidt87  I'd like your input with regards to extensibility/maintainability of this implementation

closes #98

### DIFF
--- a/cgp/node.py
+++ b/cgp/node.py
@@ -1,4 +1,4 @@
-from typing import List, Type, TYPE_CHECKING
+from typing import List, Tuple, Type, TYPE_CHECKING
 
 if TYPE_CHECKING:
     from .cartesian_graph import CartesianGraph
@@ -30,9 +30,9 @@ class Node:
     _arity: int
     _active: bool = False
     _inputs: List[int]
+    is_parametrized: bool = False
     _output: float
     _output_str: str
-    _parameter_str: str
     _idx: int
 
     def __init__(self, idx: int, inputs: List[int]) -> None:
@@ -149,16 +149,27 @@ class Node:
         """
         self.format_output_str(graph)
 
-    def format_parameter_str(self) -> None:
-        raise NotImplementedError
-
     @property
     def output_str(self) -> str:
         return self._output_str
 
+
+class ParametrizedNode(Node):
+
+    is_parametrized: bool = True
+    _parameter_str: Tuple[str, ...]
+    prefixes: Tuple[str, ...]
+
     @property
-    def parameter_str(self) -> str:
+    def parameter_str(self) -> Tuple[str, ...]:
         return self._parameter_str
+
+    def format_parameter_str(self) -> None:
+        raise NotImplementedError
+
+    @staticmethod
+    def initial_value(prefix: str) -> float:
+        raise NotImplementedError()
 
 
 class Add(Node):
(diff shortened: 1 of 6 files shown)


cgp/node.py
```diff
@@ -179,6 +190,40 @@ def format_output_str(self, graph: "CartesianGraph") -> None:
...
+        )
+
+    @staticmethod
+    def initial_value(prefix: str) -> float:
+        if prefix == "w":
+            return 1.0
+        elif prefix == "b":
+            return 0.0
+        else:
+            raise KeyError()
```
mschmidt87, line 224: Could be implemented in a shorter way with a `dict`?

mschmidt87:
```suggestion
    def initial_value(prefix: str) -> float:
        initial_values = {'w': 1.0, 'b': 0.0}
        return initial_values[prefix]

File: cgp/node.py [hunk 2]
```diff
@@ -197,6 +242,40 @@ def format_output_str(self, graph: "CartesianGraph") -> None:
...
```
mschmidt87, line 276:
```suggestion
    def initial_value(prefix: str) -> float:
        initial_values = {'w': 1.0, 'b': 0.0}
        return initial_values[prefix]
```

cgp/node.py
```diff
@@ -215,6 +294,40 @@ def format_output_str(self, graph: "CartesianGraph") -> None:
...
+        * {graph[self._inputs[1]].output_str}) + self._b{self._idx})"
+
+    def format_parameter_str(self) -> None:
+        self._parameter_str = (
+            f"self._w{self._idx} = torch.nn.Parameter(torch.DoubleTensor([<w{self._idx}>]))\n",
+            f"self._b{self._idx} = torch.nn.Parameter(torch.DoubleTensor([<b{self._idx}>]))\n",
+        )
+
+    @staticmethod
+    def initial_value(prefix: str) -> float:
```
mschmidt87, line 322: see above

jakobj:
thanks for the input @mschmidt87 :+1: since #178 introduces major changes to how nodes are defined i close this PR for now and will create an up-to-date PR once #178 is merged.

PR #160, opened 2020-06-29
Provide possibility to restrict local search to k best individuals

### LINKED ISSUE #116: Introduce a parameter to restrict local search to the best k individuals
Currently, local search is applied to all individuals from the population. Since local search is expensive, it would be computationally beneficial to only apply it to promising, i.e., high fitness individuals (see, e.g., Z-Flores et al., 2015).
Introducing this parameter will require some thinking on the order of fitness evaluation and local search in each generation.

> Z-Flores, E., Trujillo, L., Schütze, O., & Legrand, P. (2015). A local search approach to genetic programming for binary classification. In Proceedings of the 2015 Annual Conference on Genetic and Evolutionary Computation (pp. 1151-1158).

jakobj:
as discussed it can be beneficial to restrict the local search to the `k` best performing individuals (#116). this PR implements this suggestion. as expected this decreases wallclock time when used, e.g., in the `example_differential_evo_regression.py` from ~57s to ~44s when running local search only for the best two individuals instead of the whole population (5 individuals).

closes #116

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -1,7 +1,7 @@
 import concurrent.futures
 import numpy as np
 
-from typing import Callable, List, Tuple
+from typing import Callable, List, Tuple, Union
 
 from ..individual import IndividualBase
 from ..population import Population
@@ -25,6 +25,7 @@ class MuPlusLambda:
         *,
         n_processes: int = 1,
         local_search: Callable[[IndividualBase], None] = lambda combined: None,
+        k_local_search: Union[int, None] = None,
     ):
         """Init function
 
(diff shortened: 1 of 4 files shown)


cgp/ea/mu_plus_lambda.py
```diff
@@ -108,10 +110,25 @@ def step(
...
+        # during local search; finally we sort again by fitness, now
+        # taking into account the effect of local search for
+        # subsequent selection
         combined = self._compute_fitness(combined, objective)
+        combined = self._sort(combined)
+
+        if self.k_local_search is None:
+            k_local_search = len(combined)
+        else:
+            k_local_search = self.k_local_search
```
mschmidt87, line 127: what about putting this in one line? :-)

mschmidt87, line 127:
```suggestion
        k_local_search = len(combined) if self.k_local_search is None else self.k_local_search
```

jakobj: ok, in this case it's nice :P

jakobj:
thanks for the input! :ok_hand: addressed your comments, pls have another look

PR #161, opened 2020-06-29
Unify node tests

jakobj:
while some nodes were tested only for returning the correct values when calling the graph, others were only tested for some `to_...` functions. this PR unifies the tests, such that all nodes are tested for all supported uses.

### DIFF
--- a/test/test_node.py
+++ b/test/test_node.py
@@ -24,6 +24,58 @@ def test_inputs_are_cut_to_match_arity():
     assert node.inputs == inputs[:2]
 
 
+def _test_graph_call_and_to_x_compilations(
+    genome,
+    x,
+    y_target,
+    *,
+    test_graph_call=True,
+    test_to_func=True,
+    test_to_numpy=True,
+    test_to_torch=True,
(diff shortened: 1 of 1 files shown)


test/test_node.py
```diff
@@ -24,6 +24,58 @@ def test_inputs_are_cut_to_match_arity():
...
+def _test_graph_call_and_to_x_compilations(
+    genome,
+    x,
+    y_target,
+    *,
+    skip_graph_call=False,
+    skip_to_func=False,
+    skip_to_numpy=False,
+    skip_to_torch=False,
+    skip_to_sympy=False,
```
mschmidt87, line 36: Kind of minor, but I would prefer to turn these switches around and rather to something like e.g. `test_to_func=True` here and then below `if test_to_func:`.
It is easier to parse if the default arg is `True` .

jakobj: yes, good suggestion!

jakobj:
included your suggestion, thanks! :ok_hand: pls have another look

PR #168, opened 2020-07-01
Track parent idx in offsprings, i.e., all cloned individuals

### LINKED ISSUE #129: Track parent in each `Individual`
If each Individual remembers the ID of their parent, we can reconstruct lineages, i.e., investigate via which steps a final expression has evolved. Instead of showing how the expression in the fittest individual per generation evolves, this would allow us to analyze the evolution of each specific expression. Sounds like this could be useful.

jakobj:
This PR implements tracking of the parent's idx in each individual to allow the reconstruction of full lineages.

Contains changes from #167, so should be merged after.

closes #129

### DIFF
--- a/cgp/individual.py
+++ b/cgp/individual.py
@@ -32,6 +32,7 @@ class IndividualBase:
         """
         self.fitness: Union[float, None] = fitness
         self.idx: int
+        self.parent_idx: int
 
     def clone(self):
         raise NotImplementedError()
@@ -105,7 +106,9 @@ class IndividualSingleGenome(IndividualBase):
         return f"Individual(idx={self.idx}, fitness={self.fitness}, genome={self.genome}))"
 
     def clone(self) -> "IndividualSingleGenome":
-        return IndividualSingleGenome(self.fitness, self.genome.clone())
+        ind = IndividualSingleGenome(self.fitness, self.genome.clone())
+        ind.parent_idx = self.idx
+        return ind
 
     def mutate(self, mutation_rate: float, rng: np.random.RandomState) -> None:
         only_silent_mutations = self._mutate_genome(self.genome, mutation_rate, rng)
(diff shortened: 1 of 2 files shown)


cgp/individual.py
```diff
@@ -105,7 +106,9 @@ def __repr__(self):
         return f"Individual(idx={self.idx}, fitness={self.fitness}, genome={self.genome}))"
 
     def clone(self) -> "IndividualSingleGenome":
-        return IndividualSingleGenome(self.fitness, self.genome.clone())
+        ind = IndividualSingleGenome(self.fitness, self.genome.clone())
+        ind.parent_idx = self.idx
```
mschmidt87, line 110: This is a bit tricky, but I don't think that we can generally say that the clone of an individual has the individual as its parent, can we?

cgp/individual.py
```diff
@@ -149,7 +152,9 @@ def __init__(self, fitness: Union[float, None], genome: List[Genome]) -> None:
         self.genome: List[Genome] = genome
 
     def clone(self) -> "IndividualMultiGenome":
-        return IndividualMultiGenome(self.fitness, [g.clone() for g in self.genome])
+        ind = IndividualMultiGenome(self.fitness, [g.clone() for g in self.genome])
+        ind.parent_idx = self.idx
```
mschmidt87, line 156: same here.

cgp/individual.py
jakobj: uhm, i'm not sure. what other definition of "parent" is there other than the one from which we create new individuals? maybe i'm misunderstanding your questions :thinking:

jakobj:
> I think I am missing something here, but I don't quite understand what your intended mechanism is. In this PR, you implement the `parent_idx` in the `clone()` function which may or may not be a good idea (see my comment in the code, can we really assume that `clone` is only used to create new children?).

(see inline comment)

> On the other hand, in #170 , you change the implementation of `MuPlusLambda._create_new_parent_population` such that it does not use `clone()` anymore. So, the two PRs seem to contradict each other.

no i would think they don't contradict each other: this PR makes each clone, i.e., each offspring, track from which individual it was created. #170 just simplifies the creation of the parent population, where no individuals are created just not-good-enough offsprings are killed off by keeping just the first `n_parents` individuals of the combined population. maybe it was confusing that `clone()` was previously used in that function, which was logically false as implicit in my previous sentence.

cgp/individual.py
mschmidt87: Yes, okay, I think this makes sense.

mschmidt87:
Thanks, I was a bit confused because I thought that via #170 , there would be no cloning of individuals anymore, but then I realized that `clone()` is already used in the `_create_new_offspring_population`.

PR #169, opened 2020-07-01
Simplify sorting of individuals in `mu_plus_lambda`

jakobj:
My brain was getting too hot whenever i looked at this function. Isn't this implementation much easier? It seems that we can handle the replacement `nan` -> `-np.inf` on the fly in the function used for `key`, no need for creating a copy of the population which we modify and then argsort.

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -1,7 +1,7 @@
 import concurrent.futures
 import numpy as np
 
-from typing import Callable, List, Tuple, Union
+from typing import Callable, List, Union
 
 from ..individual import IndividualBase
 from ..population import Population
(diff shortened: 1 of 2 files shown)


cgp/ea/mu_plus_lambda.py
```diff
@@ -151,33 +151,22 @@ def _compute_fitness(
         return combined
 
     def _sort(self, combined: List[IndividualBase]) -> List[IndividualBase]:
-        # create copy of population
-        combined_copy = [ind.clone() for ind in combined]
+        def sort_func(ind: IndividualBase) -> float:
+            """Return fitness of an individual, return -infinity for individuals
+            with fitness equal nan, or raise error if it is None.
```
mschmidt87, line 156:
```suggestion
            with fitness equal nan, or raise error if it is not a float.
```

PR #170, opened 2020-07-01
Simplify `create_new_parent_population` in `mu_plus_lambda`

jakobj:
Seems like this function can be significantly simplified. I don't think there's a danger of sharing references, as each offspring is created via clone, which avoid specifically that.

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -182,19 +182,8 @@ class MuPlusLambda:
     def _create_new_parent_population(
         self, n_parents: int, combined: List[IndividualBase]
     ) -> List[IndividualBase]:
-        # create new parent population by picking the `n_parents` individuals
-        # with the highest fitness
-        parents: List[IndividualBase] = []
-        for i in range(n_parents):
-            # note: unclear whether clone() is needed here; using
-            # clone() avoids accidentally sharing references across
-            # individuals, but might incur a performance penalty
-            new_individual = combined[i].clone()
+        """Create the new parent population by picking the first `n_parents`
+        individuals from the combined population.
 
-            # since this individual is genetically identical to its
-            # parent, the identifier is the same
-            new_individual.idx = combined[i].idx
-
-            parents.append(new_individual)
-
-        return parents
+        """
+        return combined[:n_parents]


cgp/ea/mu_plus_lambda.py
```diff
@@ -182,19 +182,7 @@ def sort_func(zipped_ind: Tuple[int, IndividualBase]) -> float:
...
-            # parent, the identifier is the same
-            new_individual.idx = combined[i].idx
-
-            parents.append(new_individual)
-
-        return parents
+        """Create the new parent population by picking the `n_parents` individuals
+        with the highest fitness
+        """
+        return combined[:n_parents]
```
mschmidt87, line 188: this implementations seems to assume that `combined` is sorted by fitness. While this might be true due at the place it is called in the implementation (I haven't checked but I guess you did), it is dangerous to just silently assume this. I would therefore suggest to add a line that sorts `combined` here.

mschmidt87:
Actually, it seems that without this PR, the tracking of lineage of individuals across the evolution made possible via #167 and #168 , would not be possible, because in the current implementation, we are cloning individuals to create a new parent population which would destroy the provenance tracking. Right?

jakobj:
> Actually, it seems that without this PR, the tracking of lineage of individuals across the evolution made possible via #167 and #168 , would not be possible, because in the current implementation, we are cloning individuals to create a new parent population which would destroy the provenance tracking. Right?

yes, this is true. of course only if we don't copy the idx in the `clone` operation. ;)

cgp/ea/mu_plus_lambda.py
jakobj: that is a good point! but this function should _not_ sort the individuals! this should be done before (orchestrated by the `step` function), and if that's not true, this function should fail, loudly! i'll implement an `assert(sorted)`-style check just to make sure.

jakobj: actually (besides having trouble with mypy) i don't think we should implement this check. it's not necessary the case that we only want to create a new parent population from a sorted combined population. it's completely up to the algorithm (or derivatives/subclasses of it) in which order it arranges the individuals in `combined`. all this function does, is pick the first `n_parents` as new parents for the new generation, whatever their order.

PR #171, opened 2020-07-03
Extend testing of MuPlusLambda class

### LINKED ISSUE #166: More thorough testing of `mu_plus_lambda`
The tests for the `mu_plus_lambda` implementation are a bit sparse at the moment (`test_objective_with_label`, `test_fitness_contains_nan`). To assure continued correct functionality we should strive to test each function individually.

mschmidt87:
closes #166

### DIFF
--- a/test/test_ea_mu_plus_lambda.py
+++ b/test/test_ea_mu_plus_lambda.py
@@ -28,7 +28,7 @@ def test_objective_with_label(population_params, genome_params):
     assert pop.champion.fitness == pytest.approx(-1.0)
 
 
-def test_fitness_contains_and_maintains_nan(population_params, genome_params):
+def test_fitness_contains_and_maintains_nan(population_params, genome_params, ea_params):
     def objective(individual):
         if np.random.rand() < 0.95:
             individual.fitness = np.nan
@@ -38,21 +38,23 @@ def test_fitness_contains_and_maintains_nan(population_params, genome_params):
 
     pop = cgp.Population(**population_params, genome_params=genome_params)
 
-    ea = cgp.ea.MuPlusLambda(10, 10, 1)
+    ea = cgp.ea.MuPlusLambda(**ea_params)
     ea.initialize_fitness_parents(pop, objective)
     ea.step(pop, objective)
 
     assert np.nan in [ind.fitness for ind in pop]
 
 
-def test_offspring_individuals_are_assigned_correct_indices(population_params, genome_params):
+def test_offspring_individuals_are_assigned_correct_indices(
+    population_params, genome_params, ea_params
+):
     def objective(ind):
         ind.fitness = 0.0
         return ind
 
     pop = cgp.Population(**population_params, genome_params=genome_params)
 
-    ea = cgp.ea.MuPlusLambda(10, 10, 1)
+    ea = cgp.ea.MuPlusLambda(**ea_params)
     ea.initialize_fitness_parents(pop, objective)
 
     offsprings = ea._create_new_offspring_generation(pop)
@@ -62,16 +64,17 @@ def test_offspring_individuals_are_assigned_correct_indices(population_params, g
 
 
 def test_offspring_individuals_are_assigned_correct_parent_indices(
-    population_params, genome_params
+    population_params, genome_params, ea_params
 ):
     def objective(ind):
         ind.fitness = 0.0
         return ind
 
     population_params["n_parents"] = 1
+
     pop = cgp.Population(**population_params, genome_params=genome_params)
 
-    ea = cgp.ea.MuPlusLambda(10, 10, 1)
+    ea = cgp.ea.MuPlusLambda(**ea_params)
     ea.initialize_fitness_parents(pop, objective)
 
     offsprings = ea._create_new_offspring_generation(pop)
@@ -113,7 +116,7 @@ def test_local_search_is_only_applied_to_best_k_individuals(
     pop = cgp.Population(**population_params, genome_params=genome_params)
 
     local_search = functools.partial(
-        cgp.local_search.gradient_based, objective=inner_objective, **local_search_params,
+        cgp.local_search.gradient_based, objective=inner_objective, **local_search_params
     )
 
     ea = cgp.ea.MuPlusLambda(5, 5, 1, local_search=local_search, k_local_search=k_local_search)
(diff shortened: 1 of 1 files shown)


mschmidt87:
Coverage behaves strangely: I wrote explicit tests for both errors raised in the algorithm, but it still reports the 2nd one to be not covered by the tests.

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
...
     for idx, ind in enumerate(offsprings):
         assert ind.idx == len(pop.parents) + idx
+
+
+def test_n_offspring_less_than_n_breeding():
+    n_offsprings = 10
+    n_breedings = 5
+    with pytest.raises(ValueError):
+        _ = cgp.ea.MuPlusLambda(n_offsprings, n_breedings, 1)
+
```
jakobj, line 67:
```suggestion
    n_offspring = 10
    n_breeding = 5
    with pytest.raises(ValueError):
        cgp.ea.MuPlusLambda(n_offsprings, n_breeding, 1)

```

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
...
+
+def test_fitness_wrong_return_type(population_params, genome_params):
+    def objective(individual):
+        individual.fitness = 5
```
jakobj, line 71:
```suggestion
        individual.fitness = int(5.0)  # should raise error since fitness should be float
```
just to be super explicit

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
...
+        return individual
+
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+
+    ea = cgp.ea.MuPlusLambda(10, 10, 1)
```
jakobj, line 76: would it make sense to use the `ea_params` fixture here and in other places?

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
...
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    n_offsprings = 10
+    ea = cgp.ea.MuPlusLambda(n_offsprings, 10, 1)
+    ea.initialize_fitness_parents(pop, objective)
+    sorted_parents = ea._sort(pop.parents)
+    # Assert that the sorting inverted the list of parents (because the fitness is equal to the id)
+    assert sorted_parents == pop.parents[::-1]
+
+
+def test_create_new_populations(population_params, genome_params):
```
jakobj, line 128:
```suggestion
def test_create_new_offspring_and_parent_generation(population_params, genome_params):
```

jakobj: actually, shouldn't this be two tests since both might break independently?

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
...
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    n_offsprings = 10
+    ea = cgp.ea.MuPlusLambda(n_offsprings, 10, 1)
+
+    ea.initialize_fitness_parents(pop, objective)
+
+    offsprings = ea._create_new_offspring_generation(pop)
+    assert len(offsprings) == n_offsprings
+    assert all([ind.idx >= pop.n_parents for ind in offsprings])
+    # Assert that all offspring dna are different from all parents dna
```
jakobj, line 142: does this check for "not identity" or "not equality"? in other words, does this make sure that references are not shared between parents and offsprings or that mutations have occurred?

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
...
+    offspring_dna = [ind.genome.dna for ind in offsprings]
+    parent_dna = [ind.genome.dna for ind in pop.parents]
+    assert all([odna != pdna for odna in offspring_dna for pdna in parent_dna])
+
+    # Create new "parent" population from the offsprings and assert
+    # that we picked the 5 best ones (here, by construction, the 5
+    # highest ids)
+    offsprings = ea._compute_fitness(offsprings, objective)
```
jakobj, line 150: shouldn't you introduce `combined = offsprings + parents` here to use in the following? this would be more in line with the implementation, maybe less surprising to the reader?

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
...
+
+
+def test_n_offspring_less_than_n_breeding():
+    n_offsprings = 10
+    n_breedings = 5
+    with pytest.raises(ValueError):
+        _ = cgp.ea.MuPlusLambda(n_offsprings, n_breedings, 1)
+
+
+def test_fitness_wrong_return_type(population_params, genome_params):
```
jakobj, line 69:
```suggestion
def test_fitness_has_wrong_type(population_params, genome_params):
```

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
 
     for idx, ind in enumerate(offsprings):
         assert ind.idx == len(pop.parents) + idx
+
+
+def test_n_offspring_less_than_n_breeding():
```
jakobj, line 62:
```suggestion
def test_n_offsprings_less_than_n_breeding():
```

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
...
+
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+
+    ea = cgp.ea.MuPlusLambda(10, 10, 1)
+    ea.initialize_fitness_parents(pop, objective)
+    with pytest.raises(ValueError):
+        ea.step(pop, objective)
+
+
+def test_fitness_initialization(population_params, genome_params):
```
jakobj, line 82:
```suggestion
def test_initialize_fitness_parents(population_params, genome_params):
```

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
...
+        return individual
+
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+
+    ea = cgp.ea.MuPlusLambda(10, 10, 1)
+    ea.initialize_fitness_parents(pop, objective)
+    assert all([ind.fitness is not None for ind in pop.parents])
+
+
+def test_one_step(population_params, genome_params):
```
jakobj, line 94:
```suggestion
def test_step(population_params, genome_params):
```

test/test_ea_mu_plus_lambda.py
```diff
@@ -57,3 +57,97 @@ def objective(ind):
...
+    new_parent_ids = sorted([ind.idx for ind in pop.parents])
+    # After one step, the new parent population should have IDs that
+    # are offset from the old parent ids by n_offsprings
+    # This is by construction in this test because the fitness is equal to the id
+    assert all(
+        [new_id == old_id + n_offsprings for new_id, old_id in zip(new_parent_ids, old_parent_ids)]
+    )
+
+
+def test_fitness_sort(population_params, genome_params):
```
jakobj, line 114:
```suggestion
def test_sort(population_params, genome_params):
```

test/test_ea_mu_plus_lambda.py
mschmidt87: Yes! Done.

test/test_ea_mu_plus_lambda.py
mschmidt87: Yes, I had that and then realized that I had just rewritten the step function. So, I wanted to keep this more simple here. But I can change that.

test/test_ea_mu_plus_lambda.py
mschmidt87: This checks that all offspring have different dna, i.e. that they have been mutated.

test/test_ea_mu_plus_lambda.py
mschmidt87: I now split the test in two, and simplified the parent_population test significantly, because the functionality was simplified in #170

mschmidt87:
Thanks for your feedback, I implemented the suggestions and addressed all comments. Please check again @jakobj

test/test_ea_mu_plus_lambda.py
```diff
@@ -125,3 +128,100 @@ def objective(ind):
 
     for idx in range(k_local_search, population_params["n_parents"]):
         assert pop[idx].genome._parameter_names_to_values["<p1>"] == pytest.approx(1.0)
+
+
+def test_n_offsprings_less_than_n_breeding():
+    n_offspring = 10
+    n_breedings = 5
+    with pytest.raises(ValueError):
+        cgp.ea.MuPlusLambda(n_offspring, n_breedings, 1)
```
jakobj, line 137:
```suggestion
def test_raise_n_offsprings_less_than_n_breeding():
    n_offsprings = 10
    n_breeding = 5
    with pytest.raises(ValueError):
        cgp.ea.MuPlusLambda(n_offsprings, n_breeding, 1)
```

test/test_ea_mu_plus_lambda.py
```diff
@@ -125,3 +128,100 @@ def objective(ind):
...
+
+
+def test_fitness_has_wrong_type(population_params, genome_params, ea_params):
```
jakobj, line 140:
```suggestion
def test_raise_fitness_has_wrong_type(population_params, genome_params, ea_params):
```

test/test_ea_mu_plus_lambda.py
```diff
@@ -125,3 +128,100 @@ def objective(ind):
...
+
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    ea = cgp.ea.MuPlusLambda(**ea_params)
+
+    ea.initialize_fitness_parents(pop, objective)
+
+    offsprings = ea._create_new_offspring_generation(pop)
+    assert len(offsprings) == ea_params["n_offsprings"]
+    assert all([ind.idx >= pop.n_parents for ind in offsprings])
+    # Assert that all offspring dna are different from all parents dna
```
jakobj, line 214: whether this is true will depend on our decision in #172. no need to change this test now, just making sure we're on the same page

test/test_ea_mu_plus_lambda.py
```diff
@@ -125,3 +128,100 @@ def objective(ind):
...
+    offspring_dna = [ind.genome.dna for ind in offsprings]
+    parent_dna = [ind.genome.dna for ind in pop.parents]
+    assert all([odna != pdna for odna in offspring_dna for pdna in parent_dna])
+
+
+def test_create_new_parent_population(population_params, genome_params, ea_params):
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    ea = cgp.ea.MuPlusLambda(**ea_params)
+
+    # Create new "parent" population from the parents and assert that
```
jakobj, line 224:
```suggestion
    # Create new parent population from the parents and assert that
```

test/test_ea_mu_plus_lambda.py
mschmidt87: Yes, if we're following my suggestion, with a small chance, there might be no change at all. However, we could control for that in this case here by setting the mutation rate to 1 (or approx. 1).

mschmidt87:
> your inline comments all start with a capital letter, while most of the others in the package start with a lower case one. am i missing something here? shouldn't this somehow be regulated by black?

I am not aware of any policy regarding capitalization of inline comments, neither in pep8 (https://www.python.org/dev/peps/pep-0008/#inline-comments) nor in black.

mschmidt87:
Thanks @jakobj , I used your suggestions.

jakobj:
oh one more thing: could you please squash the commits before merging?

Issue #133, opened 2020-06-04
Add test involving IndividualMultiGenome


mschmidt87:
Only a few tests cover the `IndividualMultiGenome` class. However, most, perhaps all, of the tests could be executed with an `IndividualMultiGenome`.

Thus, we should rewrite the tests such that they are carried out for `IndividualSingleGenome` and `IndividualMultiGenome`.

PR #176, opened 2020-07-06
Tests/individual multi genome

mschmidt87:
This PR addresses #133 . It simply executes all tests for both `IndividualSingleGenome` and `IndividualMultiGenome` by wrapping the `Genome` into a list for the multi-genome tests.

To keep the additional code to a minimum, I introduced some helper functions.

Edit:
Additionally, I reorganized the tests to use pytest fixtures which makes the test more modular and extending them more scalable. Essentially, the `to_sympy` etc. tests have three dimensions of parameters:
- `individual_type`
- `params` = the combination of genome, primitives, dna, and target function to test for
- `arg_funcs` = the function arguments that the functions are tested for

Edit 2:
I also added a similar test for `to_numpy` which was lacking before.

### DIFF
--- a/test/test_individual.py
+++ b/test/test_individual.py
@@ -1,139 +1,165 @@
 import math
+import numpy as np
 import pickle
+from collections import namedtuple
+
 import pytest
 
 import cgp
-from cgp.individual import IndividualSingleGenome
+from cgp import IndividualSingleGenome, IndividualMultiGenome
 from cgp.genome import ID_INPUT_NODE, ID_OUTPUT_NODE, ID_NON_CODING_GENE
 
 
-def test_pickle_individual():
+TestParams = namedtuple("TestParams", ["genome_params", "primitives", "dna", "target_function"])
+params_list = [
+    TestParams(
+        genome_params={
+            "n_inputs": 1,
+            "n_outputs": 1,
+            "n_columns": 2,
+            "n_rows": 1,
+            "levels_back": 2,
+        },
+        primitives=(cgp.Add, cgp.Parameter),
+        dna=[
+            ID_INPUT_NODE,
+            ID_NON_CODING_GENE,
+            ID_NON_CODING_GENE,
+            1,
+            0,
+            0,
+            0,
+            0,
+            1,
+            ID_OUTPUT_NODE,
+            2,
+            ID_NON_CODING_GENE,
+        ],
+        target_function=lambda x, c: x + c,
+    )
+]
+GraphInputValues = namedtuple("GraphInputValues", ["x", "c"])
+
+graph_input_values_list = [GraphInputValues(x=[3.0, 5.0], c=1.0), GraphInputValues(x=[3.0], c=2.0)]
+
+
+def _create_genome(genome_params, primitives, dna):
+    genome = cgp.Genome(**genome_params, primitives=primitives)
+    genome.dna = dna
+    return genome
+
+
+def _create_individual(genome, fitness=None, individual_type="SingleGenome"):
+    if individual_type == "SingleGenome":
+        return IndividualSingleGenome(fitness, genome)
+    elif individual_type == "MultiGenome":
+        return IndividualMultiGenome(fitness, [genome])
+    else:
+        raise NotImplementedError("Unknown individual type.")
+
+
+def _unpack_evaluation(value, individual_type="SingleGenome"):
+    if individual_type == "SingleGenome":
+        return value
+    elif individual_type == "MultiGenome":
+        return value[0]
+    else:
+        raise NotImplementedError("Unknown individual type.")
+
+
+def _unpack_genome(individual, individual_type="SingleGenome"):
+    if individual_type == "SingleGenome":
+        return individual.genome
+    elif individual_type == "MultiGenome":
+        return individual.genome[0]
+    else:
+        raise NotImplementedError("Unknown individual type.")
+
+
+@pytest.mark.parametrize("individual_type", ["SingleGenome", "MultiGenome"])
+def test_pickle_individual(individual_type):
 
     primitives = (cgp.Add,)
     genome = cgp.Genome(1, 1, 1, 1, 1, primitives)
-    individual = IndividualSingleGenome(None, genome)
+    individual = _create_individual(genome, individual_type=individual_type)
 
     with open("individual.pkl", "wb") as f:
         pickle.dump(individual, f)
 
 
-def test_individual_with_parameter_python():
+@pytest.mark.parametrize("individual_type", ["SingleGenome", "MultiGenome"])
+@pytest.mark.parametrize("params", params_list)
+@pytest.mark.parametrize("graph_input_values", graph_input_values_list)
+def test_individual_with_parameter_python(individual_type, params, graph_input_values):
+    genome_params, primitives, dna, target_function = params
+    genome = _create_genome(genome_params, primitives, dna)
+    individual = _create_individual(genome, individual_type=individual_type)
 
-    primitives = (cgp.Add, cgp.Parameter)
-    genome = cgp.Genome(1, 1, 2, 1, 2, primitives)
-    # f(x) = x + c
-    genome.dna = [
-        ID_INPUT_NODE,
-        ID_NON_CODING_GENE,
-        ID_NON_CODING_GENE,
-        1,
-        0,
-        0,
-        0,
-        0,
-        1,
-        ID_OUTPUT_NODE,
-        2,
-        ID_NON_CODING_GENE,
-    ]
-    individual = IndividualSingleGenome(None, genome)
+    x, c = graph_input_values.x, graph_input_values.c
+    _unpack_genome(individual, individual_type)._parameter_names_to_values["<p1>"] = c
+    f = _unpack_evaluation(individual.to_func(), individual_type)
 
-    c = 1.0
-    x = [3.0]
-
-    f = individual.to_func()
-    y = f(x)
-
-    assert y[0] == pytest.approx(x[0] + c)
-
-    c = 2.0
-    individual.genome._parameter_names_to_values["<p1>"] = c
-
-    f = individual.to_func()
-    y = f(x)
-
-    assert y[0] == pytest.approx(x[0] + c)
+    for xi in x:
+        y = f([xi])
+        assert y[0] == pytest.approx(target_function(xi, c))
 
 
-def test_individual_with_parameter_torch():
+@pytest.mark.parametrize("individual_type", ["SingleGenome", "MultiGenome"])
+@pytest.mark.parametrize("params", params_list)
+@pytest.mark.parametrize("graph_input_values", graph_input_values_list)
+def test_individual_with_parameter_torch(individual_type, params, graph_input_values):
     torch = pytest.importorskip("torch")
-    primitives = (cgp.Add, cgp.Parameter)
-    genome = cgp.Genome(1, 1, 2, 1, 2, primitives)
-    # f(x) = x + c
-    genome.dna = [
-        ID_INPUT_NODE,
-        ID_NON_CODING_GENE,
-        ID_NON_CODING_GENE,
-        1,
-        0,
-        0,
-        0,
-        0,
-        1,
-        ID_OUTPUT_NODE,
-        2,
-        ID_NON_CODING_GENE,
-    ]
-    individual = IndividualSingleGenome(None, genome)
+    genome_params, primitives, dna, target_function = params
+    genome = _create_genome(genome_params, primitives, dna)
+    individual = _create_individual(genome, individual_type=individual_type)
 
-    c = 1.0
-    x = torch.empty(2, 1).normal_()
-
-    f = individual.to_torch()
+    x, c = torch.tensor(graph_input_values.x).unsqueeze(1), graph_input_values.c
+    _unpack_genome(individual, individual_type)._parameter_names_to_values["<p1>"] = c
+    f = _unpack_evaluation(individual.to_torch(), individual_type)
     y = f(x)
 
-    assert y[0, 0].item() == pytest.approx(x[0, 0].item() + c)
-    assert y[1, 0].item() == pytest.approx(x[1, 0].item() + c)
+    for i in range(x.shape[0]):
+        assert y[i, 0].item() == pytest.approx(target_function(x[i, 0].item(), c))
 
-    c = 2.0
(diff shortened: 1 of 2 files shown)


test/test_individual.py
```diff
@@ -3,21 +3,46 @@
 import pytest
 
 import cgp
-from cgp.individual import IndividualSingleGenome
+from cgp import Genome, IndividualSingleGenome, IndividualMultiGenome
```
jakobj, line 6: why this `from` import? couldn't use those classes with the `cgp.` prefix as well?

test/test_individual.py
```diff
@@ -3,21 +3,46 @@
...
 from cgp.genome import ID_INPUT_NODE, ID_OUTPUT_NODE, ID_NON_CODING_GENE
 
 
-def test_pickle_individual():
+def _create_individual(fitness, genome, ind_type="single"):
```
jakobj, line 10: maybe be extra verbose and always write `individual` instead of `ind`?

test/test_individual.py
```diff
@@ -3,21 +3,46 @@
...
+    if ind_type == "single":
+        return IndividualSingleGenome(fitness, genome)
+    elif ind_type == "multi":
+        if isinstance(genome, Genome):
+            genome = [genome]
+        return IndividualMultiGenome(fitness, genome)
```
jakobj, line 16: `else` branch should raise `NotImplementedError`

test/test_individual.py
```diff
@@ -3,21 +3,46 @@
...
+from cgp import Genome, IndividualSingleGenome, IndividualMultiGenome
 from cgp.genome import ID_INPUT_NODE, ID_OUTPUT_NODE, ID_NON_CODING_GENE
 
 
-def test_pickle_individual():
+def _create_individual(fitness, genome, ind_type="single"):
+    if ind_type == "single":
+        return IndividualSingleGenome(fitness, genome)
+    elif ind_type == "multi":
+        if isinstance(genome, Genome):
```
jakobj, line 14: why do we need this `if`? shouldn't there be only one way to use this function to create `IndividualMultiGenome` either with genome being a `Genome` or `[Genome]`?

test/test_individual.py
```diff
@@ -3,21 +3,46 @@
...
+            genome = [genome]
+        return IndividualMultiGenome(fitness, genome)
+
+
+def _unpack_evaluation(value, ind_type="single"):
+    if ind_type == "single":
+        return value
+    elif ind_type == "multi":
+        return value[0]
```
jakobj, line 23: (see above, missing `else`)

test/test_individual.py
```diff
@@ -3,21 +3,46 @@
...
+
+
+def _unpack_genome(individual, ind_type="single"):
+    if ind_type == "single":
+        return individual.genome
+    elif ind_type == "multi":
+        return individual.genome[0]
```
jakobj, line 30: (see above, missing `else`)

test/test_individual.py
```diff
@@ -158,25 +185,26 @@ def test_to_and_from_torch_plus_backprop():
         3,
         ID_NON_CODING_GENE,
     ]
-    individual = IndividualSingleGenome(None, genome)
+    individual = _create_individual(None, genome, ind_type)
 
     def f_target(x):
         return math.pi * x
 
     f = individual.to_torch()
```
jakobj, line 193: shouldn't this just use the `_unpack_evaluation`?

test/test_individual.py
mschmidt87: yes, it just saves characters in the code.

test/test_individual.py
mschmidt87: okay, implemented.

mschmidt87: 9e92d2814a90543959b922660f462987ddb538fc

test/test_individual.py
mschmidt87: 9e92d2814a90543959b922660f462987ddb538fc

test/test_individual.py
mschmidt87: 9e92d2814a90543959b922660f462987ddb538fc

test/test_individual.py
mschmidt87: 9e92d2814a90543959b922660f462987ddb538fc

test/test_individual.py
mschmidt87: I guess, I was being overly cautious or wanted to prepare possible future cases where we could also pass a list of Genome. Removed.

mschmidt87: 44f5d161a2fcdf1c96c6afc278c37b683f7f3675

test/test_individual.py
mschmidt87: This is a bit tricky but the problem is that we need the packed `f` for the `update_parameters_from_torch_class` call below.

test/test_individual.py
```diff
@@ -1,139 +1,165 @@
...
+        raise NotImplementedError
+
+
+def _unpack_evaluation(value, individual_type="SingleGenome"):
+    if individual_type == "SingleGenome":
+        return value
+    elif individual_type == "MultiGenome":
+        return value[0]
+    else:
+        raise NotImplementedError
```
jakobj, line 67:
```suggestion
        raise NotImplementedError()
```
it think you need to call the exception, don't you?

test/test_individual.py
```diff
@@ -1,139 +1,165 @@
...
+
+
+def _unpack_genome(individual, individual_type="SingleGenome"):
+    if individual_type == "SingleGenome":
+        return individual.genome
+    elif individual_type == "MultiGenome":
+        return individual.genome[0]
+    else:
+        raise NotImplementedError
```
jakobj, line 76: (see above)

test/test_individual.py
jakobj: ah, i see. and creating two instances is not possible, as you need the optimizer and the `update_...` function to work on the same one, right? ok, let's leave it like this.

test/test_individual.py
```diff
@@ -1,139 +1,165 @@
...
-    assert y[0, 0].item() == pytest.approx(x[0, 0].item() + c)
-    assert y[1, 0].item() == pytest.approx(x[1, 0].item() + c)
-
-
-def test_individual_with_parameter_sympy():
+@pytest.mark.parametrize("individual_type", ["SingleGenome", "MultiGenome"])
+@pytest.mark.parametrize("params", params_list)
+@pytest.mark.parametrize("func_args", func_args_list)
+def test_individual_with_parameter_sympy(individual_type, params, func_args):
     sympy = pytest.importorskip("sympy")  # noqa
```
jakobj, line 129:
```suggestion
    pytest.importorskip("sympy")
```

test/test_individual.py
```diff
@@ -1,139 +1,165 @@
...
 import pytest
 
 import cgp
-from cgp.individual import IndividualSingleGenome
+from cgp import IndividualSingleGenome, IndividualMultiGenome
 from cgp.genome import ID_INPUT_NODE, ID_OUTPUT_NODE, ID_NON_CODING_GENE
 
 
-def test_pickle_individual():
+TestParams = namedtuple("TestParams", ["genome_params", "primitives", "dna", "target_function"])
```
jakobj, line 13: why are you using a `namedtuple` instead of a dict here? is it much cleaner to offset the overhead of introducing it?

test/test_individual.py
```diff
@@ -1,139 +1,165 @@
...
+            0,
+            1,
+            ID_OUTPUT_NODE,
+            2,
+            ID_NON_CODING_GENE,
+        ],
+        target_function=lambda x, c: x + c,
+    )
+]
+FuncArgs = namedtuple("FuncArgs", ["x", "c"])
```
jakobj, line 41: (see above)
also, i'd rename this to something more specific like `graph_input_values`.

test/test_individual.py
```diff
@@ -1,139 +1,165 @@
...
+        raise NotImplementedError
+
+
+@pytest.mark.parametrize("individual_type", ["SingleGenome", "MultiGenome"])
+def test_pickle_individual(individual_type):
 
     primitives = (cgp.Add,)
     genome = cgp.Genome(1, 1, 1, 1, 1, primitives)
-    individual = IndividualSingleGenome(None, genome)
+    individual = _create_individual(None, genome, individual_type)
```
jakobj, line 84: maybe you could provide a default value for `fitness` s/t you don't need to provide `None`?

test/test_individual.py
jakobj: also, having this nice modularity, shouldn't we use it in the tests, i.e., introduce an additional "TestParams" instance?

test/test_individual.py
```diff
@@ -1,139 +1,165 @@
...
-        0,
-        0,
-        0,
-        1,
-        ID_OUTPUT_NODE,
-        2,
-        ID_NON_CODING_GENE,
-    ]
-    individual = IndividualSingleGenome(None, genome)
+    x, c = func_args.x, func_args.c
```
jakobj, line 98: i think an `assert` got lost that made sure than before setting the parameter in the next line, it's value is equal to the initial value, `1.0`. this also applies for more tests below.

test/test_individual.py
```diff
@@ -1,139 +1,165 @@
...
+
+    for xi in x:
+        y = f.subs("x_0", xi).evalf()
+        assert y == pytest.approx(target_function(xi, c))
 
-    f = individual.to_sympy()[0]
-    y = f.subs("x_0", x[0]).evalf()
 
-    assert y == pytest.approx(x[0] + c)
+@pytest.mark.parametrize("individual_type", ["SingleGenome", "MultiGenome"])
```
jakobj, line 143: good catch that this was missing! :ok_hand:

test/test_individual.py
mschmidt87: We could use a dict, of course, but I think this is the perfect use case for a `NamedTuple`. By using a `NamedTuple`, the code for unpacking it in the tests, gets shorter, because we can treat it as a tuple.

mschmidt87: > also, having this nice modularity, shouldn't we use it in the tests, i.e., introduce an additional "TestParams" instance?

I don't understand what you mean here, sorry.

test/test_individual.py
mschmidt87: Thanks, renamed. Note that namedtuples fall under classes when it comes to naming conventions, therefore `GraphInputValues`. 

Changed in 6cf310c53036ba0d3cb1057fc0a8d837bac938d8

test/test_individual.py
mschmidt87: No, `NotImplementedError` is a shortcut for `NotImplementedError()`. We just need to call it if we want to print a message.

test/test_individual.py
mschmidt87: see above.

test/test_individual.py
mschmidt87: why? I think it would be misleading to put a value here, if it is no needed.

test/test_individual.py
mschmidt87: Sorry, I don't see that `assert` in the old code (left side of here). As far as I can see, the asserts only tested for the function evaluation, e.g. in line 47 and 55 of the old version.

mschmidt87:
Thank you for your comments @jakobj , I addressed all of them, please take another look.

test/test_individual.py
jakobj: sorry, my comment wasn't clear enough: i mean in the `_create_individual` function you could provide a default value such that you don't need to provide anything here

test/test_individual.py
jakobj: well, you define a list of `TestParams` but right now this list only has one entry, e.g., tests only one graph version and target function. wouldn't it make sense to test multiple ones?

test/test_individual.py
jakobj: cool, i learned something! :) but wouldn't it make sense to provide a message? ;)

test/test_individual.py
jakobj: yes, you are right. i was misinterpreting the two consecutive asserts in the old cold: in both cases the parameter is assigned and then the output is checked. since you implemented this via `parametrized` i got a bit confused. since defaults are checked in `test_nodes`, never mind, everything's fine here :+1:

test/test_individual.py
mschmidt87: Fair point. Added a message. 27ff50681435a44ab24f87dc6efef799c55d26a3

test/test_individual.py
mschmidt87: Yes, but I would postpone that to a different PR. Let's not overload this one.

mschmidt87: Added #182 for this.

test/test_individual.py
mschmidt87: Yes, good idea, added that in 7772699715c2e633bdace8fda363eefe9f80dca9

mschmidt87:
Okay, I addressed the remaining comments.

I also added a change to the test utils to make this test more robust (currently failing in CI here), ff097b3397211b93396487f03650c1b22667e891

Also note: the coverage display is just wrong. I don't know what coveralls is doing here, but the coverage increased by this PR. We might just get rid of coveralls altogether, to be honest.

test/test_utils.py
```diff
@@ -54,7 +54,7 @@ def evaluate_objective_on_list(x):
     # second call should be faster as result is retrieved from cache
     t0 = time.time()
     evaluate_objective_on_list(x)
-    assert (time.time() - t0) < (sleep_time / 5.0)
+    assert (time.time() - t0) < pytest.approx(sleep_time / 5.0, rel=0.1)
```
jakobj, line 57: hmm, travis doesn't seem to like this, and i'd also choose a different route: we should just increase the `sleep_time` a bit, maybe to 2s. this test fails as travis only gives us a single core and running multiple threads in it seems to slow everything down to the point where the effect of caching is not visible any more.

mschmidt87: Okay, I reverted the change. Miraculously, the CI seems to pass this test now. I created this ticket to refactor the test: #185

mschmidt87:
I reverted the change and squashed all commits. Please approve @jakobj

jakobj:
good timing :) we can merge once travis is happy :+1:

PR #178, opened 2020-07-07
Enh/simplify node

### LINKED ISSUE #165: Simplify definition of nodes
There is a lot of duplicate code accumulating in `node.py`. Can we find a simpler, more compact way to define node operations that can be translated to the different compilation options (numpy, sympy etc) behind the scenes? It would be great if the definition of, for example, the addition and parameter nodes would reduce to this:
```python
class Add(Node):
    _output = "x_0 + x_1"

class Parameter(ParametrizedNode):
    _parameters = "p"
    _output = "p"
```

It's possible that we'll need to introduce new subclasses of `Node` such as `UnaryNode`, `BinaryNode` etc to implement this cleanly.

This would also allow us to easily add new compilation options without touching the definition of every single `Node` subclass.

jakobj:
This PR implements abstract classes that make the definition of new nodes much easier and more compact. New nodes, i.e., operators can now be introduced in two lines:

```python
class Add(BinaryOperatorNode):
    _def_output = "x_0 + x_1"
```

For operators that are "special", a few more lines are needed to define the different flavors for the `to_...` functions:
```python
class Sqrt(UnaryOperatorNode):
    _def_output = "math.sqrt(x_0)"
    _def_numpy_output = "np.sqrt(x_0)"
    _def_torch_output = "torch.sqrt(x_0)"
    _def_sympy_output = "sqrt(x_0)"
```
or
```python
class Parameter(NullaryParametrizedOperatorNode):
    _parameters = ("p",)
    _initial_values = (lambda: 1.0,)
    _def_output = "p"
```

Please have a look at the suggested implementation @mschmidt87 @HenrikMettler and let me know what you think. Is this approach sensible, user friendly, maintainable, flexible enough, etc pp?

Note that there still seems to be some duplication of code, however I am not sure how much further we can condense this while keeping the implementation (fairly) transparent.

closes #165


mschmidt87:
Looks like a really good approach. I just have a few remarks, some are more minor:
- the file `node.py` is quite long. What about distributing the code over multiple files, for instance:
    - `node.py` for the basic `Node` class
    - `node_derived.py` (or a better name) for the derived classes `Nullary`, `Unary`, `Binary`, etc.
    - `node_impl.py` for the concrete instances
- I feel like we need a good mechanism to catch syntax errors in the strings of the class definition. What I mean is: What if the user makes a mistake in e.g. the `_def_torch_output` string. Now, I think that this will only fail when the `to_torch` method is called, perhaps after the entire evolution has finished. It would good to either check this at instantiation time or implement a `check_class_definition` method that can be called to check if a node definition is valid.
- [minor]: I would reorder the definition of the classes, such that it is `Nullary` -> `Unary` -> `Binary`
- regarding duplicate code: I would opt for reducing duplicate code as much as possible. We can mitigate intransparencies with good function names etc. For instance, parsing the inputs in the class definition can be outsourced from the classes, because it is the same code (just with different numbers of inputs).

mschmidt87:
Oh and I would keep the definition of new nodes out of this PR, i.e. the `Sqrt` node.

jakobj:
Thanks for the input! I followed your suggestion and created a unified `NAryOperatorNode` from which all other nodes are derived. The arity is now only defined in concrete subclasses that implement certain operators. I've also split the definitions across multiple files and added checks to make sure the operator definitions make sense. Also removed the `Sqrt` node again. please have another look @mschmidt87

cgp/node.py
```diff
@@ -156,191 +159,124 @@ def format_parameter_str(self) -> None:
     def output_str(self) -> str:
         return self._output_str
 
-    @property
-    def parameter_str(self) -> str:
-        return self._parameter_str
 
+class NAryOperatorNode(Node):
```
mschmidt87, line 163: To improve our documentation, can you add a short docstring to this class that explains its purpose?

mschmidt87: Also, can we change the name to just `OperatorNode`? this `NAry` makes my eye stumble every time.

cgp/node_validation.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def check_to_numpy(cls: Type["NAryOperatorNode"]) -> None:
+    # delayed imports to avoid circular imports
+    from .cartesian_graph import CartesianGraph
+
+    genome = _create_genome(cls)
+
+    f = CartesianGraph(genome).to_numpy()
+    x = np.ones((3, 1)) * 1.5
```
mschmidt87, line 64: any reason for the magic 1.5?

test/test_node.py
```diff
@@ -17,7 +17,7 @@ def test_inputs_are_cut_to_match_arity():
     node = cgp.ConstantFloat(idx, inputs)
     assert node.inputs == []
 
-    node = cgp.node.OutputNode(idx, inputs)
+    node = cgp.node_input_output.OutputNode(idx, inputs)
```
mschmidt87, line 20: I would either directly import `OutputNode` and `InputNode` in `cgp.__init__.py` or rename the the file to `node_io`.

test/test_node.py
```diff
@@ -65,7 +65,7 @@ def _test_to_numpy(genome, x, y_target):
 def _test_to_torch(genome, x, y_target):
     torch = pytest.importorskip("torch")
     graph = cgp.CartesianGraph(genome)
-    assert graph.to_numpy()(torch.Tensor(x).reshape(1, -1)) == pytest.approx(y_target)
+    assert graph.to_torch()(torch.Tensor(x).reshape(1, -1)) == pytest.approx(y_target)
```
mschmidt87, line 68: Wow, is that on master??

jakobj: it shouldn't be, maybe i need to rebase?!

test/test_node.py
jakobj: as this is no user-facing code (users never create instances of `Input/OutputNode`s) i would actually leave this ugly, but clear module name and avoid the unnecessary import in `__init__`. are there other good reasons that i am missing?

cgp/node_validation.py
jakobj: hmm, no, it's a pretty number, no? i can use `3.14159`, if that's more obvious ;)

cgp/node.py
jakobj: sure, this is just legacy naming, since i was using `BinaryOperatorNode` etc before

jakobj: f18b8f87

cgp/node_validation.py
jakobj: 76015a61a

jakobj:
thanks for the input @mschmidt87! i've implemented some of your suggestions and improved the doc slightly. please have another look!

test/test_node.py
mschmidt87: Okay, that's okay. I would still prefer to rename the file to `node_io.py`, since `io` is a common abbreviation for `input_output`. But it's a matter of taste, I guess.

cgp/node_validation.py
mschmidt87: why not just `1.0`, i.e., no number at all?

cgp/node.py
mschmidt87: Yep, sure, thanks for the change.

cgp/node_validation.py
mschmidt87: I think this is more maintainable because it does not raise questions like mine.

jakobj: i generally don't like using `1.0` since it can lead to false positives: for example imagine a test that should use a constant node (which by default has output 1.0) but mistakenly forwards the input, or a test that should use multiplication but mistakenly uses division, and so on. by using "non-trivial" values we can avoid some of these problems. maybe we should introduce a `magic_constant` in conftest with such an explanation?

mschmidt87: Okay, that makes sense, thanks for the explanation. Are we using this in other places, too? If yes, than a fixture would be a good idea. If not, just adding a short explanation here would suffice, I think.

jakobj:
ok, squashed my commits and removed the magic number again, it doesn't make sense here, you were right! let's see if travis is happy, then we can merge :+1:

Issue #173, opened 2020-07-03
`only_silent_mutations` has false negatives


HenrikMettler:
The boolean `silent` in `Genome._mutate_output_region` and `Genome._mutate_hidden_region` does not take into account whether there was a change in Dna. This can lead to cases where the boolean `only_silent_mutations` is set to false, despite the effective mutations (the ones that changed the Dna) all being silent. 

Possible fix: Remove the current value from the list `permissible_inputs` - so every mutation would effectively lead to a change in Dna. (with this the loop in `mutate` could also be simplified)

Issue #124, opened 2020-05-29
Number of mutations should not be fixed to average


jakobj:
By definition each gene has a certain probability of mutation, leading to a binomial distribution for the number of mutations per individual per generation. Instead of deciding for each gene whether it should be mutated, we calculate the number of mutations in advance, currently just using the average of the distribution (https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/individual.py#L51) as pointed out by @HenrikMettler. A reasonable fix would be to draw the number of mutations from the correct distribution.

What do you think @mschmidt87?

mschmidt87:
Yeo, I agree. To be honest, I always found this a bit strange.

jakobj:
great. @HenrikMettler do you want to give this a try?

HenrikMettler:
Yes, I will. How would you parameterize the distribution? As a binomial
with p=mut_rate and n=num_genes?

Am Fr., 29. Mai 2020 um 13:25 Uhr schrieb Jakob Jordan <
notifications@github.com>:

jakobj:
yes, exactly, i would just use https://numpy.org/devdocs/reference/random/generated/numpy.random.binomial.html. please also add a corresponding test. maybe even start with the test (https://en.wikipedia.org/wiki/Test-driven_development) ;)

PR #180, opened 2020-07-10
Genome mutate - implements new logic for mutations

### LINKED ISSUE #172: Issue warning when `mutation_rate` is high and `mutate` potentially takes long
#157 made sure that exactly the correct number of mutations occurs in a genome by checking that the mutated genome differs from the original one in `n_mutations` position. when using a high mutation rate and small number of primitives, this can lead to a long runtime for `mutate` as it struggles to find positions in the genome to mutate and alternative genes to insert. we should figure out exactly under which conditions this tends to occur and issue a user warning when we detect such a situation.

@HenrikMettler, since you're already familiar with that part of the code, maybe you want to look into this?

HenrikMettler:
Implements the new logic for mutation discussed in #172. Closes #172 , #173 and makes #124 obsolete. Does not pass  `test_parallel_population` in test/test_hl_api.py, since fitnes is not the with 1 or 2 parallel processes. I created the PR anyway since I couldn't figure out why and whether this is a problem with the new logic in mutate

### DIFF
--- a/cgp/genome.py
+++ b/cgp/genome.py
@@ -107,7 +107,7 @@ class Genome:
     def dna(self, value: List[int]) -> None:
         self._validate_dna(value)
         self._dna = value
-        self._initialize_unkown_parameters()
+        self._initialize_unknown_parameters()
 
     @property
     def _n_hidden(self) -> int:
@@ -332,6 +332,9 @@ class Genome:
     def _is_function_gene(self, gene_idx: int) -> bool:
         return (gene_idx % self._length_per_region) == 0
 
+    def _is_hidden_input_gene(self, gene_idx: int, region_idx: int) -> bool:
+        return self._is_hidden_region(region_idx) & ((gene_idx % self._length_per_region) != 0)
+
     def _is_active_input_gene(self, gene_idx: int) -> bool:
         input_index = gene_idx % self._length_per_region
         assert input_index > 0
@@ -346,15 +349,37 @@ class Genome:
         else:
             assert False  # should never be reached
 
+    def _select_gene_indices_for_mutation(
+        self, mutation_rate: float, rng: np.random.RandomState
+    ) -> List[int]:
+        """Selects the gene indices for mutations
+
+        Parameters
+        ----------
+        mutation_rate : float
+            Probability of a gene to be mutated, between 0 (excluded) and 1 (included).
+        rng : numpy.random.RandomState
+            Random number generator instance to use for selecting the indices.
+
+        Returns
+        ----------
+        selected_gene_indices: np.ndarray
+            indices of the genes selected for mutation.
+        """
+
+        selected_gene_indices = np.nonzero(rng.rand(len(self.dna)) < mutation_rate)[0]
+        return selected_gene_indices
+
     def mutate(self, mutation_rate: float, rng: np.random.RandomState):
         """Mutate the genome.
 
         Parameters
         ----------
         mutation_rate : float
-            Proportion of genes to be mutated, between 0 and 1.
+            Probability of a gene to be mutated, between 0 (excluded) and 1 (included).
         rng : numpy.random.RandomState
             Random number generator instance to use for mutating.
+
 
         Returns
         ----------
@@ -362,71 +387,75 @@ class Genome:
             True if only inactive regions of the genome were mutated, False otherwise.
         """
 
-        def count_dna_differences(dna0: List[int], dna1: List[int]) -> int:
-            return len([1 for gene0, gene1 in zip(dna0, dna1) if gene0 != gene1])
-
-        n_mutations = int(mutation_rate * len(self.dna))
-        assert n_mutations > 0
-
         graph = CartesianGraph(self)
         active_regions = graph.determine_active_regions()
-
         dna = list(self._dna)
-
         only_silent_mutations = True
-        while count_dna_differences(self.dna, dna) < n_mutations:
 
-            gene_idx = rng.randint(0, self._n_genes)
+        selected_gene_indices = self._select_gene_indices_for_mutation(mutation_rate, rng)
+
+        for (gene_idx, allele) in zip(selected_gene_indices, np.array(dna)[selected_gene_indices]):
+
             region_idx = gene_idx // self._length_per_region
 
-            if self._is_input_region(region_idx):
-                continue  # nothing to do here
+            permissible_values = self._determine_alternative_permissible_values(
+                gene_idx, allele, region_idx
+            )
+            if len(permissible_values) > 0:
 
-            elif self._is_output_region(region_idx):
-                silent = self._mutate_output_region(dna, gene_idx, rng)
+                dna[gene_idx] = rng.choice(permissible_values)
+                silent = region_idx not in active_regions
 
-            elif self._is_hidden_region(region_idx):
-                silent = self._mutate_hidden_region(dna, gene_idx, active_regions, rng)
-
-            else:
-                assert False  # should never be reached
-
-            only_silent_mutations = only_silent_mutations and silent
+                only_silent_mutations = only_silent_mutations and silent
 
         self.dna = dna
         return only_silent_mutations
 
-    def _mutate_output_region(
-        self, dna: List[int], gene_idx: int, rng: np.random.RandomState
-    ) -> bool:
-        assert self._is_gene_in_output_region(gene_idx)
+    def _determine_alternative_permissible_values(
+        self, gene_idx: int, allele: int, region_idx: int
+    ) -> List[int]:
 
-        if not self._is_function_gene(gene_idx) and self._is_active_input_gene(gene_idx):
-            permissible_inputs = self._permissible_inputs_for_output_region()
-            dna[gene_idx] = rng.choice(permissible_inputs)
-            return False
+        if self._is_input_region(region_idx):
+            return []  # genes in input regions have no alternative permissible values
+
+        elif self._is_hidden_region(region_idx):
+            return self._determine_alternative_permissible_values_hidden_gene(
+                gene_idx, allele, region_idx
+            )
+
+        elif self._is_output_region(region_idx):
+            return self._determine_alternative_permissible_values_output_gene(gene_idx, allele)
+
         else:
-            return True
+            assert False  # should never be reached
 
-    def _mutate_hidden_region(
-        self, dna: List[int], gene_idx: int, active_regions: List[int], rng: np.random.RandomState
-    ) -> bool:
-
-        assert self._is_gene_in_hidden_region(gene_idx)
-
-        region_idx = gene_idx // self._length_per_region
-        silent_mutation = region_idx not in active_regions
-
+    def _determine_alternative_permissible_values_hidden_gene(
+        self, gene_idx: int, allele: int, region_idx: int
+    ) -> List[int]:
         if self._is_function_gene(gene_idx):
-            dna[gene_idx] = self._primitives.sample_allele(rng)
-            return silent_mutation
+            permissible_values = list(np.arange(len(self._primitives._primitives)))
+
+        elif self._is_hidden_input_gene(gene_idx, region_idx):
+            permissible_values = self._permissible_inputs(region_idx)
 
         else:
-            permissible_inputs = self._permissible_inputs(region_idx)
-            dna[gene_idx] = rng.choice(permissible_inputs)
+            assert False
+        permissible_values.remove(allele)
+        return permissible_values
 
-            silent_mutation = silent_mutation or (not self._is_active_input_gene(gene_idx))
-            return silent_mutation
+    def _determine_alternative_permissible_values_output_gene(
+        self, gene_idx: int, allele: int
+    ) -> List[int]:
+        input_index = (
+            gene_idx % self._length_per_region
+        )  # assumes that the second gene in all output regions is the index of the input
+        if input_index == 1:
+            permissible_values = self._permissible_inputs_for_output_region()
+            permissible_values.remove(allele)
+        else:
+            permissible_values = []
+
+        return permissible_values
 
     @property
     def primitives(self) -> Primitives:
@@ -484,7 +513,7 @@ class Genome:
 
         return any_parameter_updated
 
-    def _initialize_unkown_parameters(self) -> None:
+    def _initialize_unknown_parameters(self) -> None:
         for region_idx, region in self.iter_hidden_regions():
             node_id = region[0]
             node_type = self._primitives[node_id]
--- a/cgp/population.py
+++ b/cgp/population.py
@@ -25,8 +25,7 @@ class Population:
         n_parents : int
             Number of parent individuals.
         mutation_rate : float
-            Rate of mutations determining the number of genes to be
-            mutated for offspring creation, between 0 and 1.
+            Probability of a gene to be mutated, between 0 (excluded) and 1 (included).
         seed : int
             Seed for internal random number generator.
         genome_params : dict
@@ -34,8 +33,8 @@ class Population:
         """
         self.n_parents = n_parents  # number of individuals in parent population
 
-        if not (0.0 < mutation_rate and mutation_rate < 1.0):
-            raise ValueError("mutation rate needs to be in (0, 1)")
+        if not (0.0 < mutation_rate and mutation_rate <= 1.0):
+            raise ValueError("mutation rate needs to be in (0, 1]")
         self._mutation_rate = mutation_rate  # probability of mutation per gene
 
         self.seed = seed
--- a/test/test_genome.py
+++ b/test/test_genome.py
@@ -3,6 +3,7 @@ import pytest
 
 import cgp
 from cgp.genome import ID_INPUT_NODE, ID_OUTPUT_NODE, ID_NON_CODING_GENE
+from cgp.cartesian_graph import CartesianGraph
 
 
 def test_check_dna_consistency():
@@ -346,89 +347,205 @@ def test_is_gene_in_output_region(rng_seed):
     assert not genome._is_gene_in_output_region(11)
 
 
-def test_mutate_hidden_region(rng_seed):
-    rng = np.random.RandomState(rng_seed)
-    genome = cgp.Genome(1, 1, 3, 1, None, (cgp.Add, cgp.ConstantFloat))
-    dna = [
-        ID_INPUT_NODE,
-        ID_NON_CODING_GENE,
-        ID_NON_CODING_GENE,
-        1,
-        0,
-        0,
-        1,
-        0,
-        0,
-        0,
-        0,
-        2,
-        ID_OUTPUT_NODE,
-        3,
-        ID_NON_CODING_GENE,
-    ]
-    genome.dna = list(dna)
-    active_regions = cgp.CartesianGraph(genome).determine_active_regions()
-
-    # mutating any gene in inactive region returns True
-    assert genome._mutate_hidden_region(list(dna), 3, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 4, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 5, active_regions, rng) is True
-
-    # mutating function gene in active region returns False
-    assert genome._mutate_hidden_region(list(dna), 6, active_regions, rng) is False
-    # mutating inactive genes in active region returns True
-    assert genome._mutate_hidden_region(list(dna), 7, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 8, active_regions, rng) is True
-
-    # mutating any gene in active region without silent genes returns False
-    assert genome._mutate_hidden_region(list(dna), 9, active_regions, rng) is False
-    assert genome._mutate_hidden_region(list(dna), 10, active_regions, rng) is False
-    assert genome._mutate_hidden_region(list(dna), 11, active_regions, rng) is False
-
-
-def test_mutate_output_region(rng_seed):
-    rng = np.random.RandomState(rng_seed)
-    genome = cgp.Genome(1, 1, 2, 1, None, (cgp.Add,))
-    dna = [
-        ID_INPUT_NODE,
-        ID_NON_CODING_GENE,
-        ID_NON_CODING_GENE,
-        0,
-        0,
-        0,
-        0,
-        0,
-        0,
-        ID_OUTPUT_NODE,
-        2,
-        ID_NON_CODING_GENE,
-    ]
-
-    assert genome._mutate_output_region(list(dna), 9, rng) is True
-    assert genome._mutate_output_region(list(dna), 10, rng) is False
-    assert genome._mutate_output_region(list(dna), 11, rng) is True
-
-
-@pytest.mark.parametrize("mutation_rate", [0.02, 0.05, 0.2])
-def test_correct_number_of_mutations(mutation_rate, rng_seed):
-
-    n_inputs = 2
+def test_mutation_rate(rng_seed, mutation_rate):
+    n_inputs = 1
     n_outputs = 1
-    n_columns = 10
-    n_rows = 2
-    levels_back = 5
-    primitives = (cgp.Add, cgp.Sub, cgp.Mul, cgp.Div, cgp.ConstantFloat)
-
+    n_columns = 4
+    n_rows = 3
+    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, None, (cgp.Add, cgp.Sub))
     rng = np.random.RandomState(rng_seed)
-    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives)
     genome.randomize(rng)
 
-    n_mutations = 0
-    genome_new = genome.clone()
-    genome_new.mutate(mutation_rate, rng)
-    for (gene_0, gene_1) in zip(genome.dna, genome_new.dna):
-        if gene_0 != gene_1:
-            n_mutations += 1
+    def count_n_immutable_genes(n_inputs, n_outputs, n_rows):
+        length_per_region = genome.primitives.max_arity + 1  # function gene + input gene addresses
+        n_immutable_genes = n_inputs * length_per_region  # none of the input genes are mutable
+        n_immutable_genes += n_outputs * (
+            length_per_region - 1
+        )  # only one gene per output can be mutated
+        if n_inputs == 1:
+            n_immutable_genes += (
+                n_rows * genome.primitives.max_arity
+            )  # input gene addresses in the first (hidden) column can't be mutated
+            # if only one input node exists
+        return n_immutable_genes
 
-    n_mutations_expected = int(mutation_rate * len(genome.dna))
-    assert n_mutations == n_mutations_expected
+    def count_mutations(dna0, dna1):
+        n_differences = 0
+        for (allele0, allele1) in zip(dna0, dna1):
+            if allele0 != allele1:
+                n_differences += 1
+        return n_differences
+
+    n_immutable_genes = count_n_immutable_genes(n_inputs, n_outputs, n_rows)
+    n_mutations_mean_expected = mutation_rate * (len(genome.dna) - n_immutable_genes)
+    n_mutations_std_expected = np.sqrt(
+        (len(genome.dna) - n_immutable_genes) * mutation_rate * (1 - mutation_rate)
+    )
+
+    n = 10000
+    n_mutations = []
+    for _ in range(n):
+        dna_old = genome.dna
+        genome.mutate(mutation_rate, rng)
+        n_mutations.append(count_mutations(dna_old, genome.dna))
+
+    assert np.mean(n_mutations) == pytest.approx(n_mutations_mean_expected, rel=0.04)
+    assert np.std(n_mutations) == pytest.approx(n_mutations_std_expected, rel=0.04)
+
+
+def test_only_silent_mutations(genome_params, mutation_rate, rng_seed):
+    genome = cgp.Genome(**genome_params)
+    rng = np.random.RandomState(rng_seed)
+    genome.randomize(rng)
+
+    only_silent_mutations = genome.mutate(mutation_rate=0, rng=rng)
+    assert only_silent_mutations is True
+
+    only_silent_mutations = genome.mutate(mutation_rate=1, rng=rng)
+    assert not only_silent_mutations
+
+    dna_fixed = [
+        ID_INPUT_NODE,
+        ID_NON_CODING_GENE,
+        ID_NON_CODING_GENE,
+        ID_INPUT_NODE,
+        ID_NON_CODING_GENE,
+        ID_NON_CODING_GENE,
+        2,
+        1,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        ID_OUTPUT_NODE,
+        2,
+        ID_NON_CODING_GENE,
+    ]
+    genome.dna = dna_fixed
+    graph = CartesianGraph(genome)
+    active_regions = graph.determine_active_regions()
+    length_per_region = genome.primitives.max_arity + 1  # function gene + input gene addresses
+    gene_to_be_mutated_non_active = (
+        3 * length_per_region
+    )  # operator gene of 2nd hidden node, a silent node in this dna configuration
+
+    def select_gene_indices_silent(mutation_rate, rng):
+        selected_gene_indices = [gene_to_be_mutated_non_active]
+        return selected_gene_indices
+
+    genome._select_gene_indices_for_mutation = (
+        select_gene_indices_silent  # monkey patch the selection of indices to select a silent gene
+    )
+    genome.dna = dna_fixed
+    only_silent_mutations = genome.mutate(mutation_rate, rng)
+    assert only_silent_mutations is True
+
+    gene_to_be_mutated_active = (
+        active_regions[-1] * length_per_region
+    )  # operator gene of the 1st active hidden node,
+    # should always be mutable and result in a non-silent-mutation
+
+    def select_gene_indices_non_silent(mutation_rate, rng):
+        selected_gene_indices = [gene_to_be_mutated_active]
+        return selected_gene_indices
+
+    genome._select_gene_indices_for_mutation = select_gene_indices_non_silent
+    # monkey patch the selection of indices to select a non-silent gene
+    only_silent_mutations = genome.mutate(mutation_rate, rng)
+    assert not only_silent_mutations
+
+
+def test_permissible_values_hidden(rng_seed):
+    genome_params = {
+        "n_inputs": 2,
+        "n_outputs": 1,
+        "n_columns": 3,
+        "n_rows": 3,
+        "levels_back": 2,
+        "primitives": (cgp.Add, cgp.Sub, cgp.ConstantFloat),
+    }
+
+    genome = cgp.Genome(**genome_params)
+
+    # test function gene
+    gene_idx = 6  # function gene of first hidden region
+    allele = 0  #
+    region_idx = None  # can be any number, since not touched for function gene
+    permissible_func_gene_values = genome._determine_alternative_permissible_values_hidden_gene(
+        gene_idx, allele, region_idx
+    )
+    assert permissible_func_gene_values == [
+        1,
+        2,
+    ]  # function idx 1,2 (cgp.Sub, cgp.ConstantFloat)
+
+    # test input gene
+    gene_idx = 16  # first input gene in second column of hidden region
+    allele = 0
+    region_idx = gene_idx // (
+        genome.primitives.max_arity + 1
+    )  # function gene + input gene addresses
+    permissible_input_gene_values = genome._determine_alternative_permissible_values_hidden_gene(
+        gene_idx, allele, region_idx
+    )
+
+    assert permissible_input_gene_values == [
+        1,
+        2,
+        3,
+        4,
+    ]  # gene indices of input, 1st hidden layer
+
+
+def test_permissible_values_output(rng_seed):
+    genome_params = {
+        "n_inputs": 2,
+        "n_outputs": 1,
+        "n_columns": 3,
+        "n_rows": 3,
+        "levels_back": 2,
+        "primitives": (cgp.Add, cgp.Sub, cgp.ConstantFloat),
+    }
+    genome = cgp.Genome(**genome_params)
+
+    gene_idx_function = 33
+    gene_idx_input0 = 34
+    gene_idx_input1 = 35
+
+    allele = 5  # set current value for input0 -> should be excluded from permissible values
+
+    permissible_values = genome._determine_alternative_permissible_values_output_gene(
+        gene_idx_function, allele
+    )
+    assert permissible_values == []
+
+    permissible_values = genome._determine_alternative_permissible_values_output_gene(
+        gene_idx_input0, allele
+    )
+    assert permissible_values == [0, 1, 2, 3, 4, 6, 7, 8, 9, 10]
+
+    permissible_values = genome._determine_alternative_permissible_values_output_gene(
+        gene_idx_input1, allele
+    )
+    assert permissible_values == []
--- a/test/test_hl_api.py
+++ b/test/test_hl_api.py
@@ -65,6 +65,7 @@ def test_parallel_population(population_params, genome_params, ea_params):
         fitness_per_n_processes.append(
             _test_population(population_params, genome_params, ea_params)
         )
+
     assert fitness_per_n_processes[0] == pytest.approx(fitness_per_n_processes[1])
     assert fitness_per_n_processes[0] == pytest.approx(fitness_per_n_processes[2])
 
--- a/test/test_ea_mu_plus_lambda.py
+++ b/test/test_ea_mu_plus_lambda.py
@@ -196,6 +196,8 @@ def test_create_new_offspring_and_parent_generation(population_params, genome_pa
         individual.fitness = float(individual.idx)
         return individual
 
+    population_params["mutation_rate"] = 1.0  # ensures every offspring has mutations
+
     pop = cgp.Population(**population_params, genome_params=genome_params)
     ea = cgp.ea.MuPlusLambda(**ea_params)
 


cgp/genome.py
```diff
@@ -354,77 +357,75 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
-                success = self._mutate_output_region(dna, gene_idx, rng)
-                if success:
-                    silent = False
-                    only_silent_mutations = only_silent_mutations and silent
-                    successful_mutations += 1
+            silent = True
+            region_idx = gene_idx // self._length_per_region
+
+            permissible_values = self._determine_permissible_values(gene_idx, gene, region_idx)
+            if not len(permissible_values) == 0:
```
mschmidt87, line 382: simpler check:

mschmidt87:
```suggestion
            if len(permissible_values) > 0:
```

cgp/genome.py
```diff
@@ -354,77 +357,75 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
+            return []  # input region have no other permissible values
 
-        assert self._is_gene_in_hidden_region(gene_idx)
+        elif self._is_hidden_region(region_idx):
+            return self._determine_permissible_values_hidden(gene_idx, gene, region_idx)
 
-        region_idx = gene_idx // self._length_per_region
-        silent_mutation = region_idx not in active_regions
+        else:
+            assert self._is_output_region(region_idx)
```
mschmidt87, line 403: why not another `elif` here, i.e. `elif self._is_output_region(region_idx)`?

cgp/genome.py
```diff
@@ -354,77 +357,75 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
+    ) -> List[int]:
         if self._is_function_gene(gene_idx):
-            dna[gene_idx] = self._primitives.sample_allele(rng)
-            return silent_mutation
+            permissible_values = list(np.arange(len(self._primitives._primitives)))
 
         else:
-            permissible_inputs = self._permissible_inputs(region_idx)
-            dna[gene_idx] = rng.choice(permissible_inputs)
+            assert self._is_hidden_input_gene(gene_idx, region_idx)
```
mschmidt87, line 413: same here, why not `elif self._is_hidden_input_gene(gene_idx, region_idx)`

cgp/genome.py
HenrikMettler: Just because it is shorter than `elif self._is_output_region(region_idx)` and `else False`

mschmidt87: I would prefer having an `elif` and `else` statement here, because this is more readable and maintainable. Same below.

cgp/genome.py
```diff
@@ -354,79 +357,79 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
         only_silent_mutations = True
-        while count_dna_differences(self.dna, dna) < n_mutations:
 
-            gene_idx = rng.randint(0, self._n_genes)
-            region_idx = gene_idx // self._length_per_region
+        selected_gene_indices = np.nonzero(np.random.rand(len(dna)) < mutation_rate)
 
-            if self._is_input_region(region_idx):
-                continue  # nothing to do here
+        for (gene_idx, gene) in zip(selected_gene_indices[0], dna_array[selected_gene_indices]):
```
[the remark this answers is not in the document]
jakobj: also, what you call `gene` here, is technically an `allele`, i.e., a realization of a gene. maybe we should try to stick to this nomenclature.

cgp/genome.py
```diff
@@ -354,79 +357,79 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
+    ) -> List[int]:
 
-        if not self._is_function_gene(gene_idx) and self._is_active_input_gene(gene_idx):
-            permissible_inputs = self._permissible_inputs_for_output_region()
-            dna[gene_idx] = rng.choice(permissible_inputs)
-            return False
-        else:
-            return True
+        if self._is_input_region(region_idx):
+            return []  # input region have no other permissible values
```
jakobj, line 397:
```suggestion
            return []  # genes in input regions have no alternative permissible values
```

cgp/genome.py
```diff
@@ -354,79 +357,79 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
             only_silent_mutations = only_silent_mutations and silent
 
         self.dna = dna
         return only_silent_mutations
 
-    def _mutate_output_region(
-        self, dna: List[int], gene_idx: int, rng: np.random.RandomState
-    ) -> bool:
-        assert self._is_gene_in_output_region(gene_idx)
+    def _determine_permissible_values(
```
jakobj, line 392:
```suggestion
    def _determine_alternative_permissible_values(
```

cgp/genome.py
```diff
@@ -354,79 +357,79 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
-        else:
-            return True
+        if self._is_input_region(region_idx):
+            return []  # input region have no other permissible values
 
-    def _mutate_hidden_region(
-        self, dna: List[int], gene_idx: int, active_regions: List[int], rng: np.random.RandomState
-    ) -> bool:
+        elif self._is_hidden_region(region_idx):
+            return self._determine_permissible_values_hidden(gene_idx, gene, region_idx)
```
jakobj, line 400:
```suggestion
            return self._determine_permissible_values_hidden_gene(gene_idx, gene, region_idx)
```

cgp/genome.py
```diff
@@ -354,79 +357,79 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
 
-        assert self._is_gene_in_hidden_region(gene_idx)
+        elif self._is_output_region(region_idx):
+            return self._determine_permissible_values_output(gene_idx, gene)
```
jakobj, line 403:
```suggestion
            return self._determine_permissible_values_output_gene(gene_idx, gene)
```

cgp/genome.py
```diff
@@ -354,79 +357,79 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
+
+        else:
+            assert False
+        permissible_values.remove(gene)
+        return permissible_values
+
+    def _determine_permissible_values_output(self, gene_idx: int, gene: int) -> List[int]:
+        input_index = (
+            gene_idx % self._length_per_region
+        )  # assumes that the second gene in the output is the address of the input
```
jakobj, line 425:
```suggestion
        )  # assumes that the second gene in all output regions is the index of the input
```

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
-    genome.dna = list(dna)
-    active_regions = cgp.CartesianGraph(genome).determine_active_regions()
+    genome.randomize(rng)
 
-    # mutating any gene in inactive region returns True
-    assert genome._mutate_hidden_region(list(dna), 3, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 4, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 5, active_regions, rng) is True
+    def count_n_immutable_genes(n_input, n_output, n_row):
+        counter = n_input * 3  # 3 is the arity
```
jakobj, line 360:
```suggestion
        n_immutable_genes = n_input * genome.primitives.max_arity
```

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
     assert not genome._is_gene_in_output_region(11)
 
 
-def test_mutate_hidden_region(rng_seed):
+def test_mutation_rate(rng_seed, mutation_rate):
+    n_input = 1
+    n_output = 1
+    n_column = 4
+    n_row = 3
```
jakobj, line 354:
```suggestion
    n_inputs = 1
    n_outputs = 1
    n_columns = 4
    n_rows = 3
```

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
-    active_regions = cgp.CartesianGraph(genome).determine_active_regions()
+    genome.randomize(rng)
 
-    # mutating any gene in inactive region returns True
-    assert genome._mutate_hidden_region(list(dna), 3, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 4, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 5, active_regions, rng) is True
+    def count_n_immutable_genes(n_input, n_output, n_row):
+        counter = n_input * 3  # 3 is the arity
+        counter += n_output * 2  # 2 of 3 output genes can not be mutated
```
jakobj, line 361:
```suggestion
        n_immutable_genes += n_output * (genome.primitives.max_arity)  # only one gene per output can be mutated
```

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
+        if n_input == 1:
```
jakobj, line 362: what happens for `n_input != 1`?

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
+        counter = 0
+        for idx in range(len(prev_dna)):
+            if prev_dna[idx] != new_dna[idx]:
+                counter += 1
+        return counter
+
+    acceptable_error_interval = 0.05  # Todo: get comment for this value from Jakob and Max
+    n = 1000
+    n_immutable_genes = count_n_immutable_genes(n_input, n_output, n_row)
+    n_mutations_expected = n * mutation_rate * (len(genome.dna) - n_immutable_genes)
```
jakobj, line 378: i would not include `n` here, rather take mean further down

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
+    print(len(genome.dna), n_immutable_genes)
```
jakobj, line 379:
```suggestion
```

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
+    n_mutations = 0
+    for _ in range(n):
+        temp_dna = genome.dna
+        genome.mutate(mutation_rate, rng)
+        current_n_mutations = count_mutations(temp_dna, genome.dna)
+        n_mutations += current_n_mutations
+
+    assert abs(n_mutations_expected - n_mutations) < (
+        n_mutations_expected * acceptable_error_interval
+    )
```
jakobj, line 389:
```suggestion
    n_mutations = []
    for _ in range(n):
        dna_old = genome.dna
        genome.mutate(mutation_rate, rng)
        n_mutations.append(count_mutations(dna_old, genome.dna))

    assert np.mean(n_mutations) == pytest.approx(n_mutations_mean_expected)
    assert np.std(n_mutations) == pytest.approx(n_mutations_std_expected)
```
where you'd compute the mean and standard deviation from the binomial distribution.

test/test_hl_api.py
```diff
@@ -65,6 +65,8 @@ def test_parallel_population(population_params, genome_params, ea_params):
         fitness_per_n_processes.append(
             _test_population(population_params, genome_params, ea_params)
         )
+
```
jakobj, line 68: why do you need this?

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
-        0,
-        0,
-        0,
-        0,
-        ID_OUTPUT_NODE,
+    only_silent_mutations_1 = genome.mutate(mutation_rate=1, rng=rng)
+    assert not only_silent_mutations_1
+
+
+# Overwrite global genome_params() from conftest,
```
jakobj, line 404: i'd not overwrite it globally, but rather introduce the dict in each of the tests as a local variable

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
-    assert genome._mutate_hidden_region(list(dna), 6, active_regions, rng) is False
-    # mutating inactive genes in active region returns True
-    assert genome._mutate_hidden_region(list(dna), 7, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 8, active_regions, rng) is True
 
-    # mutating any gene in active region without silent genes returns False
-    assert genome._mutate_hidden_region(list(dna), 9, active_regions, rng) is False
-    assert genome._mutate_hidden_region(list(dna), 10, active_regions, rng) is False
-    assert genome._mutate_hidden_region(list(dna), 11, active_regions, rng) is False
+def test_only_silent_mutations(genome_params, rng_seed):
```
jakobj, line 392: can we somehow test that mutating specific genes returns the correct value for `only_silent_mutations` in addition to this brute force approach?

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
+        "levels_back": 2,
+        "primitives": (cgp.Add, cgp.Sub, cgp.ConstantFloat),
+    }
+
+
+def test_permissible_values_hidden(genome_params, rng_seed):
+    genome = cgp.Genome(**genome_params)
+
+    # test function gene
+    gene_idx = 6  # function gene of first hidden gene
```
jakobj, line 422:
```suggestion
    gene_idx = 6  # function gene of first hidden region
```
?

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
+    gene = 0  #
+    region_idx = 1000  # can be any number, since not touched for function gene
```
jakobj, line 424: i'd rather use `None` then

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
+
+    # test input gene
+    gene_idx = 16  # first input gene in second hidden layer
+    gene = 0
+    region_idx = 5
+    permissible_hidden_input_gene_values = genome._determine_permissible_values_hidden(
+        gene_idx, gene, region_idx
+    )
+
+    assert permissible_function_gene_values == [
```
jakobj, line 437: i would move this up to line 428

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
+    # test function gene
+    gene_idx = 6  # function gene of first hidden gene
+    gene = 0  #
+    region_idx = 1000  # can be any number, since not touched for function gene
+    permissible_function_gene_values = genome._determine_permissible_values_hidden(
+        gene_idx, gene, region_idx
+    )
+
+    # test input gene
+    gene_idx = 16  # first input gene in second hidden layer
```
jakobj, line 430:
```suggestion
    gene_idx = 16  # first input gene in second hidden region
```
?

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
+    gene = 0
+    region_idx = 5
```
jakobj, line 432: can't you compute this from the gene idx?

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
-    assert genome._mutate_output_region(list(dna), 11, rng) is True
 
+def test_permissible_values_output(genome_params, rng_seed):
+    genome = cgp.Genome(**genome_params)
 
-@pytest.mark.parametrize("mutation_rate", [0.02, 0.05, 0.2])
-def test_correct_number_of_mutations(mutation_rate, rng_seed):
+    gene_idx_0 = 33
+    gene_idx_1 = 34
+    gene_idx_2 = 35
```
jakobj, line 454:
```suggestion
    gene_idx_function = 33
    gene_idx_input0 = 34
    gene_idx_input1 = 35
```
just to make this more easily understandable

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
 
-    n_inputs = 2
-    n_outputs = 1
-    n_columns = 10
-    n_rows = 2
-    levels_back = 5
-    primitives = (cgp.Add, cgp.Sub, cgp.Mul, cgp.Div, cgp.ConstantFloat)
+    gene = 5  # possible input genes: 0-10
```
jakobj, line 456: comment is confusing here

test/test_genome.py
```diff
@@ -346,89 +347,118 @@ def test_is_gene_in_output_region(rng_seed):
...
-    n_mutations = 0
-    genome_new = genome.clone()
-    genome_new.mutate(mutation_rate, rng)
-    for (gene_0, gene_1) in zip(genome.dna, genome_new.dna):
-        if gene_0 != gene_1:
-            n_mutations += 1
-
-    n_mutations_expected = int(mutation_rate * len(genome.dna))
-    assert n_mutations == n_mutations_expected
+    assert permissible_values_0 == []
```
jakobj, line 462: as above, i'd put the asserts right under the corresponding function calls, this way tests fail early

jakobj:
> [snip] Does not pass `test_parallel_population` in test/test_hl_api.py, since fitnes is not the with 1 or 2 parallel processes. I created the PR anyway since I couldn't figure out why and whether this is a problem with the new logic in mutate

the fitness should be identical, independent of the number of processes as far as I remember. this is obviously desirable from a reproducibility viewpoint. is it possible that any of the stochastic operations are not using the internal rng of the population? that would be my first guess.

test/test_genome.py
HenrikMettler: Is the arity of the primitives correct here? Isn't the "full" arity genome.primitives.max_arity + 1 (for the function gene?)

test/test_genome.py
HenrikMettler: `counter` remains on the previous value. If I am not mistaking only in the case of `n_input = 1` the input gene addresses can not be mutated (for hidden genes in the first column). If `n_input >= 2` they have more than one permissible value and are therefore mutable.

test/test_genome.py
HenrikMettler: eg: `region_idx = int(gene_idx / (genome.primitives.max_arity + 1)) ` ? (+1 because of the function gene)

test/test_genome.py
jakobj: good point, you're correct, a `+1` is missing

test/test_genome.py
jakobj: i see. well, i think your test is anyway adapted to this specific configuration of the genome, isn't it? for example would it be robust against increasing the number of columns? if it's specific, you don't need the `if` since you define `n_inputs =1` before.

test/test_genome.py
jakobj: yep! you could also use the floor division operator (`//`) directly instead of casting: `gene_idx // (genome.primitives.max_arity + 1)`

test/test_genome.py
HenrikMettler: Changing `n_columns` wouldn't matter. Only in the first column the input gene of hidden can not be mutated (in the case of `n_input = 1`). Thats why n_rows matters. (Unless I am missing some other case for which only one 1 allele value is permitted).  The test would still pass with `n_input != 1` but it would loose some capacity.

test/test_genome.py
HenrikMettler: What should I use as values for `rel` in `pytest.approx()`? The assertion passes with `n=10000` and `rel=0.02` which seems reasonable to me. What do you think?

test/test_genome.py
HenrikMettler: It is tricky, since the indices selected for mutation `selected_gene_indices` are calculated locally within `genome.mutate`. Is there maybe a way to overwrite a local variable with a global one in pytest? (couldn't find anything with a quick search in the docs...)

HenrikMettler:
Just noticed that `test_create_new_offspring_and_parent_generation` in `test_ea_mu_plus_lambda.py` occasionally fails. I assume this is since non of the genes that have more than one permissible value is selected for mutation. Should I adapt the test logic or adapt `mutate` such that always at least one gene in mutated?

HenrikMettler:
> the fitness should be identical, independent of the number of processes as far as I remember. this is obviously desirable from a reproducibility viewpoint. is it possible that any of the stochastic operations are not using the internal rng of the population? that would be my first guess.

As far as I can see the problem just arises when setting `n_process = 1` and not for any other (valid) number of processes. Could the error have something to do with https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/ea/mu_plus_lambda.py#L157 ? I don't know how my changes influence this though

jakobj:
> Just noticed that `test_create_new_offspring_and_parent_generation` in `test_ea_mu_plus_lambda.py` occasionally fails. I assume this is since non of the genes that have more than one permissible value is selected for mutation. Should I adapt the test logic or adapt `mutate` such that always at least one gene in mutated?

i would adapt the test logic as we are happy with the logic in mutate. maybe you can increase the mutation rate to `1.0`?

jakobj:
> > the fitness should be identical, independent of the number of processes as far as I remember. this is obviously desirable from a reproducibility viewpoint. is it possible that any of the stochastic operations are not using the internal rng of the population? that would be my first guess.
> 
> As far as I can see the problem just arises when setting `n_process = 1` and not for any other (valid) number of processes. Could the error have something to do with https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/ea/mu_plus_lambda.py#L157 ? I don't know how my changes influence this though

what do you mean by "the problem"? is the fitness always the same except for running a single process?

cgp/genome.py
```diff
@@ -354,79 +357,82 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
         only_silent_mutations = True
-        while count_dna_differences(self.dna, dna) < n_mutations:
 
-            gene_idx = rng.randint(0, self._n_genes)
-            region_idx = gene_idx // self._length_per_region
+        selected_gene_indices = np.nonzero(np.random.rand(len(dna)) < mutation_rate)[0]
 
-            if self._is_input_region(region_idx):
-                continue  # nothing to do here
+        for (allele_idx, allele) in zip(
```
jakobj, line 375: sorry to be annoying, but i think `gene_idx` is fine, this refers to a particular position in the genome. really it should be called just `gene` but i think it's ok to add the suffix. `allele_idx` could be misinterpreted as an index in the list of possible settings for this particular gene.

test/test_genome.py
jakobj: i see, that is a bit tricky. two options come to mind: i) adding a function to `Genome` that returns the `selected_gene_indices` and monkey patching it in the test to return the gene indices that you'd like to use or ii) split `mutate` at the line where `selected_gene_indices` is set such that `Genome` has a mutate function that operates on a list of gene indices. i think i would go for i) for now and after this PR has been merged think about how we can refactor `mutate` to make it simpler to understand and test

cgp/genome.py
HenrikMettler: Ah yes makes sense, will change back to  `gene_idx`

HenrikMettler:
> > Just noticed that `test_create_new_offspring_and_parent_generation` in `test_ea_mu_plus_lambda.py` occasionally fails. I assume this is since non of the genes that have more than one permissible value is selected for mutation. Should I adapt the test logic or adapt `mutate` such that always at least one gene in mutated?
> 
> i would adapt the test logic as we are happy with the logic in mutate. maybe you can increase the mutation rate to `1.0`?

I can't set it to '1.0' because of https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/population.py#L37 (`if not (0.0 < mutation_rate and mutation_rate < 1.0):
            raise ValueError("mutation rate needs to be in (0, 1)")` But 0.99 should be fine, I guess? (The chance of not having 0 mutations with that mutation_rate can be neglectable)

HenrikMettler:
> > > the fitness should be identical, independent of the number of processes as far as I remember. this is obviously desirable from a reproducibility viewpoint. is it possible that any of the stochastic operations are not using the internal rng of the population? that would be my first guess.
> > 
> > 
> > As far as I can see the problem just arises when setting `n_process = 1` and not for any other (valid) number of processes. Could the error have something to do with https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/ea/mu_plus_lambda.py#L157 ? I don't know how my changes influence this though
> 
> what do you mean by "the problem"? is the fitness always the same except for running a single process?

Yes. (It also doesn't matter at which position in the vector `n_process = 1` is, eg if I set `for n_process in [2, 1, 4]` still the one where `n_process = 1` is having a different fitness)

jakobj:
> > i would adapt the test logic as we are happy with the logic in mutate. maybe you can increase the mutation rate to `1.0`?
> 
> I can't set it to '1.0' because of https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/population.py#L37 (`if not (0.0 < mutation_rate and mutation_rate < 1.0): raise ValueError("mutation rate needs to be in (0, 1)")` But 0.99 should be fine, I guess? (The chance of not having 0 mutations with that mutation_rate can be neglectable)

ah, interesting. well, why don't we allow a mutation rate of `1.0`? excluding the lower bound makes sense since no evolution takes place if no mutation can happen, but the upper bound could be included, right?

jakobj:
> > > > the fitness should be identical, independent of the number of processes as far as I remember. this is obviously desirable from a reproducibility viewpoint. is it possible that any of the stochastic operations are not using the internal rng of the population? that would be my first guess.
> > > 
> > > 
> > > As far as I can see the problem just arises when setting `n_process = 1` and not for any other (valid) number of processes. Could the error have something to do with https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/ea/mu_plus_lambda.py#L157 ? I don't know how my changes influence this though
> > 
> > 
> > what do you mean by "the problem"? is the fitness always the same except for running a single process?
> 
> Yes. (It also doesn't matter at which position in the vector `n_process = 1` is, eg if I set `for n_process in [2, 1, 4]` still the one where `n_process = 1` is having a different fitness)

this is not good and we should fix it before merging. i'll see whether i find time today to look into this.

cgp/genome.py
```diff
@@ -346,6 +349,10 @@ def _is_active_input_gene(self, gene_idx: int) -> bool:
         else:
             assert False  # should never be reached
 
+    def _select_gene_indices(self, mutation_rate, dna):
```
jakobj, line 352: could you please add annotations? and maybe rename the function to `_select_gene_indices_for_mutation`? and maybe you don't need to pass the dna, just the length of it since you never use it besides in `len(dna)`

cgp/genome.py
```diff
@@ -354,79 +361,80 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
-            region_idx = gene_idx // self._length_per_region
+        selected_gene_indices = self._select_gene_indices(mutation_rate, dna)
 
-            if self._is_input_region(region_idx):
-                continue  # nothing to do here
+        for (gene_idx, allele) in zip(selected_gene_indices, np.array(dna)[selected_gene_indices]):
 
-            elif self._is_output_region(region_idx):
-                silent = self._mutate_output_region(dna, gene_idx, rng)
+            silent = True
```
jakobj, line 381:
```suggestion
```
i think you can remove this if you add indentation below

cgp/genome.py
```diff
@@ -354,79 +361,80 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
+                gene_idx, allele, region_idx
+            )
+            if len(permissible_values) > 0:
 
-            else:
-                assert False  # should never be reached
+                dna[gene_idx] = rng.choice(permissible_values)
+                silent = region_idx not in active_regions
 
             only_silent_mutations = only_silent_mutations and silent
```
jakobj, line 392:
```suggestion
                only_silent_mutations = only_silent_mutations and silent
```
move this into the `if` block

cgp/genome.py
```diff
@@ -354,79 +361,80 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
-        assert self._is_gene_in_hidden_region(gene_idx)
+        elif self._is_output_region(region_idx):
+            return self._determine_permissible_values_output_gene(gene_idx, gene)
 
-        region_idx = gene_idx // self._length_per_region
-        silent_mutation = region_idx not in active_regions
+        else:
+            assert False  # should never be reached
 
+    def _determine_permissible_values_hidden_gene(
```
jakobj, line 413:
```suggestion
    def _determine_alternative_permissible_values_hidden_gene(
```

cgp/genome.py
```diff
@@ -354,79 +361,80 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
 
+        elif self._is_hidden_input_gene(gene_idx, region_idx):
+            permissible_values = self._permissible_inputs(region_idx)
+
+        else:
+            assert False
+        permissible_values.remove(gene)
+        return permissible_values
+
+    def _determine_permissible_values_output_gene(self, gene_idx: int, gene: int) -> List[int]:
```
jakobj, line 427:
```suggestion
    def _determine_alternative_permissible_values_output_gene(self, gene_idx: int, gene: int) -> List[int]:
```

test/conftest.py
```diff
@@ -37,7 +37,7 @@ def ea_params():
 
 @fixture
 def mutation_rate():
-    return 0.05
+    return 0.1
```
jakobj, line 40: why has this been changed?

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
+    genome.randomize(rng)
 
-    # mutating any gene in inactive region returns True
-    assert genome._mutate_hidden_region(list(dna), 3, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 4, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 5, active_regions, rng) is True
+    def count_n_immutable_genes(n_inputs, n_output, n_row):
+        n_immutable_genes = n_inputs * (
+            genome.primitives.max_arity + 1
+        )  # none of the input gene is mutable (+1 for function gene)
```
jakobj, line 362:
```suggestion
        )  # none of the input genes are mutable (+1 for function gene)
```

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
-    genome.dna = list(dna)
-    active_regions = cgp.CartesianGraph(genome).determine_active_regions()
+    genome.randomize(rng)
 
-    # mutating any gene in inactive region returns True
-    assert genome._mutate_hidden_region(list(dna), 3, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 4, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 5, active_regions, rng) is True
+    def count_n_immutable_genes(n_inputs, n_output, n_row):
+        n_immutable_genes = n_inputs * (
```
jakobj, line 360:
```suggestion
    def count_n_immutable_genes(n_inputs, n_output, n_row):
        length_per_region = genome.primitives.max_arity + 1
        n_immutable_genes = n_inputs * (
```
i would suggest to introduce this variable to make the code below more readable

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
+        )  # only one gene per output can be mutated
+        if n_inputs == 1:
+            n_immutable_genes += (
+                n_row * 2
+            )  # input gene addresses in the first hidden layer can't be mutated in that case
+        return n_immutable_genes
+
+    def count_mutations(prev_dna, new_dna):
+        counter = 0
+        for idx in range(len(prev_dna)):
```
jakobj, line 374:
```suggestion
        for (allele0, allele1) in zip(dna0, dna1):
```

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
+    n_mutations_std_expected = np.sqrt(
+        (len(genome.dna) - n_immutable_genes) * mutation_rate * (1 - mutation_rate)
+    )
 
-    # mutating function gene in active region returns False
-    assert genome._mutate_hidden_region(list(dna), 6, active_regions, rng) is False
-    # mutating inactive genes in active region returns True
-    assert genome._mutate_hidden_region(list(dna), 7, active_regions, rng) is True
-    assert genome._mutate_hidden_region(list(dna), 8, active_regions, rng) is True
+    n = 10000
```
jakobj, line 385: how long does this test take? we should try to avoid a test suite that's running too long -- people will get lazy about executing it :|

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
+    assert only_silent_mutations_0
+
+    only_silent_mutations_1 = genome.mutate(mutation_rate=1, rng=rng)
+    assert not only_silent_mutations_1
+
+    graph = CartesianGraph(genome)
+    active_regions = graph.determine_active_regions()
+
+    gene_to_be_mutated_active = active_regions[-1] * (
+        genome.primitives.max_arity + 1
```
jakobj, line 411: again, i would introduce `length_per_region` to avoid repeating `max_arity + 1`

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
+    )  # function gene of the 1st active hidden gene, should always be mutable
+    gene_to_be_mutated_non_active = 4 * (
+        genome.primitives.max_arity + 1
+    )  # 4 is a non active gene in this seed
```
jakobj, line 415: i would avoid making this depend on the seed, rather hardcode the dna above and avoid `randomize`

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
-        0,
-        0,
-        0,
-        0,
-        0,
-        ID_OUTPUT_NODE,
+    genome.randomize(rng)
+
+    only_silent_mutations_0 = genome.mutate(mutation_rate=0, rng=rng)
+    assert only_silent_mutations_0
```
jakobj, line 402:
```suggestion
    only_silent_mutations = genome.mutate(mutation_rate=0, rng=rng)
    assert only_silent_mutations is True
```
you don't need the suffix and i would add (redundant) `is True` to be explicit

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
+        "n_rows": 3,
+        "levels_back": 2,
+        "primitives": (cgp.Add, cgp.Sub, cgp.ConstantFloat),
+    }
+
+    genome = cgp.Genome(**genome_params)
+
+    # test function gene
+    gene_idx = 6  # function gene of first hidden region
+    gene = 0  #
```
jakobj, line 448:
```suggestion
    allele = 0  #
```

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
+    assert permissible_function_gene_values == [
+        1,
         2,
-        ID_NON_CODING_GENE,
-    ]
+    ]  # function idx 1,2 (cgp.Sub, cgp.ConstantFloat)
+
+    # test input gene
+    gene_idx = 16  # first input gene in second column of hidden region
+    gene = 0
```
jakobj, line 460:
```suggestion
    allele = 0
```

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
-    levels_back = 5
-    primitives = (cgp.Add, cgp.Sub, cgp.Mul, cgp.Div, cgp.ConstantFloat)
+    gene_idx_function = 33
+    gene_idx_input0 = 34
+    gene_idx_input1 = 35
 
-    rng = np.random.RandomState(rng_seed)
-    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives)
-    genome.randomize(rng)
+    gene = 5  # set current value for input0 -> should be excluded from permissible values
```
jakobj, line 489:
```suggestion
    allele = 5  # set current value for input0 -> should be excluded from permissible values
```

test/test_genome.py
```diff
@@ -346,89 +347,154 @@ def test_is_gene_in_output_region(rng_seed):
...
+
+    permissible_values_0 = genome._determine_permissible_values_output_gene(
```
jakobj, line 491: again, i think you can drop the suffix

test/test_genome.py
HenrikMettler: all tests in test_genome.py take just 2.3s on my laptop (all tests 18s). So I don't think it is an issue, but I can reduce it and increase the relative tolerance if you want.

test/conftest.py
HenrikMettler: My bad, I changed this while developing and commited accidentally. it is not necessary

test/test_genome.py
HenrikMettler: addressed in [https://github.com/Happy-Algorithms-League/hal-cgp/pull/180/commits/a59116918515c098c719caff22e53237abb7e7f7](url)

cgp/genome.py
```diff
@@ -354,79 +376,83 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
-                continue  # nothing to do here
+        for (gene_idx, allele) in zip(selected_gene_indices, np.array(dna)[selected_gene_indices]):
 
-            elif self._is_output_region(region_idx):
-                silent = self._mutate_output_region(dna, gene_idx, rng)
+            region_idx = gene_idx // self._length_per_region
 
-            elif self._is_hidden_region(region_idx):
-                silent = self._mutate_hidden_region(dna, gene_idx, active_regions, rng)
+            permissible_values = self._determine_alternative_permissible_values(
```
mschmidt87, line 398: One question: Can this check be done once at initialization time of the `Genome` such that we don't need to repeat this every time we want to mutate the genome? The permissible values only depend on the position of the gene in the dna, right? And since the genome parameters don't change over time, I am suspecting that it might be possible to do this once and be done with it. This might save some computation time.

jakobj: in principle yes, i think. one could build a dictionary, e.g., in `__init__`  that lists all permissible values for each gene idx. i agree, this might save quite some computation time at the expense of increased memory usage and slower instantiation of `Genome` objects. as always, the root of all evil is premature optimization, so i wouldn't implement this without profiling and finding that this is indeed a bottleneck. maybe a more lightweight compromise would be some kind of memoization, e.g., decorating the function determining all permissible values with the [lru cache decorator](https://docs.python.org/3/library/functools.html#functools.lru_cache). we should address this, and in general some lightweight performance improvements, in a separate issue/PRs.

cgp/genome.py
```diff
@@ -346,6 +349,25 @@ def _is_active_input_gene(self, gene_idx: int) -> bool:
         else:
             assert False  # should never be reached
 
+    def _select_gene_indices_for_mutation(self, mutation_rate, len_dna):
```
jakobj, line 352:
```suggestion
    def _select_gene_indices_for_mutation(self, mutation_rate, len_dna, rng):
```
this function should receive an `np.random.RandomState` instance that should be used instead of `np.random` below. i believe this should also fix the reproducibility issues with multiple processes.

cgp/genome.py
```diff
@@ -346,6 +349,25 @@ def _is_active_input_gene(self, gene_idx: int) -> bool:
...
+        len_dna : int
+            Length of the genome dna.
+
+        Returns
+        ----------
+        selected_gene_indices: numpy array
+            indices of the genes selected for mutation.
+        """
+
+        selected_gene_indices = np.nonzero(np.random.rand(len_dna) < mutation_rate)[0]
```
jakobj, line 368:
```suggestion
        selected_gene_indices = np.nonzero(rng.rand(len_dna) < mutation_rate)[0]
```

cgp/genome.py
```diff
@@ -354,79 +376,83 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
         active_regions = graph.determine_active_regions()
-
         dna = list(self._dna)
-
         only_silent_mutations = True
-        while count_dna_differences(self.dna, dna) < n_mutations:
 
-            gene_idx = rng.randint(0, self._n_genes)
-            region_idx = gene_idx // self._length_per_region
+        selected_gene_indices = self._select_gene_indices_for_mutation(mutation_rate, len(dna))
```
jakobj, line 392:
```suggestion
        selected_gene_indices = self._select_gene_indices_for_mutation(mutation_rate, len(dna), rng)
```

cgp/genome.py
mschmidt87: Thanks for the calendar motto ;-)
I think that the advantages of performing this calculation instead of many times outweigh the disadvantage of increased memory consumption (This will be a tiny data structure of a few bytes). But I am fine with postponing this to another PR. I created a ticket to remind us: #196

HenrikMettler: That's a good input, thanks! Yes I don't see why this shouldn't work. I agree that memory shouldn't be an issue, the number of values to store is `n < n_units * len(dna)`. I will take care of the #196 once I made this PR ready to merge

cgp/genome.py
```diff
@@ -346,6 +349,27 @@ def _is_active_input_gene(self, gene_idx: int) -> bool:
         else:
             assert False  # should never be reached
 
+    def _select_gene_indices_for_mutation(self, mutation_rate, len_dna, rng):
```
jakobj, line 352: please add type annotations

cgp/genome.py
```diff
@@ -346,6 +349,27 @@ def _is_active_input_gene(self, gene_idx: int) -> bool:
...
+        """Selects the gene indices for mutations
+
+        Parameters
+        ----------
+        mutation_rate : float
+            Proportion of genes to be mutated, between 0 and 1.
```
jakobj, line 358:
```suggestion
            Probability of a gene to be mutated, between 0 (excluded) and 1 (included).
```
btw, you should also check in other doc strings whether the description of `mutation_rate` needs to be updated.

cgp/genome.py
```diff
@@ -346,6 +349,27 @@ def _is_active_input_gene(self, gene_idx: int) -> bool:
...
+        len_dna : int
+            Length of the genome dna.
+        rng : numpy.random.RandomState
+            Random number generator instance to use for selecting the indices.
+
+        Returns
+        ----------
+        selected_gene_indices: numpy array
```
jakobj, line 366:
```suggestion
        selected_gene_indices: np.ndarray
```

cgp/genome.py
jakobj: do you actually need to be able to pass the length of the dna? you could just use `len(self.dna)` here. or do you need to be able to modify this for any of the tests?

cgp/genome.py
```diff
@@ -354,79 +378,85 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
 
         self.dna = dna
         return only_silent_mutations
 
-    def _mutate_output_region(
-        self, dna: List[int], gene_idx: int, rng: np.random.RandomState
-    ) -> bool:
-        assert self._is_gene_in_output_region(gene_idx)
+    def _determine_alternative_permissible_values(
+        self, gene_idx: int, gene: int, region_idx: int
```
jakobj, line 416:
```suggestion
        self, gene_idx: int, allele: int, region_idx: int
```

cgp/genome.py
```diff
@@ -354,79 +378,85 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
+        elif self._is_output_region(region_idx):
+            return self._determine_alternative_permissible_values_output_gene(gene_idx, gene)
 
-        region_idx = gene_idx // self._length_per_region
-        silent_mutation = region_idx not in active_regions
+        else:
+            assert False  # should never be reached
 
+    def _determine_alternative_permissible_values_hidden_gene(
+        self, gene_idx: int, gene: int, region_idx: int
```
jakobj, line 434:
```suggestion
        self, gene_idx: int, allele: int, region_idx: int
```

cgp/genome.py
```diff
@@ -354,79 +378,85 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
...
+        elif self._is_hidden_input_gene(gene_idx, region_idx):
+            permissible_values = self._permissible_inputs(region_idx)
+
+        else:
+            assert False
+        permissible_values.remove(gene)
+        return permissible_values
+
+    def _determine_alternative_permissible_values_output_gene(
+        self, gene_idx: int, gene: int
```
jakobj, line 448:
```suggestion
        self, gene_idx: int, allele: int
```

cgp/population.py
```diff
@@ -34,7 +34,7 @@ def __init__(
         """
         self.n_parents = n_parents  # number of individuals in parent population
 
-        if not (0.0 < mutation_rate and mutation_rate < 1.0):
+        if not (0.0 < mutation_rate and mutation_rate <= 1.0):
             raise ValueError("mutation rate needs to be in (0, 1)")
```
jakobj, line 38:
```suggestion
            raise ValueError("mutation rate needs to be in (0, 1]")
```

test/test_genome.py
```diff
@@ -346,89 +347,203 @@ def test_is_gene_in_output_region(rng_seed):
...
+    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, None, (cgp.Add, cgp.Sub))
     rng = np.random.RandomState(rng_seed)
-    genome = cgp.Genome(1, 1, 3, 1, None, (cgp.Add, cgp.ConstantFloat))
-    dna = [
-        ID_INPUT_NODE,
-        ID_NON_CODING_GENE,
-        ID_NON_CODING_GENE,
+    genome.randomize(rng)
+
+    def count_n_immutable_genes(n_inputs, n_output, n_row):
```
jakobj, line 359:
```suggestion
    def count_n_immutable_genes(n_inputs, n_outputs, n_rows):
```

test/test_genome.py
```diff
@@ -346,89 +347,203 @@ def test_is_gene_in_output_region(rng_seed):
...
+        length_per_region = genome.primitives.max_arity + 1  # function gene + input gene addresses
+        n_immutable_genes = n_inputs * length_per_region  # none of the input genes are mutable
+        n_immutable_genes += n_output * (
+            length_per_region - 1
+        )  # only one gene per output can be mutated
+        if n_inputs == 1:
+            n_immutable_genes += (
+                n_row * 2
+            )  # input gene addresses in the first hidden layer can't be mutated in that case
```
jakobj, line 368:
```suggestion
            )  # input gene addresses in the first (hidden) column can't be mutated if only one input node exists
```

test/test_genome.py
```diff
@@ -346,89 +347,203 @@ def test_is_gene_in_output_region(rng_seed):
...
+    genome.randomize(rng)
+
+    only_silent_mutations = genome.mutate(mutation_rate=0, rng=rng)
+    assert only_silent_mutations is True
+
+    only_silent_mutations = genome.mutate(mutation_rate=1, rng=rng)
+    assert not only_silent_mutations
+
+    dna_fixed = [
+        -1,
```
jakobj, line 407: instead of `-1` etc you should use the constants `ID_INPUT_NODE` etc defined in `genome.py`

test/test_genome.py
```diff
@@ -346,89 +347,203 @@ def test_is_gene_in_output_region(rng_seed):
...
     ]
+    genome.dna = dna_fixed
+    graph = CartesianGraph(genome)
+    active_regions = graph.determine_active_regions()
+    length_per_region = genome.primitives.max_arity + 1  # function gene + input gene addresses
+    gene_to_be_mutated_non_active = (
+        3 * length_per_region
+    )  # operator gene of 2nd internal unit, a silent unit in this dna
+
+    def select_gene_indices_silent(mutation_rate, dna, rng):
```
jakobj, line 452:
```suggestion
    def select_gene_indices_silent(mutation_rate, len(dna), rng):
```
only if you keep the length as an argument

test/test_hl_api.py
```diff
@@ -65,6 +65,8 @@ def test_parallel_population(population_params, genome_params, ea_params):
...
+    assert fitness_per_n_processes[1] == pytest.approx(fitness_per_n_processes[2])
```
jakobj, line 69: please move this below the other asserts. first 0 & 1, then 0 & 2, then 1 & 2 should be checked.

test/test_genome.py
```diff
@@ -346,89 +347,203 @@ def test_is_gene_in_output_region(rng_seed):
...
-        ID_NON_CODING_GENE,
+        -3,
     ]
+    genome.dna = dna_fixed
+    graph = CartesianGraph(genome)
+    active_regions = graph.determine_active_regions()
+    length_per_region = genome.primitives.max_arity + 1  # function gene + input gene addresses
+    gene_to_be_mutated_non_active = (
+        3 * length_per_region
+    )  # operator gene of 2nd internal unit, a silent unit in this dna
```
jakobj, line 450:
```suggestion
    )  # operator gene of 2nd hidden node, a silent node in this dna configuration
```
while it might be a bit confusing to use "hidden" for internal nodes, this is currently the convention we are using. i would suggest to stick to this for now. if we decide to change this, we should change it throughout the whole code base

test/test_hl_api.py
HenrikMettler: Can I also re-remove it? I only used it for finding the error in the test

jakobj: sure, you can also remove it. if 0 == 1 and 0 == 2, this implies 1 == 2 ;)

test/test_genome.py
```diff
@@ -346,53 +347,91 @@ def test_is_gene_in_output_region(rng_seed):
...
+
+    def count_n_immutable_genes(n_inputs, n_outputs, n_rows):
+        length_per_region = genome.primitives.max_arity + 1  # function gene + input gene addresses
+        n_immutable_genes = n_inputs * length_per_region  # none of the input genes are mutable
+        n_immutable_genes += n_outputs * (
+            length_per_region - 1
+        )  # only one gene per output can be mutated
+        if n_inputs == 1:
+            n_immutable_genes += (
+                n_rows * 2
```
jakobj, line 367:
```suggestion
                n_rows * genome.primitives.max_arity
```

test/test_genome.py
```diff
@@ -403,32 +442,110 @@ def test_mutate_output_region(rng_seed):
...
+    gene_to_be_mutated_non_active = (
+        3 * length_per_region
+    )  # operator gene of 2nd hidden node, a silent node in this dna configuration
+
+    def select_gene_indices_silent(mutation_rate, rng):
+        selected_gene_indices = [gene_to_be_mutated_non_active]
+        return selected_gene_indices
+
+    genome._select_gene_indices_for_mutation = \
+        select_gene_indices_silent  # monkey patch the selection of indices to selecting a silent gene
```
jakobj, line 458:
```suggestion
        select_gene_indices_silent  # monkey patch the selection of indices to select a silent gene
```

test/test_genome.py
```diff
@@ -403,32 +442,110 @@ def test_mutate_output_region(rng_seed):
...
+    gene_to_be_mutated_active = (
+        active_regions[-1] * length_per_region
+    )  # operator gene of the 1st active hidden node,
+    # should always be mutable and result in a non-silent-mutation
+
+    def select_gene_indices_non_silent(mutation_rate, rng):
+        selected_gene_indices = [gene_to_be_mutated_active]
+        return selected_gene_indices
+
+    genome._select_gene_indices_for_mutation = select_gene_indices_non_silent
```
jakobj, line 472:
```suggestion
    genome._select_gene_indices_for_mutation = select_gene_indices_non_silent  # monkey patch the selection of indices to select a non-silent gene
```

cgp/genome.py
```diff
@@ -346,6 +349,27 @@ def _is_active_input_gene(self, gene_idx: int) -> bool:
...
 
+    def _select_gene_indices_for_mutation(
+        self, mutation_rate: float, rng: np.random.RandomState
+    ) -> List[int]:
+        """Selects the gene indices for mutations
+
+        Parameters
+        ----------
+        mutation_rate : float
+            Proportion of genes to be mutated, between 0 (excluded) and 1 (included)
```
jakobj, line 360:
```suggestion
            Probability of a gene to be mutated, between 0 (excluded) and 1 (included)
```
please also fix this in the docstring of  `Population.__init__`

HenrikMettler:
I am not sure how to fix to travis CI build issue, when I introduce a line break, black "refixes" it

cgp/genome.py
jakobj: ?

cgp/genome.py
```diff
@@ -346,13 +349,34 @@ def _is_active_input_gene(self, gene_idx: int) -> bool:
...
+        return selected_gene_indices
+
     def mutate(self, mutation_rate: float, rng: np.random.RandomState):
         """Mutate the genome.
 
         Parameters
         ----------
         mutation_rate : float
-            Proportion of genes to be mutated, between 0 and 1.
+            Proportion of a gene to be mutated, between 0 (excluded) and 1 (included).
```
jakobj, line 379:
```suggestion
            Probability of a gene to be mutated, between 0 (excluded) and 1 (included)
```

test/test_genome.py
```diff
@@ -403,32 +442,110 @@ def test_mutate_output_region(rng_seed):
...
+        "primitives": (cgp.Add, cgp.Sub, cgp.ConstantFloat),
+    }
+
+    genome = cgp.Genome(**genome_params)
+
+    # test function gene
+    gene_idx = 6  # function gene of first hidden region
+    allele = 0  #
+    region_idx = None  # can be any number, since not touched for function gene
+    permissible_function_gene_values = genome._determine_alternative_permissible_values_hidden_gene(
```
jakobj, line 495:
```suggestion
    permissible_values = genome._determine_alternative_permissible_values_hidden_gene(
```

test/test_genome.py
```diff
@@ -403,32 +442,110 @@ def test_mutate_output_region(rng_seed):
...
+        2,
+    ]  # function idx 1,2 (cgp.Sub, cgp.ConstantFloat)
+
+    # test input gene
+    gene_idx = 16  # first input gene in second column of hidden region
+    allele = 0
+    region_idx = gene_idx // (
+        genome.primitives.max_arity + 1
+    )  # function gene + input gene addresses
+    permissible_hidden_input_gene_values = genome._determine_alternative_permissible_values_hidden_gene(
```
jakobj, line 509:
```suggestion
    permissible_values = genome._determine_alternative_permissible_values_hidden_gene(
```

jakobj:
> I am not sure how to fix to travis CI build issue, when I introduce a line break, black "refixes" it

hmm, yeah, black and flake seem to bite each other here. i suggested a shorter variable name, that should do the trick :)

PR #181, opened 2020-07-12
Doc/sphinx docs

mschmidt87:
This PR implements proper documentation using sphinx. It does the following things:
- Add `docs/` folder with sphinx infrastructure
- Reformat README file as restructured text and add markers to separate individual sections. We need to use rst for the README to enable `sphinx` to read single section from the README
- Add placeholders for single parts of the documentation (e.g. `installation.rst`) that read the content of the corresponding section from the README (indicated by the markers in the README)
- Add autodoc to documentation to document the API which reads automatically from the docstrings in the source code.
- Remove `long_description.md` and make `setup.py` read the long description directly from README, again using the marker.
- Add building of documentation to continuous integration

The idea is that we should not duplicate any documentation, i.e., anything that lives in README should be just read from `sphinx` to display it in the documentation. The documentation can then (in later PRs) add more information to the actual documentation which will not live in the README.

Outside of the scope of this PR:
- proper documentation of the examples in the documentation 
- decision where to host the documentation (candidates would be github pages or readthedoc, e.g.) 

To test, run:
```shell
pip install sphinx recommonmark msmb_theme sphinx-rdt-theme
cd docs/
make html
firefox _builx/html/index.html
```


.travis.yml
```diff
@@ -32,6 +32,9 @@ script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
+  - if [ "$DEP" = "[all]" ]; then
+       cd docs && make html && cd ../ || exit 1;
```
jakobj, line 36:
```suggestion
       cd docs && make html || exit 1;
```
you don't need to go back, do you?

README.rst
```diff
@@ -1,109 +1,138 @@
...
-[![Coverage Status](https://coveralls.io/repos/github/Happy-Algorithms-League/hal-cgp/badge.svg?branch=master)](https://coveralls.io/github/Happy-Algorithms-League/hal-cgp?branch=master)
-[![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.3889163.svg)](https://doi.org/10.5281/zenodo.3889163)
+========
+ HAL-CGP
+========
 
 
+.. image:: https://badge.fury.io/py/hal-cgp.svg
+    :target: https://badge.fury.io/py/hal-cgp
+
```
jakobj, line 8:
```suggestion
```

README.rst
```diff
@@ -1,109 +1,138 @@
...
-```
 
+   .. code-block:: shell
+
+      git clone git@github.com:Happy-Algorithms-League/hal-cgp.git
+      cd hal-cgp
+      pip install .[all]
 
+.. installation-end
+.. basic-usage-start
```
jakobj, line 75:
```suggestion

.. basic-usage-start
```

README.rst
```diff
@@ -1,109 +1,138 @@
...
-```python
-def objective(individual):
-    individual.fitness = ...
-    return individual
-```
+
+   .. code-block:: python
+		
+      def objective(individual):
+          individual.fitness = ...
```
jakobj, line 88: this looks like too much indentation

README.rst
```diff
@@ -1,109 +1,138 @@
...
-
-evolve_params = {"max_generations": 1000, "min_fitness": 0.0}
-```
+   
+   .. code-block:: python
+
+      population_params = {"n_parents": 10, "mutation_rate": 0.5, "seed": 8188211}
+
+      genome_params = {
+           "n_inputs": 2,
```
jakobj, line 99: this look like too much indentation

docs/api_reference/cgp.rst
```diff
@@ -0,0 +1,39 @@
...
+.. autoclass:: cgp.individual.IndividualMultiGenome
+   :members: 
+
+===============
+Cartesian Graph
+===============
+
+.. autoclass:: CartesianGraph
+   :members: 
+
```
jakobj, line 31:
```suggestion
```

docs/api_reference/ea.rst
```diff
@@ -0,0 +1,4 @@
+Evolutionary Algorithms
+=========================
```
jakobj, line 2:
```suggestion
Evolutionary Algorithms
=======================
```

docs/api_reference/local_search.rst
```diff
@@ -0,0 +1,6 @@
+============
+Local search
+============
+
```
jakobj, line 4:
```suggestion
```

docs/api_reference/nodes.rst
```diff
@@ -0,0 +1,7 @@
+Functional nodes
```
jakobj, line 1:
```suggestion
Operator nodes
```
or maybe just "Nodes"?

docs/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
jakobj, line 1: is this file needed somewhere?

extra-requirements.txt
```diff
@@ -6,3 +6,7 @@ scipy ~= 1.4.1
 sympy ~= 1.5.1
 torch ~= 1.5.0
 mypy ~= 0.770
+sphinx ~=3.1.2
```
jakobj, line 9: i'm not sure, but maybe we want to create yet another requirements file that is only necessary for building the docs? what if i'd like to have all extra requirements (sympy, torch, etc) but don't want to build the docs?

README.rst
```diff
@@ -1,109 +1,138 @@
...
+Figure from Jordan, Schmidt, Senn & Petrovici, "Evolving to learn: discovering interpretable plasticity rules for spiking networks", arxiv:2005.14149_.
 
-<sub>Figure from Jordan, Schmidt, Senn & Petrovici, "Evolving to learn: discovering interpretable plasticity rules for spiking networks", [ arxiv:2005.14149](https://arxiv.org/abs/2005.14149).</sub>
+.. _arxiv:2005.14149: https://arxiv.org/abs/2005.14149
 
+.. long-description-end
 
 A simple example of CGP applied to a symbolic regression problem can be found in `examples/example_evo_regression.py`.
 
-Searching numerical values for constants via local search
```
jakobj, line 26: somehow this bit disappeared. shouldn't this be included somewhere or do you plan to readd again in a separate PR?

.travis.yml
mschmidt87: not sure, don't we need to be in the parent directory for the following steps?

README.rst
mschmidt87: Yes, this will be added in a separate PR. I don't think that this should be in the README, at least not in this detailed version. Will add it into the docs in a later PR.

README.rst
mschmidt87: I think this looks fine in the rendered README because it aligns with the enumerated list items, please check here: https://github.com/Happy-Algorithms-League/hal-cgp/tree/doc/sphinx_docs

same for all the other comments below.

docs/api_reference/nodes.rst
mschmidt87: I prefer a more descriptive name.

docs/make.bat
mschmidt87: I though it would be necessary but I think it is actually not. I removed it.

extra-requirements.txt
mschmidt87: yes, good point. Implemented that.

mschmidt87: e91c89152ecadf38e62f2bd16c437cf4b9cace0a

mschmidt87: Also refined the instructions for installation: 96f232e9c727be072bd43fde5fb8793cef9b49da

mschmidt87: and here: 2b53c21a4afd66e6e6cce6b1c8593e9a7db692b7

.travis.yml
jakobj: i think after the command you're anyway in the directory where you started from

.travis.yml
```diff
@@ -32,6 +32,9 @@ script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
+  - if [ "$DEP" = "[all]" ]; then
```
jakobj:
```suggestion
  - if [ "$DEP" = "[doc]" ]; then
```
?

.travis.yml
```diff
@@ -4,7 +4,7 @@ python:
   - "3.7"
   - "3.8"
 env:
-  - DEP=[all]
+  - DEP=[all, doc, dev]
```
jakobj: does this start three jobs? or just one with everything? shouldn't this just be "all" then?

.travis.yml
```diff
@@ -32,8 +32,11 @@ script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
+  - if [ "$DEP" = "[all,doc,dev]" ]; then
+       cd docs && make html && cd ../ || exit 1;
+    fi
   - pytest --cov=cgp
-  - if [ "$DEP" = "[all]" -a $TRAVIS_PYTHON_VERSION = 3.8 ]; then
+  - if [ "$DEP" = "[all,doc,dev]" -a $TRAVIS_PYTHON_VERSION = 3.8 ]; then
```
jakobj, line 39: why do the examples depend on "doc"?

.travis.yml
```diff
@@ -32,8 +32,11 @@ script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
+  - if [ "$DEP" = "[all,doc,dev]" ]; then
```
jakobj, line 35:
```suggestion
  - if [ "$DEP" = "[all,doc]" ]; then
```
?

.travis.yml
```diff
@@ -4,7 +4,7 @@ python:
   - "3.7"
   - "3.8"
 env:
-  - DEP=[all]
+  - DEP=[all,doc,dev]
```
jakobj, line 7: does this start three jobs? if not, shouldn't "all" be sufficient?

docs/api_reference/nodes.rst
jakobj: i can live with that ;)

setup.py
```diff
@@ -24,19 +25,30 @@ def read_requirements():
...
-    extra_requirements = {}
-    extra_requirements["all"] = []
+    extra_requirements = defaultdict(lambda: [])
+    collect_key = "all"
     with open("./extra-requirements.txt") as f:
         for req in f:
             req = req.replace("\n", " ")
-            extra_requirements[_cut_version_number_from_requirement(req)] = [req]
-            extra_requirements["all"].append(req)
+            if req.startswith("#"):
```
jakobj, line 33: i think this logic here is a bit misleading. why not really put all reqs in the key "all", and only consider the other two ("dev", "doc") separately. alternatively, maybe use "extra" for all except "dev" and "doc"? using "all" but excluding "dev", "doc" confused me

mschmidt87: `all` here means "all requirements necessary for using the library as a user". But okay, we can use `extra` for that purpose and then `all` for everything.

All your other comments are already outdated because I fixed the logic in the travis file.

mschmidt87: Done in 93f2adcf19e319ce8ee98798e83b296116e4878d

.travis.yml
mschmidt87: now it is, because I changed the definition.

.travis.yml
mschmidt87: I think, we need the `cd ../` here. After removing it, the tests were not found, I guess, because we're not in the parent directory.

mschmidt87:
FYI, I think the `make.bat` is required, I've seen it checked in in other repos (legit repos like scikit-learn) and CI failed without it.

setup.py
```diff
@@ -24,19 +25,34 @@ def read_requirements():
...
         for req in f:
             req = req.replace("\n", " ")
-            extra_requirements[_cut_version_number_from_requirement(req)] = [req]
-            extra_requirements["all"].append(req)
-
+            if req.startswith("#"):  # new block in requirements file
+                collect_key = req.split(" ")[1]
+                collect_keys.append(collect_key)
+                continue
+            # We only list dependencies required for using the library as individual options
```
jakobj, line 38:
```suggestion
            # We only list extra dependencies required for using the library as individual options, not dev or doc dependencies
```

PR #189, opened 2020-07-15
Add deployment of documentation on github pages to CI

### LINKED ISSUE #184: Decide and implement hosting option for documentation
Once #181 is merged and we have proper documentation, we want to host the documentation on a website. We could go for two options:

1. Read the docs
https://readthedocs.org/
- external service that we can easily hook to our repository
- we would end up with a URL like https://hal-cgp.readthedocs.org 

2. Github Pages
Works similar to read the docs but requires a bit more work to set it up. On the plus side, we would have a nicer URL like `hal-cgp.github.io`. Also nicely integrates with sphinx and supports versioned documentation (just like readthedocs, this a concern of mine, but it seems possible without much difficulties: https://sphinxcontrib-versioning.readthedocs.io/en/latest/github_pages.html).

My preference would be Option 2: GitHub pages. What do you think, @jakobj ?

mschmidt87:
This fixes #184 .

First we build the documentation in one of the jobs (I simplified the command here) and place a `.nojekyll` file in the html folder (which tells github pages that we're not using jekyll as a static site builder).
In addition, the PR adds a deploy stage to travis CI, which will push the contents of `doc/_build/html` .

I followed the steps in this tutorial: https://medium.com/@FelipeFaria/sphinx-travis-ci-github-pages-full-integration-for-auto-docs-publishing-91ce93179e17

Update:
I also added a bit of simplification to the travis file:
- I refactored the documentation building line
- Instead of installing not extra requirements, we're installing the `[dev]` requirements in the last job so that we can skip the manual install of `black`, `mypy`, and `flake8`.

I also added references to the documentation to the README.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,16 @@ python:
   - "3.8"
 env:
   - DEP=[all]
-  - DEP=
+  - DEP=[dev]
 branches:
   only:
     - master
 jobs:
   exclude:
   - python: "3.6"
-    env: DEP=
+    env: DEP=[dev]
   - python: "3.7"
-    env: DEP=
+    env: DEP=[dev]
 before_install:
   - wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
   - chmod +x miniconda.sh
@@ -24,16 +24,16 @@ before_install:
   - source activate myenv
   - pip install --upgrade pip
 install:
-  - pip install pytest pytest-cov coveralls
-  - pip install flake8 black
-  - pip install mypy
   - pip install -e .$DEP
+  - pip install pytest-cov coveralls
 script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
   - if [ "$DEP" = "[all]" ]; then
-       cd docs && make html && cd ../ || exit 1;
+       make -C docs/ html || exit 1;
+       touch docs/_build/html/.nojekyll
+       
     fi
   - pytest --cov=cgp
   - if [ "$DEP" = "[all]" -a $TRAVIS_PYTHON_VERSION = 3.8 ]; then
@@ -43,3 +43,14 @@ script:
     fi
 after_success:
   - coveralls
+deploy:
+  provider: pages:git
+  skip_cleanup: true
+  verbose: true
+  token: $GITHUB_TOKEN
+  edge: true
+  local_dir: ./docs/_build/html/
+  keep_history: true
+  on:
+    python: 3.8
+    condition: $DEP = "[all]"
--- a/README.rst
+++ b/README.rst
@@ -21,6 +21,8 @@
 	   :target: https://github.com/psf/black
 .. image:: https://coveralls.io/repos/github/Happy-Algorithms-League/python-gp/badge.svg?branch=master
 	   :target: https://coveralls.io/github/Happy-Algorithms-League/python-gp?branch=master
+.. image:: https://readthedocs.org/projects/ansicolortags/badge/?version=latest
+	   :target: https://happy-algorithms-league.github.io/hal-cgp/
 
 Cartesian genetic programming (CGP) in pure Python.
 
@@ -35,8 +37,6 @@ Figure from Jordan, Schmidt, Senn & Petrovici, "Evolving to learn: discovering i
 .. _arxiv:2005.14149: https://arxiv.org/abs/2005.14149
 
 .. long-description-end
-
-A simple example of CGP applied to a symbolic regression problem can be found in `examples/example_evo_regression.py`.
 
 .. installation-start
 ============
@@ -72,10 +72,13 @@ The adventurous can install the most recent development version directly from ou
 
 .. installation-end
 
-.. basic-usage-start
 ===========
 Basic usage
 ===========
+
+For detailed documentation, please refer to `https://happy-algorithms-league.github.io/hal-cgp/ <https://happy-algorithms-league.github.io/hal-cgp/>`_. Here we only provide a preview.
+
+.. basic-usage-start
 
 Follow these steps to solve a basic regression problem:
 
--- a/docs/basic_usage.rst
+++ b/docs/basic_usage.rst
@@ -1,0 +1,7 @@
+===========
+Basic usage
+===========
+
+.. include:: ../README.rst
+   :start-after: basic-usage-start
+   :end-before: basic-usage-end


.travis.yml
```diff
@@ -43,3 +45,13 @@ script:
...
+deploy:
+  - provider: pages:git
+    skip_cleanup: true
+    verbose: true
+    token: $GITHUB_TOKEN
+    edge: true
+    local_dir: ./docs/_build/html/
+    keep_history: true
+    on:
+	condition:  "$DEP" = "[all]" -a $TRAVIS_PYTHON_VERSION = 3.8
```
jakobj, line 57: github highlights indentation and complains about missing end of line and end of file, maybe a copy/paste issue?

.travis.yml
```diff
@@ -33,7 +33,9 @@ script:
   - flake8 --config=.flake8 .
   - mypy cgp
   - if [ "$DEP" = "[all]" ]; then
-       cd docs && make html && cd ../ || exit 1;
+       make -C docs/ html || exit 1;
+       touch docs/build/html/.nojekyll
```
jakobj, line 37: is the difference between `build` and `_build` (used below) on purpose or a bug?

mschmidt87: oops, no, this is wrong.

mschmidt87:
> can I somehow check the github pages generated by this? is the doc deployed every time travis runs? how does this work? does every new PR update the docs? is that what we want? sorry, i am not familiar with this infrastructure

Yes, I was wondering, too, how we can actually check this PR, I honestly don't know.

In general, the deploy phase is skipped for PR builds, i.e., this will only be executed once a PR is merged into master and master is rebuilt in Travis. That means, only then the documentation is rebuild and deployed to github pages. I think, this is what we want, right? 

Regarding versioned documentation, i.e., having a documentation for release version alongside with the latest documentation: I am not sure how to do that at the moment. Will have to investigate, but I am pretty sure it is possible.

.travis.yml
mschmidt87: fixed, b5462067211cef93dcd68d602b6514aac917bb29

.travis.yml
mschmidt87: fixed. b5462067211cef93dcd68d602b6514aac917bb29

jakobj:
> In general, the deploy phase is skipped for PR builds, i.e., this will only be executed once a PR is merged into master and master is rebuilt in Travis. That means, only then the documentation is rebuild and deployed to github pages. I think, this is what we want, right?

yes, that sounds like what we want! :ok_hand: can you somehow force deploy for this specific PR or do we need to work on master? :grimacing: 
 
> Regarding versioned documentation, i.e., having a documentation for release version alongside with the latest documentation: I am not sure how to do that at the moment. Will have to investigate, but I am pretty sure it is possible.

this would indeed be great! and if we have to trigger it manually once per release, that would also be fine.

mschmidt87:
I don't think it's possible to trigger the deploy manually here, because PR builds just skip this phase altogether and there seems to be no option to circumvent this. (See here for documentation about deploy phases: https://docs.travis-ci.com/user/deployment#pull-requests )

The CI here also doesn't run, I am not sure what the reason for this is. I hesitate to just merge the PR like this, basically untested, but I don't know any other way. This seems a bit absurd.

mschmidt87:
Maybe we can merge this first to a different branch (e.g. `test_deploy`) and reconfigure travis before so that travis is also run on `test_deploy`. Then, after merging, we can see if everything works.

mschmidt87:
I found a different way to test this:
- I added this branch here to the branches that travis runs on each commit to trigger the deployment.
- This works fine and you can already see the result here: https://happy-algorithms-league.github.io/hal-cgp/ . This documentation publishes the documentation pushed to the `gh-pages` branch.

However, one problem remains: Currently, the deplog phase is run for each of the 4 jobs, which is a) unnecessary because we only need to run it once and b) fails for the job that does not install the doc requirements.
To change that, I think that we need to refactor the whole travis file a bit, but I am still figuring out how.

mschmidt87:
Update: I think I figured out a way to achieve everything, but now the deploy phase is skipped for *all* jobs, even for the one that should execute it (py=3.8, DEP=[all]). I don't understand why and have created a thread in the Travis CI community forum to ask for help: https://travis-ci.community/t/conditional-deploy-in-build-matrix-does-not-work/9362?u=mschmidt87

mschmidt87:
Update: I received a reply and could solve the issue (some simple syntax error).
Now everything is working. I removed this feature branch from the branches in travis CI again, so that the deployment is now only carried out on each commit to `master`.

For evidence that the pipeline is working, see the build job on this feature branch here: https://travis-ci.org/github/Happy-Algorithms-League/hal-cgp/builds/709031726
when you click on the 4 individual builds, you will see that the deployment is carried out on only one branch, namely the `python 3.8, [all]` build as intended. The documentation gets push to the `gh-pages` branch and published here: https://happy-algorithms-league.github.io/hal-cgp/

README.rst
```diff
@@ -21,6 +21,8 @@
 	   :target: https://github.com/psf/black
 .. image:: https://coveralls.io/repos/github/Happy-Algorithms-League/python-gp/badge.svg?branch=master
 	   :target: https://coveralls.io/github/Happy-Algorithms-League/python-gp?branch=master
+.. image:: https://readthedocs.org/projects/ansicolortags/badge/?version=latest
```
jakobj, line 24: does this badge break if the docs are not built correctly?

README.rst
```diff
@@ -77,6 +77,8 @@ The adventurous can install the most recent development version directly from ou
 Basic usage
 ===========
 
+For the detailed documentation, please refer to [https://happy-algorithms-league.github.io/hal-cgp/](https://happy-algorithms-league.github.io/hal-cgp/).
```
jakobj, line 80:
```suggestion
For detailed documentation, please refer to [https://happy-algorithms-league.github.io/hal-cgp/](https://happy-algorithms-league.github.io/hal-cgp/). Here we only provide a preview.
```
or something like that to keep the reader reading ;)

mschmidt87: Thanks. Done.

mschmidt87:
I incorporated your suggestion, @jakobj , and also added the missing file about basic usage to the docs. Seems I forgot to add it in #181

README.rst
mschmidt87: No, this badge is "stolen" from readthedocs, so it will not change, it's a static picture. However, we're testing the documentation build in CI, so we'll (almost certainly) never have breaking documentation on master.

jakobj: excellent, thanks!

PR #190, opened 2020-07-15
Add test for randomize_genome function of individual classes

### LINKED ISSUE #186: Add test for `randomize_genome` of individual classes
This function is currently untested, we should add a test for it.

mschmidt87:
Adds a simple test for the `randomize_genome` function of the individual classes. All it tests is that the `randomize_genome` function does indeed change the dna.

Closes #186

### DIFF
--- a/test/test_individual.py
+++ b/test/test_individual.py
@@ -294,3 +294,17 @@ def test_update_parameters_from_torch_class_does_not_reset_fitness_for_unused_pa
     g = _unpack_evaluation(individual.to_func(), individual_type)
     x = 2.0
     assert g([x])[0] == pytest.approx(x ** 2)
+
+
+@pytest.mark.parametrize("individual_type", ["SingleGenome", "MultiGenome"])
+def test_individual_randomize_genome(individual_type, rng_seed):
+    rng = np.random.RandomState(rng_seed)
+    primitives = (cgp.Add, cgp.Mul)
+    genome = cgp.Genome(1, 1, 2, 1, 1, primitives)
+    genome.randomize(rng)
+
+    dna_old = list(genome.dna)
+    individual = _create_individual(genome, individual_type=individual_type)
+
+    individual.randomize_genome(rng)
+    assert dna_old != _unpack_genome(individual, individual_type)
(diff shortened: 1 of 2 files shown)


test/test_individual.py
```diff
@@ -294,3 +295,17 @@ def test_update_parameters_from_torch_class_does_not_reset_fitness_for_unused_pa
...
+
+
+@pytest.mark.parametrize("individual_type", ["SingleGenome", "MultiGenome"])
+def test_individual_randomize_genome(individual_type, rng_seed):
+    rng = np.random.RandomState(rng_seed)
+    primitives = (cgp.Add, cgp.Mul)
+    genome = cgp.Genome(1, 1, 2, 1, 1, primitives)
+    genome.randomize(rng)
+
+    dna_old = copy.copy(genome.dna)
```
jakobj, line 307: i think `list(genome.dna)` would also work and avoid the import.

jakobj:
oh and i'm confused how this "closes" itself? ;)

mschmidt87:
> oh and i'm confused: how does this "close" itself? ;)

??

mschmidt87:
>  i am wondering however, whether this would not better fit into `test_genome` the individuals are just forwarding everything to the genome aren't they? furthermore, from a quick grep it seems like `randomize_genome` is never used?! is this possible? and if so, shouldn't we rather remove it to keep confusion levels low?

My main motivation here was to not leave the code untested, but yeah, this function is pretty redundant and if it not used, we can also remove it altogether, I agree. If that is our decision, we can close this PR and I'll open another one tomorrow, removing this functionality from the classes.

mschmidt87:
On 2nd thought, I think that we should instead of removing the function, rather use it in the `Population` in the `generate_random_individual`. Currently, we are doing a rather convoluted type check which is necessary because we call the `genome.randomize()` method. If instead, we call `individual.randomize_genome()`, this should become much easier.

mschmidt87:
I implemented the change in `Population`

mschmidt87:
Please take another look @jakobj

test/test_individual.py
jakobj: (bump)

mschmidt87: Ah thanks, I had forgotten about that. Fixed now.

mschmidt87:
Yep, thanks, I had forgotten about that, fixed it now and already squashed commits to 2 commits. @jakobj

PR #191, opened 2020-07-16
Allow multiple parameters for nodes and add example with parametrized additive node

### LINKED ISSUE #98: Introduce multiplicative and additive parameters for each node

jakobj:
With #178 this is now incredibly easy: one can just define new custom nodes that implement certain operations and may contain parameters that can be adapted via local search. This PR includes a fix for nodes that use more than one parameter and example for how to make use of the new power of #178 ;)

Should we include this parametrized add as a new default node? Or is is so simple now to define nodes that we don't need it?

closes #98

### DIFF
--- a/examples/example_parametrized_nodes.py
+++ b/examples/example_parametrized_nodes.py
@@ -1,0 +1,172 @@
+"""
+Example for evolutionary regression with parametrized nodes
+===========================================================
+
+Example demonstrating the use of Cartesian genetic programming for a
+regression task that requires fine tuning of constants in parametrized
+nodes. This is achieved by introducing a new node, "ParametrizedAdd"
+which produces a scaled and shifted version of the sum of its inputs.
+
+"""
+
+import functools
+import math
+import matplotlib.pyplot as plt
+import numpy as np
+import scipy.constants
+import torch
+
+import cgp
+
+
+# %%
+# We first define a new node that adds two inputs then scales and
+# finally shifts the result. The scale ("w") and shift factors ("b")
+# are parameters that are adapted by local search. We need to define
+# the arity of the node, callables for the initial values for the
+# parameters and the operation of the node as a string. In this string
+# parameters are enclosed in angle brackets, inputs are denoted by "x_i"
+# with i representing their corresponding index.
+
+
+class ParametrizedAdd(cgp.OperatorNode):
+    """A node that adds its two inputs.
+
+    The result of addition is scaled by w and shifted by b. Both these
+    parameters can be adapted via local search are passed on from
+    parents to their offspring.
+
+    """
+
+    _arity = 2
+    _initial_values = {"<w>": lambda: 1.0, "<b>": lambda: 0.0}
+    _def_output = "<w> * (x_0 + x_1) + <b>"
+
+
+# %%
+# We define a target function which contains numerical constants that
+# are not available as constants for the search and need to be found
+# by local search on parameterized nodes.
+
+
+def f_target(x):
+    return math.pi * (x[:, 0] + x[:, 1]) + math.e
+
+
+# %%
+# Then we define a differentiable(!) inner objective function for the
+# evolution. This function accepts a torch class as a parameter. It
+# returns the mean-squared error between the output of the forward
+# pass of this class and the target function evaluated on a set of
+# random points. This inner objective is then used by actual objective
+# function to determine the fitness of the individual.
+
+
+def inner_objective(f, seed):
+    torch.manual_seed(seed)
+    batch_size = 500
+    x = torch.DoubleTensor(batch_size, 2).uniform_(-5, 5)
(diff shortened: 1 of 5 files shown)


examples/example_parametrized_nodes.py
```diff
@@ -0,0 +1,110 @@
+import functools
```
mschmidt87, line 1: Please add more docstrings to get a nice documentation of this example in our sphinx-generated docs.

jakobj:
good point! i've adapted the example to include proper documentation that can be properly rendered. please have another look.

examples/example_parametrized_nodes.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+# %%
+# We define a target function which contains numerical constants that
+# are not available as constants for the search and need to be found
+# by local search on parameterized nodes.
+
+
+def f_target(x):
+    return np.pi * x[:, 0] + 1.0 + np.pi
```
mschmidt87, line 52: why the `+ 1.0` here?

examples/example_parametrized_nodes.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+# %%
+# Then we define a differentiable(!) objective function for the
```
mschmidt87, line 56:
```suggestion
# Then we define a differentiable(!) inner objective function for the
```

examples/example_parametrized_nodes.py
```diff
@@ -0,0 +1,160 @@
...
+# evolution. It accepts a torch tensor as input variable and uses
```
mschmidt87, line 57: Doesn't `inner_objective` take a function `f` as input?

mschmidt87, line 57:
```suggestion
# evolution. It accepts a function that returns a torch tensor as input variable and uses
```

examples/example_parametrized_nodes.py
```diff
@@ -0,0 +1,160 @@
...
+
+obj = functools.partial(objective, seed=population_params["seed"])
+
+cgp.evolve(
+    pop, obj, ea, **evolve_params, print_progress=True, callback=recording_callback,
+)
+
+# %%
+# After the evolutionary search has ended, we print the expression
+# with the highest fitness and plot the progression of the search.
```
mschmidt87, line 144: Here, I would be interested to see the development of expressions over the course of the evolution. The fitness seems to increase very smoothly. Has the evolution found the right expression right in the beginning and then just keeps finetuning this over the course of the evolution?

examples/example_parametrized_nodes.py
jakobj: no reason, just to avoid the same constant twice. i've changed this to `math.pi * x[:, 0] + math.e` now to avoid such confusion

examples/example_parametrized_nodes.py
jakobj: i've improved the output of the example, this now shows that for the first few generations an expression without constants is used, then the correct form is found and the rest of the generations are busy finding the correct numerical values.

jakobj:
thanks for your input @mschmidt87! i've included your suggestions, please have another look

jakobj:
thanks! added the example to travis, sqashed, now waiting for green light from travis before merging

PR #200, opened 2020-07-20
Add links to source code on github to API docs

mschmidt87:
This PR adds links to the source code to the API docs in the documentation.

To achieve, we:
- add linkcode extension of sphinx
- add linkcode_resolve function from numpy docs (see https://github.com/numpy/numpy/blob/a0028bca0117874606bce99261d978df8d3f6610/doc/source/conf.py#L332)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,6 +10,8 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
+import cgp
+import inspect
 import os
 import sys
 import msmb_theme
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -63,3 +70,54 @@
     "examples_dirs": "../examples",  # path to your example scripts
     "gallery_dirs": "auto_examples",  # path to where to save gallery generated output
 }
+
+
+def linkcode_resolve(domain, info):
```
jakobj, line 75: maybe add reference to original implementation here?

mschmidt87: yes, thanks, added it.

Issue #196, opened 2020-07-20
Investigate moving `determine_alternative_permissible_values` function for genes to `__init__` of `Genome`.


mschmidt87:
In the implementation introduced in #180 , we check for alternative permissible values for each gene every time we call the mutate method. This is wasteful since we can just do this once in the `__init__` of `Genome` and store the results in a data structure.

This ticket is therefore about investigating this and implementing the change if deemed feasible.

HenrikMettler:
As a side note: We can not directly initialize the alternative permissible value, but only all permissible values for a gene, since the former depends on the current allele (basically: `alternative_permissible_values = all_permissible_values \ {current_value}`)

jakobj:
for completeness i'm quoting from #180 
> this might save quite some computation time at the expense of increased memory usage and slower instantiation of Genome objects. as always, the root of all evil is premature optimization, so i wouldn't implement this without profiling and finding that this is indeed a bottleneck. maybe a more lightweight compromise would be some kind of memoization, e.g., decorating the function determining all permissible values with the lru cache decorator.

mschmidt87:
Given that the memory footprint of the data structure to be stored is probably very small, and that the function determining all permissible value is always called with the same argument (the configuration of the dna, if it it determine *all* permissible value as suggested by @HenrikMettler above), I think that memoization is not a good alternative.

Issue #158, opened 2020-06-24
Change `Genome.dna` from List[int] to numpy.ndarray


jakobj:
The dna list has homogeneous types (all entries are integers), so we might as well use a numpy array to store its contents. For example this would make operations on the whole dna, e.g., as suggested in #157, faster.

jakobj:
i think as we make this change we can also change from a one-dimensional representation with our custom access functions to a three-dimensional structure with (rows, cols, genes) dimensions. i think this would simplify the implementation significantly. however, we'd need to think a bit about how to treat input/output nodes as their number might exceed the row count.

HenrikMettler:
You can assign me, if you want. Imo there are 3 options: 

- Do a 2 dim array: `n_unit x n_genes_per_unit` this has the disadvantage that `levels-back` is more complex to apply
- split the Dna into 3 arrays: 1 for the inputs (`n_inputs x 1 x n_genes`), 1 for the hidden (`n_rows x n_cols x n_genes)` 1 for the output (`1,1,n_genes`) - apart from the obvious disadvantages, this would have the advantage that we could also change the arity of input and output since they are seperate
- fill inputs and outputs with Nan values to match the dimensionality

jakobj:
good suggestions! this is not as easy as i thought as the number of inputs, rows, and outputs can differ, so a 2d array would always have regions that are not used, as you also observe. maybe we should rethink the benefits of changing the representation of the dna. what would we gain over the current implementation by either of the three options?

HenrikMettler:
I guess I am not telling any news that the theoretical benefits are less memory and faster compute (see eg: https://www.geeksforgeeks.org/python-lists-vs-numpy-arrays/). 
But in practical if I am not mistaking only few things happen with the dna per individual per generation: Evaluation to a function (learning-rule) and replacement of some values in mutation. I don't think that in any practical use case this is computationally expensive compared to the fitness evaluation. So in essence i don't think the benefits would be large

mschmidt87:
I think that one advantage of a representation as a multi-dimensional array would be that a lot of our code becomes easier. Especially the [`iter_input_regions`](https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/genome.py#L277), and other iterators become easier because we don't have to do this complicated indexing of the 1-d data structure.

mschmidt87:
Any opinions here?

jakobj:
[...]

PR #201, opened 2020-07-21
Permissible values per gene in initialization of genome

HenrikMettler:
Determine permissible values per gene in initialization of genome. Addresses #196. 
I was not sure which of the `determine_permissible_value_xxx` functions should be private. And please tell me whether you think testing with the permissible values for a single genome is sufficient.

### DIFF
--- a/cgp/genome.py
+++ b/cgp/genome.py
@@ -90,6 +90,9 @@ class Genome:
         # dictionary to store values of Parameter nodes
         self._parameter_names_to_values: Dict[str, float] = {}
 
+        # list of permissible values for every gene
+        self._permissible_values: List[np.ndarray] = self.determine_permissible_values()
+
     def __getitem__(self, key: int) -> int:
         if self.dna is None:
             raise RuntimeError("dna not initialized")
@@ -133,6 +136,70 @@ class Genome:
         s = s[:-3]
         s += ")"
         return s
+
+    def determine_permissible_values_per_gene(self, gene_idx: int) -> np.ndarray:
+        region_idx = gene_idx // self._length_per_region
+
+        if self._is_input_region(region_idx):
+            return self._determine_permissible_values_input_region(gene_idx)
+
+        elif self._is_hidden_region(region_idx):
+            return self._determine_permissible_values_hidden_region(gene_idx, region_idx)
+
+        elif self._is_output_region(region_idx):
+            return self._determine_permissible_values_output_region(gene_idx)
+
+        else:
+            assert False  # should never be reached
+
+    def determine_permissible_values(self) -> List[np.ndarray]:
+        """Determine permissible values for every gene.
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        ----------
+        permissible_values
+            List[numpy.ndarray]: List of permissible values for every gene
+        """
+        permissible_values: List[np.ndarray] = []
+        for gene_idx in range(self._n_genes):
+            permissible_values_per_gene = self.determine_permissible_values_per_gene(gene_idx)
+            permissible_values.append(permissible_values_per_gene)
+        return permissible_values
+
+    def _determine_permissible_values_input_region(self, gene_idx: int,) -> np.ndarray:
+
+        if self._is_function_gene(gene_idx):
+            return np.array(self._id_input_node)
+        else:
+            return np.array(self._id_unused_gene)
+
+    def _determine_permissible_values_hidden_region(
+        self, gene_idx: int, region_idx: int
+    ) -> np.ndarray:
+
+        if self._is_function_gene(gene_idx):
+            return np.arange(len(self._primitives._primitives))
+
+        elif self._is_hidden_input_gene(gene_idx, region_idx):
+            return np.array(self._permissible_inputs(region_idx))
+
+        else:
+            assert False  # should never be reached
+
+    def _determine_permissible_values_output_region(self, gene_idx: int,) -> np.ndarray:
+
+        if self._is_function_gene(gene_idx):
+            return np.array(self._id_output_node)
+        else:
+            input_index = gene_idx % self._length_per_region - 1
+            if input_index == 0:
+                return np.array(self._permissible_inputs_for_output_region())
+            else:
+                return np.array(self._id_unused_gene)
 
     def _create_input_region(self) -> List[int]:
 
@@ -399,64 +466,18 @@ class Genome:
 
             region_idx = gene_idx // self._length_per_region
 
-            permissible_values = self._determine_alternative_permissible_values(
-                gene_idx, allele, region_idx
-            )
-            if len(permissible_values) > 0:
+            permissible_values = self._permissible_values[gene_idx]
+            permissible_alternative_values = permissible_values[permissible_values != allele]
 
-                dna[gene_idx] = rng.choice(permissible_values)
+            if len(permissible_alternative_values) > 0:
+
+                dna[gene_idx] = rng.choice(permissible_alternative_values)
                 silent = region_idx not in active_regions
 
                 only_silent_mutations = only_silent_mutations and silent
 
         self.dna = dna
         return only_silent_mutations
-
-    def _determine_alternative_permissible_values(
-        self, gene_idx: int, allele: int, region_idx: int
-    ) -> List[int]:
-
-        if self._is_input_region(region_idx):
-            return []  # genes in input regions have no alternative permissible values
-
-        elif self._is_hidden_region(region_idx):
-            return self._determine_alternative_permissible_values_hidden_gene(
-                gene_idx, allele, region_idx
-            )
-
-        elif self._is_output_region(region_idx):
-            return self._determine_alternative_permissible_values_output_gene(gene_idx, allele)
-
-        else:
-            assert False  # should never be reached
-
-    def _determine_alternative_permissible_values_hidden_gene(
-        self, gene_idx: int, allele: int, region_idx: int
-    ) -> List[int]:
-        if self._is_function_gene(gene_idx):
-            permissible_values = list(np.arange(len(self._primitives._primitives)))
-
-        elif self._is_hidden_input_gene(gene_idx, region_idx):
-            permissible_values = self._permissible_inputs(region_idx)
-
-        else:
-            assert False
-        permissible_values.remove(allele)
-        return permissible_values
-
-    def _determine_alternative_permissible_values_output_gene(
-        self, gene_idx: int, allele: int
-    ) -> List[int]:
-        input_index = (
-            gene_idx % self._length_per_region
-        )  # assumes that the second gene in all output regions is the index of the input
-        if input_index == 1:
-            permissible_values = self._permissible_inputs_for_output_region()
-            permissible_values.remove(allele)
-        else:
-            permissible_values = []
-
-        return permissible_values
 
     @property
     def primitives(self) -> Primitives:
--- a/test/test_genome.py
+++ b/test/test_genome.py
@@ -476,76 +476,61 @@ def test_only_silent_mutations(genome_params, mutation_rate, rng_seed):
     assert not only_silent_mutations
 
 
-def test_permissible_values_hidden(rng_seed):
-    genome_params = {
-        "n_inputs": 2,
-        "n_outputs": 1,
-        "n_columns": 3,
-        "n_rows": 3,
-        "levels_back": 2,
-        "primitives": (cgp.Add, cgp.Sub, cgp.ConstantFloat),
-    }
+def test_permissible_values(genome_params):
+    n_inputs = 2
+    n_outputs = 1
+    n_columns = 3
+    n_rows = 3
+    levels_back = 2
+    primitives = (cgp.Add, cgp.Sub, cgp.ConstantFloat)
+    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives)
 
-    genome = cgp.Genome(**genome_params)
+    permissible_function_gene_values = np.arange(len(genome._primitives._primitives))
+    permissible_inputs_first_internal_column = np.array([0, 1])
+    permissible_inputs_second_internal_column = np.array([0, 1, 2, 3, 4])
+    permissible_inputs_third_internal_column = np.array([0, 1, 2, 3, 4, 5, 6, 7])
+    permissible_inputs_output_gene = np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
 
-    # test function gene
-    gene_idx = 6  # function gene of first hidden region
-    allele = 0  #
-    region_idx = None  # can be any number, since not touched for function gene
-    permissible_func_gene_values = genome._determine_alternative_permissible_values_hidden_gene(
-        gene_idx, allele, region_idx
-    )
-    assert permissible_func_gene_values == [
-        1,
-        2,
-    ]  # function idx 1,2 (cgp.Sub, cgp.ConstantFloat)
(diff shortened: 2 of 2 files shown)


HenrikMettler:
Looking at the discussion in #158 i realized that `permissible_values_per_gene` should rather be a numpy array than a list (and therefore `permissible_values` a list of numpy array. I will address this with a new commit.

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+            permissible_values = np.arange(len(self._primitives._primitives))
+
+        elif self._is_hidden_input_gene(gene_idx, region_idx):
+            permissible_values = np.array(self._permissible_inputs(region_idx))
+
+        else:
+            assert False
+        return permissible_values
+
+    def _determine_permissible_values_output_gene(self, gene_idx: int,) -> List[int]:
```
mschmidt87, line 199: The type annotation for the return value should be `np.array`, I think?

test/test_genome.py
```diff
@@ -476,76 +476,48 @@ def select_gene_indices_non_silent(mutation_rate, rng):
...
+    permissible_values_expected = [
+        np.array(ID_INPUT_NODE),
+        np.array(ID_NON_CODING_GENE),
+        np.array(ID_NON_CODING_GENE),
+        np.array(ID_INPUT_NODE),
+        np.array(ID_NON_CODING_GENE),
+        np.array(ID_NON_CODING_GENE),
+        np.arange(len(genome._primitives._primitives)),
+        np.array([0, 1]),
+        np.array([0, 1]),
```
mschmidt87, line 490: Can you define some variable that represents `np.array([0,1])` etc. here so that you don't need to repeat this expression so many times? Ideally with a meaningful name that explains why the permissible values are 0,1 here, (and 0,1,2,3,4 elsewhere etc.).

cgp/genome.py
HenrikMettler: Yes, thanks!

test/test_genome.py
HenrikMettler: I tried to add meaningful names in the latest commit, please tell me if they are ok. Also i moved the genome params inside the function, since the permissible_values_expected depend strongly on the specific values.

cgp/genome.py
```diff
@@ -90,6 +90,9 @@ def __init__(
         # dictionary to store values of Parameter nodes
         self._parameter_names_to_values: Dict[str, float] = {}
 
+        # list of permissible values for every gene
+        self._permissible_values = self.determine_permissible_values()
```
jakobj, line 94:
```suggestion
        self._permissible_values: List[np.ndarray] = self.determine_permissible_values()
```

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+        """Determine permissible values for every gene.
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        ----------
+        permissible_values
+            List[numpy array]: List of numpy arrays over permissible values for every gene
```
jakobj, line 164:
```suggestion
            List[numpy.ndarray]: List of permissible values for every gene
```

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+        """
+        permissible_values = []
```
jakobj, line 166:
```suggestion
        permissible_values: List[np.ndarray] = []
```

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+            permissible_values_per_gene = self.determine_permissible_value_per_gene(
+                gene_idx, region_idx
+            )
+            permissible_values.append(permissible_values_per_gene)
+        return permissible_values
+
+    def _determine_permissible_values_input_gene(self, gene_idx: int, region_idx: int) -> np.array:
+
+        if region_idx == 0:
+            region_idx += self._length_per_region  # avoid issues with x % 0
```
jakobj, line 178: hmm, what does this block do?

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
         s += ")"
         return s
 
+    def determine_permissible_value_per_gene(self, gene_idx: int, region_idx: int) -> np.array:
+
+        if self._is_input_region(region_idx):
+            return self._determine_permissible_values_input_gene(gene_idx, region_idx)
```
jakobj, line 143:
```suggestion
            return self._determine_permissible_values_input_region(gene_idx, region_idx)
```

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+
+        elif self._is_hidden_region(region_idx):
+            return self._determine_permissible_values_hidden_gene(gene_idx, region_idx)
```
jakobj, line 146:
```suggestion
            return self._determine_permissible_values_hidden_region(gene_idx, region_idx)
```

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+
+        elif self._is_output_region(region_idx):
+            return self._determine_permissible_values_output_gene(gene_idx)
```
jakobj, line 149:
```suggestion
            return self._determine_permissible_values_output_region(gene_idx)
```

cgp/genome.py
jakobj: could you also further specify the type, e.g., `List[np.ndarray[int]]`?

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+            return np.array(self._id_input_node)
+        else:
+            return np.array(self._id_unused_gene)
+
+    def _determine_permissible_values_hidden_gene(
+        self, gene_idx: int, region_idx: int
+    ) -> np.array:
+
+        if self._is_function_gene(gene_idx):
+            permissible_values = np.arange(len(self._primitives._primitives))
```
jakobj, line 190:
```suggestion
            return np.arange(len(self._primitives._primitives))
```

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+
+        elif self._is_hidden_input_gene(gene_idx, region_idx):
+            permissible_values = np.array(self._permissible_inputs(region_idx))
```
jakobj, line 193:
```suggestion
            return np.array(self._permissible_inputs(region_idx))
```

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+
+        else:
+            assert False
+        return permissible_values
```
jakobj, line 197:
```suggestion
            assert False  # should never be reached
```

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+
+    def _determine_permissible_values_output_gene(self, gene_idx: int,) -> np.array:
+        input_index = (
+            gene_idx % self._length_per_region
+        )  # assumes that the second gene in all output regions is the index of the input
+        if input_index == 0:
+            permissible_values = np.array(self._id_output_node)
+        elif input_index == 1:
+            permissible_values = np.array(self._permissible_inputs_for_output_region())
```
jakobj, line 206:
```suggestion
            return np.array(self._permissible_inputs_for_output_region())
```
always return early when possible ;)

cgp/genome.py
```diff
@@ -134,6 +137,78 @@ def __repr__(self) -> str:
...
+        else:
+            permissible_values = np.array(self._id_unused_gene)
+
+        return permissible_values
```
jakobj, line 210:
```suggestion
```

cgp/genome.py
```diff
@@ -399,65 +474,19 @@ def mutate(self, mutation_rate: float, rng: np.random.RandomState):
 
             region_idx = gene_idx // self._length_per_region
 
-            permissible_values = self._determine_alternative_permissible_values(
-                gene_idx, allele, region_idx
-            )
-            if len(permissible_values) > 0:
+            permissible_values = self._permissible_values[gene_idx]
+            permissible_alternative_values = permissible_values[permissible_values != allele]
```
jakobj, line 478: nice one! :1st_place_medal:

test/test_genome.py
```diff
@@ -476,76 +476,61 @@ def select_gene_indices_non_silent(mutation_rate, rng):
...
+    genome = cgp.Genome(n_inputs, n_outputs, n_columns, n_rows, levels_back, primitives)
+
+    permissible_function_gene_values = np.arange(len(genome._primitives._primitives))
+    permissible_inputs_first_internal_column = np.array([0, 1])
+    permissible_inputs_second_internal_column = np.array([0, 1, 2, 3, 4])
+    permissible_inputs_third_internal_column = np.array([0, 1, 2, 3, 4, 5, 6, 7])
+    permissible_inputs_output_gene = np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+
+    permissible_values_expected = [
+        np.array(ID_INPUT_NODE),
```
jakobj, line 495: wow :rocket:

cgp/genome.py
HenrikMettler: Below (line 180) i check whether it is the first gene in the region: 
```python
 if gene_idx % (region_idx * self._length_per_region) == 0:

```
and the % operator can not be used with `region_idx = 0`. Therefore I pretend I am at unit 1 instead of at unit 0. (it is a bit a dirty hack i know, but i did for the lack of a better idea)

cgp/genome.py
HenrikMettler: :blush:

cgp/genome.py
mschmidt87: I think that `np.ndarray[int]` is not a valid type hint. There is the `nptyping` package which supposedly achieves such functionality but I think this is a bit of an overkill here.

cgp/genome.py
```diff
@@ -134,6 +137,77 @@ def __repr__(self) -> str:
         s += ")"
         return s
 
+    def determine_permissible_value_per_gene(self, gene_idx: int, region_idx: int) -> np.array:
```
mschmidt87, line 140:
```suggestion
    def determine_permissible_value_per_gene(self, gene_idx: int, region_idx: int) -> np.ndarray:
```

mschmidt87: Sorry, on 2nd thought, I think the type hint for the array should be like this.

mschmidt87: same for all other type hints.

cgp/genome.py
```diff
@@ -134,6 +137,77 @@ def __repr__(self) -> str:
...
+        elif self._is_hidden_region(region_idx):
+            return self._determine_permissible_values_hidden_region(gene_idx, region_idx)
+
+        elif self._is_output_region(region_idx):
+            return self._determine_permissible_values_output_region(gene_idx)
+
+        else:
+            assert False  # should never be reached
+
+    def determine_permissible_values(self) -> List[np.array]:
```
mschmidt87, line 154:
```suggestion
    def determine_permissible_values(self) -> List[np.ndarray]:
```

cgp/genome.py
```diff
@@ -134,6 +137,77 @@ def __repr__(self) -> str:
...
+            region_idx = gene_idx // self._length_per_region
+            permissible_values_per_gene = self.determine_permissible_value_per_gene(
+                gene_idx, region_idx
+            )
+            permissible_values.append(permissible_values_per_gene)
+        return permissible_values
+
+    def _determine_permissible_values_input_region(
+        self, gene_idx: int, region_idx: int
+    ) -> np.array:
```
mschmidt87, line 177:
```suggestion
    ) -> np.ndarray:
```

cgp/genome.py
```diff
@@ -134,6 +137,77 @@ def __repr__(self) -> str:
...
+            region_idx += self._length_per_region  # avoid issues with x % 0
+
+        if gene_idx % (region_idx * self._length_per_region) == 0:
+            return np.array(self._id_input_node)
+        else:
+            return np.array(self._id_unused_gene)
+
+    def _determine_permissible_values_hidden_region(
+        self, gene_idx: int, region_idx: int
+    ) -> np.array:
```
mschmidt87, line 189:
```suggestion
    ) -> np.ndarray:
```

cgp/genome.py
```diff
@@ -134,6 +137,77 @@ def __repr__(self) -> str:
...
+        if self._is_function_gene(gene_idx):
+            return np.arange(len(self._primitives._primitives))
+
+        elif self._is_hidden_input_gene(gene_idx, region_idx):
+            return np.array(self._permissible_inputs(region_idx))
+
+        else:
+            assert False  # should never be reached
+
+    def _determine_permissible_values_output_region(self, gene_idx: int,) -> np.array:
```
mschmidt87, line 200:
```suggestion
    def _determine_permissible_values_output_region(self, gene_idx: int,) -> np.ndarray:
```

cgp/genome.py
```diff
@@ -90,6 +90,9 @@ def __init__(
         # dictionary to store values of Parameter nodes
         self._parameter_names_to_values: Dict[str, float] = {}
 
+        # list of permissible values for every gene
+        self._permissible_values: List[np.ndarray[int]] = self.determine_permissible_values()
```
mschmidt87, line 94: I think the `np.ndarray[int]` is not a valid type hint. However, `mypy` anyway ignores all numpy type hints, because numpy isn't type annotated, so this does not lead to a problem.

To be consistent with the other hints, I'd recommend to remove the `[int]` here.

cgp/genome.py
jakobj: ah, i see. i would remove this and adjust the if statement below like such:
```python
 if gene_idx % (max(1, region_idx) * self._length_per_region) == 0:
```
but that's just a personal preference. either way it needs to be clearly documented :)

jakobj: ah, there's actually a function for this: `self._is_function_gene(gene_idx)` should do the trick. please also fix this in the other `if`s

cgp/genome.py
```diff
@@ -134,6 +137,72 @@ def __repr__(self) -> str:
...
+    def _determine_permissible_values_output_region(self, gene_idx: int,) -> np.ndarray:
+        input_index = (
+            gene_idx % self._length_per_region
+        )  # assumes that the second gene in all output regions is the index of the input
+        if input_index == 0:
+            return np.array(self._id_output_node)
+        elif input_index == 1:
+            return np.array(self._permissible_inputs_for_output_region())
+        else:
+            return np.array(self._id_unused_gene)
```
jakobj, line 204:
```suggestion
        if self._is_function_gene(gene_idx):
            return np.array(self._id_output_node)
        else:
            input_index = gene_idx % self._length_per_region - 1
            if input_index == 0:
                return np.array(self._permissible_inputs_for_output_region())
            else:
                return np.array(self._id_unused_gene)
```

cgp/genome.py
```diff
@@ -134,6 +137,72 @@ def __repr__(self) -> str:
         s += ")"
         return s
 
+    def determine_permissible_value_per_gene(self, gene_idx: int, region_idx: int) -> np.ndarray:
```
jakobj, line 140:
```suggestion
    def determine_permissible_values_per_gene(self, gene_idx: int, region_idx: int) -> np.ndarray:
```

HenrikMettler:
done @jakobj

jakobj:
thanks! and merging :)

PR #204, opened 2020-07-23
Fix reproducability in mountain_car example

### LINKED ISSUE #202: Mountain car example is not deterministic
Running the mountain_car example twice in a row yields different results:
mountain_car

In addition, I also observed runs where the environment was not solved by the found champion.

I think the example should behave deterministic, so we should fix this.

HenrikMettler:
Done by setting the seed before sampling observations. Not done for the visualization of the champion in the end. Closes #202

### DIFF
--- a/examples/example_mountain_car.py
+++ b/examples/example_mountain_car.py
@@ -61,8 +61,8 @@ def inner_objective(f, seed, n_total_steps, *, render):
 
     env = gym.make("MountainCarContinuous-v0")
 
-    observation = env.reset()
     env.seed(seed)
+    observation = env.reset()
 
     cum_reward_all_episodes = []
     cum_reward_this_episode = 0


mschmidt87:
I can confirm that this change leads to reproducible behavior in the example.

However, with this seed, the environment is not solved (average reward over 100 consecutive trials 87.8 ) and the evolution finds a rather complicated solution. 
See this screenshot:
mc

Thus, we need to find another seed that leads to a better solution. @jakobj

jakobj:
> I can confirm that this change leads to reproducible solution finding in the example.
> However, two issues:
> 
>     * with this seed, the environment is not solved (average reward over 100 consecutive trials 87.8 ) and the evolution finds a rather complicated solution.

previously every individual had a different initial condition as the seed was only set after the first observation was sampled. this was not done on purpose, but it might be that it had a "regularization" effect that helped guide the search to good regions of the search space. one possibility to achieve this regularization while maintaining reproducibility would be, as always, to evaluate each solution on several different runs with different initial conditions. i will try and see whether this works.

> 
>     * when testing whether the solution solves the environment, the resulting average reward is slightly different between two consecutive runs (87.73259 vs. 87.73128). Why is that?

i think that also in that part of the code the call to `env.seed` should be before sample the first observation.

mschmidt87:
> previously every individual had a different initial condition as the seed was only set after the first observation was sampled. this was not done on purpose, but it might be that it had a "regularization" effect that helped guide the search to good regions of the search space. one possibility to achieve this regularization while maintaining reproducibility would be, as always, to evaluate each solution on several different runs with different initial conditions. i will try and see whether this works.

What you're saying, can make sense, but let me just add that in the example as it is on `master`, some runs yield low performance, so this "regularizing effect", if it exists, does not guarantee convergence to a good solution per se.

jakobj:
@HenrikMettler i've created a PR against your feature branch that contains some suggestions on how to improve this PR as that was easier than annotating the code on github or describing the changes in prose. please have a look and update this PR accordingly

jakobj:
closing in favor of #204 as suggested by @HenrikMettler

PR #205, opened 2020-07-27
Use figures from examples in documentation

mschmidt87:
This PR does the following this:

- It configures sphinx such that the gallery of examples will contain the images produced by the example scripts. To that end, running `make html`  in the `docs/` folder now executes the example scripts one by one and collects the figures created by the scripts.
- To enable quick debugging of the documentation without having to run the examples every time, I introduced `make html-noplot` which skips running the examples (and produces a documentation without the figures). This is of much quicker to run.
- In the CI, I removed the step to run the examples explicitly because this is now done in the documentation build step, implicitly.

Note: It is not possible to render the openAI gym environment when running the mountain car example script with sphinx-gallery. (This will fail with an error). I suspect the problem is that the `pyglet` library that the gym is using for rendering does not support headless rendering (this is a known shortcoming of this library). Therefore, the only possible way to make the docs run is to comment out this line in the example. I added an explanation to the docstring.

To test this PR, run
```bash
cd docs/
make html
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,18 +30,12 @@ script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
-  - if [ "$DEP" = "[all]" ]; then
+  - if [ "$DEP" = "[all]" -a $TRAVIS_PYTHON_VERSION = 3.8 ]; then
+       pip install gym || exit 1;
        make -C docs/ html || exit 1;
-       touch docs/_build/html/.nojekyll
-       
+       touch docs/_build/html/.nojekyll || exit 1;
     fi
   - pytest --cov=cgp
-  - if [ "$DEP" = "[all]" -a $TRAVIS_PYTHON_VERSION = 3.8 ]; then
-       python examples/example_evo_regression.py || exit 1 ;
-       python examples/example_differential_evo_regression.py || exit 1;
-       python examples/example_caching.py || exit 1;
-       python examples/example_parametrized_nodes.py || exit 1;
-    fi
 after_success:
   - coveralls
 deploy:
--- a/examples/example_mountain_car.py
+++ b/examples/example_mountain_car.py
@@ -268,6 +268,8 @@ def visualize_behaviour_for_evolutionary_jumps(seed, history, only_final_solutio
 
 # %%
 # Finally, we execute the evolution and visualize the results.
+# To animate the behavior of the car for the found expression, uncomment
+# the last line of the example.
 
 
 if __name__ == "__main__":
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -30,17 +30,12 @@ script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
-  - if [ "$DEP" = "[all]" ]; then
+  - if [ "$DEP" = "[all]" -a $TRAVIS_PYTHON_VERSION = 3.8 ]; then
+       pip install gym;  # for the mountain car example
```
jakobj, line 34:
```suggestion
       pip install gym;  # necessary for the mountain car example
```

examples/example_mountain_car.py
```diff
@@ -265,6 +265,8 @@ def f(x):
 
 # %%
 # Finally, we execute the evolution and visualize the results.
+# To animate the behavior of the car for the final solution, uncomment
```
jakobj, line 268:
```suggestion
# To animate the behavior of the car for the found expression, uncomment
```

mschmidt87:
> however, i'm really worried about the runtime of CI with this configuration. on my machine it took ~15min to build the documentation with executing the examples, this can easily get annoying, no? maybe we should decrease the maximal number of generations, e.g., in the mountain car example (~12min)?

Yes, I agree, this will become more severe with more examples being added to the repo. As a compromise, we could say that we only build the full documentation incl. running the examples just before merging the PR, i.e., not on every travis run. Not sure how to organize this with Travis, but it might be possible.

What is definitely possible: Only build the full docs on deployment of the documentation, i.e. *after* merging the PR. That would mean, though, that we don't test the examples before merging a PR. What do you think?

jakobj:
honestly, i don't know what the best way forward is here. long CI runtime can be quite frustrating, however, having to fix examples after a PR was merged also doesn't sound like fun. i like your first suggestion, but it sounds like we'd need to manually trigger the doc build, wouldn't we? if afraid any manual step in the CI process will sooner or later lead to failure. we could try to keep the runtime of the examples really low, let's say each _must_ finish <1min, and exclude all examples that take longer from the doc build (if that's possible). well, if we have many examples, this would also quickly add up. let me think about this a bit more.

mschmidt87:
I agree that this is a bit tricky. But I think the idea of reducing the runtime of the examples is good. Let's have this discussion not in this PR, because it's actually not related to the change introduced here. I opened #208  for discussion and sketched the three options, I could come up with. I guess you can infer from the text which one I prefer. ;-)

mschmidt87:
The CI fails because the mountain car example is taking too long. Let's wait for #206 to get merged, before we try again.

mschmidt87:
CI passes, please take another look @jakobj

PR #206, opened 2020-07-27
Make Mountain Car example deterministic

### LINKED ISSUE #202: Mountain car example is not deterministic

jakobj:
building on #204 this PR makes the Mountain Car example reproducible by fixing the environment seed before the first observation is obtained. in addition, it fixes a couple of minor issues in this example and slightly adapts parameters to make sure a good solution is found in reasonable time.

closes #202

### DIFF
--- a/examples/example_mountain_car.py
+++ b/examples/example_mountain_car.py
@@ -24,6 +24,7 @@ except ImportError:
 
 import matplotlib.pyplot as plt
 import numpy as np
+import sympy
 import warnings
 
 import cgp
@@ -35,15 +36,11 @@ import cgp
 
 
 class ConstantFloatZeroPointOne(cgp.ConstantFloat):
-    def __init__(self, idx, inputs):
-        super().__init__(idx, inputs)
-        self._output = 0.1
+    _def_output = "0.1"
 
 
 class ConstantFloatTen(cgp.ConstantFloat):
-    def __init__(self, idx, inputs):
-        super().__init__(idx, inputs)
-        self._output = 10.0
+    _def_output = "10.0"
 
 
 # %%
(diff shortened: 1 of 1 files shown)


examples/example_mountain_car.py
```diff
@@ -55,30 +51,30 @@ def __init__(self, idx, inputs):
 # episodes.
 
 
-def inner_objective(f, seed, n_total_steps, *, render):
-
-    np.random.seed(seed)
+def inner_objective(f, seed, n_trials_per_individual, n_total_steps, *, render):
```
HenrikMettler, line 54: Wouldn't `n_episodes_per_individual` be proper RL-lingo?

HenrikMettler: ah I saw below that fast agents have more episodes, if they don't use all time steps. But I think it is a bit confusing that trial is bigger than episode. Maybe `n_runs_per_individual`

examples/example_mountain_car.py
```diff
@@ -106,12 +102,14 @@ def objective(ind, seed, n_total_steps):
...
-            cum_reward_all_episodes = inner_objective(f, seed, n_total_steps, render=False)
+            cum_reward_all_episodes = inner_objective(
+                f, seed, n_trials_per_individual, n_total_steps, render=False
+            )
 
         # more episodes are better, more reward is better
         n_episodes = float(len(cum_reward_all_episodes))
         mean_cum_reward = np.mean(cum_reward_all_episodes)
-        ind.fitness = n_episodes + 1.0 * mean_cum_reward
+        ind.fitness = n_episodes / n_trials_per_individual + mean_cum_reward
```
HenrikMettler, line 112: The first summand is a bit confusing, but renaming the variable as suggested above is enough I guess

examples/example_mountain_car.py
jakobj: yeah, this might indeed be a bit confusing naming. i'll use your suggestion :+1:

jakobj:
thanks! squashed and waiting for travis :)

PR #207, opened 2020-07-27
Extend cache decorator with functional equivalence checking

### LINKED ISSUE #61: Provide mechanisms to cache python functions/pytorch classes
Currently the example for caching uses SymPy expressions. This makes sense since SymPy can simplify expressions, thereby identifying functionally equivalent individuals even when their genotypes and even phenotypes (unsimplified) differ. However one might want to avoid converting the cartesian graph to a sympy expressions and rather work with python functions or pytorch classes. In these cases caching would still make sense for individuals that only different in inactive genes. How could we provide such functionality most easily?

jakobj:
This PR implements functional equivalence checking (see, e.g., https://arxiv.org/abs/2003.03384) to allow caching of (inner) objectives that are  called with individuals, not SymPy expressions. Functionally identical individuals are identified by evaluating them on a certain number of random input samples and hashing the output values.

This PR also increases the number of generations in the cache decorator example to allow evolution to converge, and adds a test that makes sure that the fitness history is identical with and without cache decorator.

~**WARNING** Currently I need to disable one other test to make the new tests pass. There is some really weird interaction on the SymPy level that I couldn't figure out yet. :|~

closes #61

### DIFF
--- a/examples/example_fec_caching.py
+++ b/examples/example_fec_caching.py
@@ -1,12 +1,14 @@
 """
-Example demonstrating the use of the caching decorator.
-=======================================================
+Example demonstrating the use of the caching decorator with functional equivalance checking
+===========================================================================================
 
 Caches the results of fitness evaluations in a pickle file
-('example_caching_cache.pkl'). To illustrate its practical use,
+(``example_fec_caching_cache.pkl``). To illustrate its practical use,
 compare the runtime of this script when you first call it vs. the
 second time and when you comment out the decorator on
-`inner_objective`."""
+`inner_objective`.
+
+"""
 
 import time
 
@@ -25,27 +27,37 @@ def f_target(x):
 # %%
 # We then define the objective function for the evolutionary
 # algorithm: It consists of an inner objective which we wrap with the
-# caching decorator. This decorator specifies a pickle file that will be used for
-# caching results of fitness evaluations. The inner objective is then used by the objective
-# function to compute (or retrieve from cache) the fitness of the individual.
+# caching decorator. This decorator specifies a pickle file that will
+# be used for caching results of fitness evaluations. In addition the
+# decorator accepts keyword arguments that specifiy the statistic of
+# the samples used for evaluation. The inner objective is then used by
+# the objective function to compute (or retrieve from cache) the
+# fitness of the individual.
 
 
-@cgp.utils.disk_cache("example_caching_cache.pkl")
-def inner_objective(expr):
-    """The caching decorator uses the function parameters to identify
-    identical function calls. Here, as many different genotypes
-    produce the same simplified SymPy expression we can use such
-    expressions as an argument to the decorated function to avoid
-    reevaluating functionally identical individuals.
-    Note that caching only makes sense for deterministic objective
-    functions, as it assumes that identical expressions will always
-    return the same fitness values.
+@cgp.utils.disk_cache(
+    "example_fec_caching_cache.pkl",
+    use_fec=True,
+    fec_seed=12345,
+    fec_min_value=-10.0,
+    fec_max_value=10.0,
+    fec_batch_size=5,
+)
+def inner_objective(ind):
+    """The caching decorator uses the return values generated from
+    providing random inputs to ind.to_numpy() to identify functionally
+    indentical individuals and avoid reevaluating them. Note that
+    caching only makes sense for deterministic objective functions, as
+    it assumes that identical phenotypes will always return the same
+    fitness values.
 
     """
+    f = ind.to_func()
+
     loss = []
-    for x0 in np.linspace(-2.0, 2.0, 100):
-        y = float(expr[0].subs({"x_0": x0}).evalf())
-        loss.append((f_target(x0) - y) ** 2)
+    for x_0 in np.linspace(-2.0, 2.0, 100):
+        y = f([x_0])
+        loss.append((f_target(x_0) - y) ** 2)
 
     time.sleep(0.25)  # emulate long fitness evaluation
 
@@ -56,7 +68,7 @@ def objective(individual):
     if individual.fitness is not None:
         return individual
 
-    individual.fitness = -inner_objective(individual.to_sympy())
+    individual.fitness = -inner_objective(individual)
 
     return individual
 
--- a/cgp/cartesian_graph.py
+++ b/cgp/cartesian_graph.py
@@ -408,8 +408,7 @@ class _C(torch.nn.Module):
             s = self._fill_parameter_values(s)
             # to get an expression that reflects the computational graph,
             # sympy should not automatically simplify the expression
-            with sympy.evaluate(False):
-                sympy_exprs.append(sympy.sympify(s))
+            sympy_exprs.append(sympy.sympify(s, evaluate=False))
 
         if not simplify:
             return sympy_exprs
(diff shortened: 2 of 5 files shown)


HenrikMettler:
This is really cool! :) I am just wondering whether if we would decide to make the standart use case the common (Miller 2020, Kaufmann 2017)  1 parent + 4 offspring population, if using Skip (Goldmann 2015: https://ieeexplore.ieee.org/document/6815728?arnumber=6815728) might not be more efficient. Since this would not require evaluating on some samples and eliminate the false positive / false negatives risk associated to the batch size here. The drawback of Skip of course is that it doesn't work well with more than one parent (it would be difficult to check whether an offspring is identical to any other individual in the population than its parent). Maybe having both could also be an option

jakobj:
as far as i understand, we already use "skip". if you look at https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/individual.py#L115 the fitness of an individual is only set to `None` if non-silent genes were changed, and combined with the early return for fitness != `None` in the objective functions (e.g., https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/examples/example_evo_regression.py#L54), this effectively implements "skip". wouldn't you agree?

functional equivalence checking or in general the cache decorator support slightly different use cases:
1) individuals with different active genes can implement the same function, e.g., `x_0 + x_0` vs `2 * x_0`; checking the active genes hence does not always identify _functionally_ equivalent individuals
2) an evolutionary run can have "recurrence", what i mean is that an individual in generation X can be functionally identical to an individual in generation X + 100; "skip" only applies to direct decedents, not to individuals which arrived at the same function via different evolutionary paths
3) broken individuals, e.g., those that generate infinite output values, are possibly identified before running an (expensive) evaluation

i would hence argue to certainly keep "skip" but provide the caching decorator as an additional method to reduce the number of function evaluations.

HenrikMettler:
> as far as i understand, we already use "skip". if you look at https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/individual.py#L115 the fitness of an individual is only set to `None` if non-silent genes were changed, and combined with the early return for fitness != `None` in the objective functions (e.g., https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/examples/example_evo_regression.py#L54), this effectively implements "skip". wouldn't you agree?

Ah yes, I agree. But after reading Goldmans paper I would maybe tend to argue for changing the while loop in `hl_api.py` https://github.com/Happy-Algorithms-League/hal-cgp/blob/4f4542854039f73e559bc1c2e22affd3e2729ec3/cgp/hl_api.py#L59 from to `while num_fitness_evaluations < max_num_fitness_evaluations:`. Since this is somewhat "more" what you want to limit. But we should probably discuss this in another issue.

> 
> 
>     3. broken individuals, e.g., those that generate infinite output values, are possibly identified before running an (expensive) evaluation
> 

This is a very good point, I didn't think of it. Surely for that it can be very useful.

jakobj:
> > as far as i understand, we already use "skip". if you look at https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/individual.py#L115 the fitness of an individual is only set to `None` if non-silent genes were changed, and combined with the early return for fitness != `None` in the objective functions (e.g., https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/examples/example_evo_regression.py#L54), this effectively implements "skip". wouldn't you agree?
> 
> Ah yes, I agree. But after reading Goldmans paper I would maybe tend to argue for changing the while loop in `hl_api.py`
> 
> https://github.com/Happy-Algorithms-League/hal-cgp/blob/4f4542854039f73e559bc1c2e22affd3e2729ec3/cgp/hl_api.py#L59
> from to `while num_fitness_evaluations < max_num_fitness_evaluations:`. Since this is somewhat "more" what you want to limit. But we should probably discuss this in another issue.

that's an interesting point, a limit on the number of objective evaluations seems to be a reasonable hyperparameter, for example it allows you to more accurately estimate the maximal run time of your script. my impression is however that most(?) implementations allow users to define the maximal number of generations instead. if this is true, i wouldn't like to break with this convention. maybe we could provide an extra parameter to `evolve` and the search stops if one of the limits (max gen, max eval) is reached? would you please create a new issue for this, migrating some of this discussion here?

jakobj:
Update: i've managed to fix the broken test. it seems like the `sympy.evaluate` context manager created some problems. hence i've removed it and implemented an equivalent way of avoiding simplification of sympy expressions if this is not desired by the user. this also led me to update the corresponding test. please have another look.

examples/example_fec_caching.py
```diff
@@ -0,0 +1,98 @@
...
+
+    """
+    f = ind.to_func()
+
+    loss = []
+    for x_0 in np.linspace(-2.0, 2.0, 100):
+        y = f([x_0])
+        loss.append((f_target(x_0) - y) ** 2)
+
+    time.sleep(0.25)  # emulate long fitness evaluation
```
HenrikMettler, line 52: Out of curiosity: What is this sleep timer for?

jakobj: this is just to make the objective take a while. if this is not clear  from the comment, maybe it should adapt it?

HenrikMettler: That I understood, what I am wondering is why the objective should take a while?

jakobj: ahaha, sorry! :D sleep is just to make the difference between using the decorator and not using the decorator obvious. without the sleep here would be no big difference as running the objective once takes about as much time as loading the result from disk. caching is only useful for long running fitness evaluations and the sleep tries to mimic this.

examples/example_fec_caching.py
```diff
@@ -0,0 +1,99 @@
+"""
+Example demonstrating the use of the caching decorator with functional equivalance checking
+===========================================================================================
+
+Caches the results of fitness evaluations in a pickle file
+('example_fec_caching_cache.pkl'). To illustrate its practical use,
```
mschmidt87, line 6: Minor: Fix verbatim formatting for file name.

mschmidt87:
```suggestion
(``example_fec_caching_cache.pkl``). To illustrate its practical use,
```

examples/example_fec_caching.py
```diff
@@ -0,0 +1,99 @@
...
+
+# %%
+# We then define the objective function for the evolutionary
+# algorithm: It consists of an inner objective which we wrap with the
+# caching decorator. This decorator specifies a pickle file that will be used for
+# caching results of fitness evaluations. The inner objective is then used by the objective
+# function to compute (or retrieve from cache) the fitness of the individual.
+
+
+@cgp.utils.disk_cache("example_fec_caching_cache.pkl", use_fec=True, fec_seed=12345)
```
mschmidt87, line 35: I think it would be instructive to use the other keyword arguments of fec caching here, too.

jakobj:
thanks for the suggestions, please have another look!

mschmidt87:
Approved. Please squash before merging. Let's ignore coveralls, I don't understand what it is doing, doesn't make sense to me.

jakobj:
yes, good idea. included `MultiGenome` and rearranged the commits to make more sense, please have a final(?) look ;)

PR #209, opened 2020-07-28
Move metadata into __init__.py

mschmidt87:
This PR introduces a few, small changes to the way we handle metadata:
- Read metadata in setup.py from __init__.py
- Remove __version__.py
- Read metadata in docs from package

This makes the handling of metadata more transparent, and exposes the metadata to the user at runtime.

I also bump up the version to `0.2.0dev` to avoid confusion with the release version. I think that in future we should always do that immediately after a release.


setup.py
```diff
@@ -8,11 +8,17 @@ def _cut_version_number_from_requirement(req):
     return req.split()[0]
 
 
-def read_version():
-    with open("./cgp/__version__.py") as f:
-        line = f.read()
-        match = re.findall(r"[0-9]+\.[0-9]+\.[0-9]+", line)
-        return match[0]
+def read_meta(meta):
```
jakobj, line 11:
```suggestion
def read_metadata(metadata_str):
```

setup.py
```diff
@@ -8,11 +8,17 @@ def _cut_version_number_from_requirement(req):
...
+    """
+    Find __"meta"__ from init file.
```
jakobj, line 13:
```suggestion
    Find __"meta"__ in init file.
```
or "Read __"meta"__ from init file", don't have a strong preference

setup.py
```diff
@@ -57,12 +63,13 @@ def read_long_description():
...
 setup(
     name="hal-cgp",
-    version=read_version(),
-    maintainer="Jakob Jordan, Maximilian Schmidt",
-    description=("Cartesian Genetic Programming in pure Python."),
+    version=read_meta("version"),
+    maintainer=read_meta("maintainer"),
+    author=read_meta("author"),
+    description=(read_meta("description")),
     license="GPLv3",
```
jakobj, line 70:
```suggestion
    license=read_meta("license"),
```

mschmidt87:
> so sexy heart_eyes where do you come up with this stuff?

I'm watching too many PyCon videos :-) This PR was inspired by this one: https://www.youtube.com/watch?v=9G2s1TN9QQY

PR #212, opened 2020-07-29
Use logical and, and reuse existing functions in helper functions of `Genome`

jakobj:
…


jakobj:
thanks for the quick review :ok_hand: and merging

PR #213, opened 2020-07-29
Fix parameters being reininitialized in Genome.mutate()

jakobj:
As `mutate` uses `dna.setter` it also triggers the function to initialize numerical values for (unknown) parameters. Before initializing any parameter it is now checked that this does have a value assigned yet. If it does, this parameter is not (re)initialized.

(ouch! :grimacing: )


jakobj:
my guess would be that the parent is maintained long enough to evolve such useful parameters that new offspring with reset parameters can not compete. as the parent is not mutated, local search will fit it's parameters correctly.

jakobj:
rebased, waiting for travis

Issue #208, opened 2020-07-28
CI takes too long because of examples


mschmidt87:
The CI currently already takes quite a long time (~15 minutes for the longest build) which is mainly due to the examples, see the start of the discussion in #205 . This will become even worse when we add more and more examples. 

There are three possible options:

### Manual testing of examples
Instead of running examples in every execution of the CI suite, we trigger the execution of examples only just before merging a PR. Unfortunately, this would require us to manually trigger this (with some mechanism) because travis (probably) does not support any automatic mechanism (as far as I know).

### Only run examples when deploying the docs (after PR merge)
Only run the examples *after* merging the PR when we deploy the full docs. Obiously, this begs the risk that we'll have to create PRs that fix examples after they have been broken on master.

### Run examples in CI only for a minimum number of generations
Instead of running each example completely, just run it for a few generations. Essentially, we are currently just testing whether the example scripts do not crash, but we're not testing whether they converge to the desired result. I think that this is fine, but if we anyway don't check for convergence, there is actually no point to run them for the full number of generations. Instead, we could reduce the number of generations just for the sake of testing them in CI, which will dramatically reduce runtime, but still enables us to catch bugs in the script (at least most of them).
The downside is that we need to introduce some mechanism in the scripts to reduce the number of generations. The least invasive way I can come up with is to add something like
```python
if os.environ['NUM_GENERATIONS']:
    num_generations = os.environ['NUM_GENERATIONS']
evolve_params = {'max_generations': num_generations ...}
...
```
and then set the env variable in CI.

jakobj:
i like your last suggestion, but how about reading a command line argument (with a default value) instead of an environment variable? that seems to me the most natural way, which also the user could intuitively control

mschmidt87:
Okay, that's actually better. I came up with the env variable option, because it is more subtle, but a proper command line argument is better.

mschmidt87:
So, I thought about this and it seems there are two options for the third suggestion, both not ideal:
- use command-line arguments via `argparse`: Definitely the better and cleaner interface for users. However, as far as I understand, `sphinx-gallery` does not have the option to parse command line arguments to example scripts (I created an issue in their repo to ask about this: https://github.com/sphinx-gallery/sphinx-gallery/issues/731)
- use the option with an env variable. Would solve the problem with sphinx-gallery, but it is not a good API.

[...]

PR #218, opened 2020-07-31
Add command line args for example scripts

mschmidt87:
This PR implements command-line args for the example scripts. It is the first step to address #208 . In a follow-up PR, I will then make use of these command-line args in the travis CI.

The following things are implemented:
- Add option to specify max_generations on the command line to example scripts
- Add switch for visualization in mountain_car example

### DIFF
--- a/examples/example_differential_evo_regression.py
+++ b/examples/example_differential_evo_regression.py
@@ -20,14 +20,29 @@ References:
 
 """
 
+# The docopt str is added explicitly to ensure compatibility with
+# sphinx-gallery.
+docopt_str = """
+  Usage:
+    example_differential_evo_regression.py [--max-generations=<N>]
+
+  Options:
+    -h --help
+    --max-generations=<N>  Maximum number of generations [default: 2000]
+
+"""
+
 import functools
 
 import matplotlib.pyplot as plt
 import numpy as np
 import scipy.constants
 import torch
+from docopt import docopt
 
 import cgp
+
+args = docopt(docopt_str)
 
 # %%
 # We first define the target function. Note that this function contains
@@ -92,7 +107,7 @@ genome_params = {
 
 ea_params = {"n_offsprings": 4, "tournament_size": 1, "n_processes": 1, "k_local_search": 2}
 
-evolve_params = {"max_generations": 2000, "min_fitness": 0.0}
+evolve_params = {"max_generations": int(args["--max-generations"]), "min_fitness": 0.0}
 
 # use an uneven number of gradient steps so they can not easily
 # average out for clipped values
--- a/examples/example_evo_regression.py
+++ b/examples/example_evo_regression.py
@@ -6,6 +6,16 @@ Example demonstrating the use of Cartesian genetic programming for
 two regression tasks.
 """
 
+# The docopt str is added explicitly to ensure compatibility with
+# sphinx-gallery.
+docopt_str = """
+   Usage:
+     example_evo_regression.py [--max-generations=<N>]
+
+   Options:
+     -h --help
+     --max-generations=<N>  Maximum number of generations [default: 1000]
+"""
 
 import functools
 import warnings
@@ -13,8 +23,11 @@ import warnings
 import matplotlib.pyplot as plt
 import numpy as np
 import scipy.constants
+from docopt import docopt
 
 import cgp
+
+args = docopt(docopt_str)
 
 # %%
 # We first define target functions. For illustration purposes, we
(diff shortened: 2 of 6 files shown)


examples/example_differential_evo_regression.py
```diff
@@ -28,6 +29,16 @@
 
 import cgp
 
+# %%
```
jakobj, line 32: mandatory comment on [docopt](http://docopt.org/) being much cooler than argparse ;)

examples/example_mountain_car.py
```diff
@@ -29,6 +29,26 @@
...
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "--max-generations", type=int, help="max. number of generations in the evolution", default=1500
+)
+visualize_parser = parser.add_mutually_exclusive_group(required=False)
+visualize_parser.add_argument(
+    "--visualize-final-champion", dest="visualize_final_champion", action="store_true"
+)
+visualize_parser.add_argument(
```
jakobj, line 46: i don't understand why you need this option. isn't it the case that if `--visualize-final-champion` is provided it's true, if it's not provided it's false?

mschmidt87: solved with replacing by `docopt`

examples/example_differential_evo_regression.py
mschmidt87: Yes, I agree, `docopt` is much nicer and shorter. However, strangely, `docopt` itself is basically not maintained anymore. See this issue: https://github.com/docopt/docopt/issues/371 with some strange replies by the maintainer (?) who "just doesn't have the time to reject all the PRs" ...

Therefore I would prefer to use [`docopt-ng`](https://github.com/bazaar-projects/docopt-ng), a maintained fork.

Two things are suboptimal in my opinion:
1. docopt does not infer the types of arguments? That is why I need to explicitly convert the value from `str` to `int`. Seems strange that they don't offer that.
2. They also just take over the argument as is, i.e. `--max_generations` is not turned into `max_generations` (what `argparse` is offering). That is also a bit strange.

Perhaps I am just not understanding how to it, though. Do you have an idea?

jakobj: > Two things are suboptimal in my opinion:
> 
>     1. docopt does not infer the types of arguments? That is why I need to explicitly convert the value from `str` to `int`. Seems strange that they don't offer that.
> 
>     2. They also just take over the argument as is, i.e. `--max_generations` is not turned into `max_generations` (what `argparse` is offering). That is also a bit strange.
> 
> 
> Perhaps I am just not understanding how to it, though. Do you have an idea?

i think you are right in both cases. you need to manually convert the arguments from str and they keep their name from the docstring. i would argue that both are not an issue tough.

examples/example_evo_regression.py
```diff
@@ -15,6 +23,8 @@
 
 import cgp
 
+args = docopt(__doc__, more_magic=True)
+breakpoint()
```
jakobj, line 27: pls remove

examples/example_differential_evo_regression.py
```diff
@@ -18,8 +18,16 @@
   programming. In European Conference on Genetic Programming
   (pp. 35-51). Springer, Cham.
 
+Usage: 
+  example_differential_evo_regression.py [--max_generations=<N>]
```
jakobj, line 22: i would prefer command line argument to be dash-separated like this `--max-generations`

examples/example_evo_regression.py
mschmidt87: thanks, done.

examples/example_differential_evo_regression.py
mschmidt87: done.

examples/example_differential_evo_regression.py
mschmidt87: Yeah, not an issue, but quite ugly. I also don't understand, why docopt doesn't infer the type from the default value (or optional documentation).

mschmidt87: This used to be the case, but was then silently removed: https://www.google.com/search?hl=en&q=docopt%20infer%20type

mschmidt87: Anyway, it is what it is.

mschmidt87:
I resolved the issues and already squashed. Please take a look @jakobj

mschmidt87:
Turns out that using `docopt` creates problems with sphinx-gallery because sphinx-gallery "steals" the first docstring of the scripts to render it in html and then docopt can no longer consume it to create the CLI. (see https://github.com/sphinx-gallery/sphinx-gallery/issues/341 for related discussion). 
Therefore I refactored the code slightly. It is not so nice anymore, but I think this it the only way to make `docopt` work with `sphinx-gallery`.

PR #219, opened 2020-07-31
Maint/isort

mschmidt87:
This PR introduces [isort](https://github.com/timothycrosley/isort) to our tools ensuring common coding standards and uniform layout across the library.

`isort` sorts import statements in a particular, reproducible order, namely like this: `FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER`

This PR:
- applies `isort` to all files in the library, testsuite, and examples. Therefore there are a lot of changes in many files, but they are just about formatting, mostly to the imports. Some changes are introduced by `black` as well (trailing commas in function definitions). I am not sure why this happens.
- adds a config file for `isort`
- adds `isort` to the required `dev` libraries
- adds `isort` as an additional check to the CI.
- adds my pre-commit configuration file to the repository. This will make it easier for other contributors to install the pre-commit hooks using that config file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,16 @@
+repos:
+-   repo: https://github.com/ambv/black
+    rev: stable
+    hooks:
+    - id: black
+      language_version: python3.8
+      args: [--config=pyproject.toml]
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v1.2.3
+    hooks:
+    - id: flake8
+      args: [--config=.flake8]
+-   repo: https://github.com/pre-commit/mirrors-isort
+    rev: v4.3.21
+    hooks:
+    - id: isort
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
+  - isort --check-only cgp examples test
   - if [ "$DEP" = "[all]" -a $TRAVIS_PYTHON_VERSION = 3.8 ]; then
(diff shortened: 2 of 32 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+repos:
+-   repo: https://github.com/ambv/black
+    rev: 19.3b0
```
jakobj, line 3: how will we go about maintaining these versions? would it be possible to put a "most recent" here?

.travis.yml
```diff
@@ -30,6 +30,7 @@ script:
   - black --check .
   - flake8 --config=.flake8 .
   - mypy cgp
+  - isort --check-only cgp
```
jakobj, line 33: should we check examples and tests as well? why restrict this to the library path?

.pre-commit-config.yaml
mschmidt87: we can put a stable for black but there is no stable tag for the other tools. Honestly, I think it is not that relevant to care about the versions here.

.travis.yml
mschmidt87: yes, totally, thanks for noticing, I added that.

.pre-commit-config.yaml
jakobj: ok

jakobj:
before we merge, please squash the commits

mschmidt87:
Thanks, I organized this now into 5 commits, which makes sense, in my opinion. Objections, @jakobj ?

jakobj:
i'm happy! merging :)

Issue #174, opened 2020-07-03
Extend documentation to cover more aspects of the API and examples


mschmidt87:
Our documentation in the readme file is very basic at the moment. We should add more aspects of the API and describe the examples. Let's collect a list of additions here:

- [x] Describe examples
- [x] Better description of local search (incl. new option to restrict it to k individuals)
- [x] Describe difference between `IndividualSingleGenome` and `IndividualMultiGenome`
- [x] Add example using `IndividualMultiGenome`
- [x]  include an example of adding a custom node
- [x] Doc page about nodes
- [x]  include an example of choosing initial values for ParameterNode

We should also discuss whether we want to switch from README-only documentation to a more advanced documentation, e.g. on readthedocs.org.

jakobj:
it's hard for me to estimate how much work goes into setting up and maintaining something like readthedocs. if most of those docs is automatically extracted from the code and updated, e.g., upon every merged PR, then I'm all for it. if it requires a lot of manual maintenance i would prefer to just constrain ourselves to the README to avoid "dead" documentation. only well maintained docs are good docs.

regarding extending the README:
- include an example of adding a custom node
- include an example of choosing initial values for `ParameterNode`

mschmidt87:
I think that the cost of maintaining separate documentation is not much larger than maintaining the README. However, for simplificity I would prefer to stick to the README at the moment. Thanks for the suggestions, I added them to the list.

mschmidt87:
Via #181 , I changed my mind and think that we should have a proper documentation using sphinx. In #181 , I have set this up in a way that allows for maximal reuse of documentation from the README. After #181  is merged, I can start creating PRs to solve this issue.

mschmidt87:
#188 adds the examples to the doc, but there are no figures from the examples in the docs. `sphinx-gallery` supports executing the scripts when building the doc and then automatically finds the figures generated from the scripts which can then be included in the documentation and as a thumbnail in the overview page of the gallery. However, in our case, executing the examples takes quite a long time, so I would refrain from that. Thus, we would need to add the figures to the repo or (and that's the best option, I think), use the figures that we anyway create in the CI. Once we host the documentation on a website, we'll generate the docs from CI and then can probably integrate that.

jakobj:
the "examples/example_parameterized_nodes.py" script demonstrates how to define a custom node and how to initialize the parameters. hence i ticked those boxes.

mschmidt87:
[...]

PR #220, opened 2020-07-31
Add page on local search to documentation

mschmidt87:
This PR adds a page to the documentation to describe local search in our library.

I also added a sketch that compares normal evolution to evolution + local search. The figure can certainly be optimized, with regards to visual quality, but more importantly, I think it is a bit misleading that in the evolution we go from `f,g,h` --> `f',g',h' --> `f',g',h'` and then to `f,g` in the next generation. But I am not sure how to solve this, perhaps you have a good idea, @jakobj .

Otherwise, please review the text, it is quite minimal. Feel free to make suggestions.

Part of #174 .

### DIFF
--- a/docs/documentation/documentation.rst
+++ b/docs/documentation/documentation.rst
@@ -1,0 +1,7 @@
+=============
+Documentation
+=============
+.. toctree::
+   :maxdepth: 2
+
+   local_search.rst
--- a/docs/documentation/local_search.rst
+++ b/docs/documentation/local_search.rst
@@ -1,0 +1,30 @@
+==========================================================
+Searching numerical values for parameters via local search
+==========================================================
+
+hal-cgp supports the use of numerical values in expression via nodes with input-independent output in order to represent expressions like `f(x) = x + c` with some fixed value `c`.
+
+While the `ConstantFloat` node has a fixed output value, the output value of a `Parameter` node is stored per individual and can be modified during evolution via local search. These parameters can be different for each position in the genome. One can define custom nodes which contain an arbitrary number of parameters (see :ref:`sphx_glr_auto_examples_example_parametrized_nodes.py`)
+
+To find suitable values for these parameters we perform a numerical optimization for the parameters of the `k` best individuals of each generation. Since this procedure is acting on each individual separately it is referred to as "local search". After this optimization the fitness of each modified individual is reevaluated taking into account the adapted parameter values, and finally used for selection in the evolutionary algorithm.
+
+.. image:: local_search.svg
+
+The library provides a local search function (:meth:`cgp.local_search.gradient_based`) that performs stochastic gradient descent on the values of `Parameter` nodes (cf. Topchy & Punch, 2001; Izzo et al., 2017). Note that this method requires the definition of a differentiable loss function 
+
+--------
+Examples
+--------
+
+See :ref:`sphx_glr_auto_examples_example_differential_evo_regression.py` for an example evolving an expression containing adjustable parameters.
+
+See :ref:`sphx_glr_auto_examples_example_parametrized_nodes.py` for an example evolving an expression containing operator nodes with adjustable parameters.
+
+
+----------
+References
+----------
+
+Izzo, D., Biscani, F., & Mereta, A. (2017). Differentiable Genetic Programming. In European Conference on Genetic Programming (pp. 35-51). Springer, Cham.
+
+Topchy, A., & Punch, W. F. (2001). Faster Genetic Programming based on Local Gradient Search of Numeric Leaf Values. In Proceedings of the Genetic and Evolutionary Computation Conference (GECCO-2001) (Vol. 155162). Morgan Kaufmann San Francisco, CA, USA.
--- a/docs/documentation/local_search.svg
+++ b/docs/documentation/local_search.svg
@@ -1,0 +1,4829 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:osb="http://www.openswatchbook.org/uri/2009/osb"
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   sodipodi:docname="local_search.svg"
+   inkscape:version="0.92.1 r15371"
+   id="svg8"
+   version="1.1"
+   viewBox="0 0 228.65956 63.300324"
+   height="63.300323mm"
+   width="228.65956mm">
+  <sodipodi:namedview
+     inkscape:window-maximized="1"
+     inkscape:window-y="27"
+     inkscape:window-x="72"
+     inkscape:window-height="1016"
+     inkscape:window-width="1848"
+     showgrid="false"
+     inkscape:current-layer="layer1"
+     inkscape:document-units="mm"
+     inkscape:cy="119.62266"
+     inkscape:cx="432.11256"
+     inkscape:zoom="1.9416237"
+     inkscape:pageshadow="2"
+     inkscape:pageopacity="0.0"
+     borderopacity="1.0"
+     bordercolor="#666666"
+     pagecolor="#ffffff"
+     id="base"
+     showguides="false"
+     inkscape:snap-global="false"
+     fit-margin-top="0"
+     fit-margin-left="0"
+     fit-margin-right="0"
+     fit-margin-bottom="0" />
+  <defs
+     id="defs2">
+    <marker
+       inkscape:stockid="StopL"
+       orient="auto"
+       refY="0"
+       refX="0"
+       id="marker5507"
+       style="overflow:visible"
+       inkscape:isstock="true">
+      <path
+         id="path5505"
+         d="M 0,5.65 V -5.65"
+         style="fill:none;fill-opacity:0.75;fill-rule:evenodd;stroke:#cdedff;stroke-width:1.00000003pt;stroke-opacity:1"
+         transform="scale(0.8)"
+         inkscape:connector-curvature="0" />
+    </marker>
+    <inkscape:path-effect
+       effect="bspline"
+       id="path-effect5503"
+       is_visible="true"
+       weight="33.333333"
+       steps="2"
+       helper_size="0"
+       apply_no_weight="true"
+       apply_with_weight="true"
+       only_selected="false" />
+    <inkscape:path-effect
+       effect="bspline"
+       id="path-effect6404"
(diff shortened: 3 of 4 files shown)


docs/details/local_search.rst
```diff
@@ -0,0 +1,30 @@
...
+
+hal-cgp supports the use of numerical values in expression via constant nodes and parameter nodes in order to represent expressions like `f(x) = x + c` with some fixed value `c`.
+
+While the `ConstantFloat` node has a fixed output value, the output value of a `Parameter` node is stored per individual and can be modified during evolution via local search.
+
+Local search means that in every generation, we add an additional step where we perform a local optimization of each individual to obtain a modified invidivual. We then evaluate the fitness of the modified individual for selection in the evolutionary algorithm.
+
+.. image:: local_search.svg
+
+The library provides a local search function (:meth:`cgp.local_search.gradient_based`) that performs stochastic gradient descent on the values of `Parameter` nodes (cf. Topchy & Punch, 2001; Izzo et al., 2017).
```
jakobj, line 13:
```suggestion
The library provides a local search function (:meth:`cgp.local_search.gradient_based`) that performs stochastic gradient descent on the values of `Parameter` nodes (cf. Topchy & Punch, 2001; Izzo et al., 2017). Note that this method requires the definition of a differentiable loss function 
```

docs/details/local_search.rst
```diff
@@ -0,0 +1,30 @@
...
+
+--------
+Examples
+--------
+
+See :ref:`sphx_glr_auto_examples_example_differential_evo_regression.py` for an example evolving an expression containing learnable parameters.
```
jakobj, line 19:
```suggestion
See :ref:`sphx_glr_auto_examples_example_differential_evo_regression.py` for an example evolving an expression with numerical constants.
```

docs/details/local_search.rst
```diff
@@ -0,0 +1,30 @@
...
+
+See :ref:`sphx_glr_auto_examples_example_parametrized_nodes.py` for an example evolving an expression containing operator nodes with learnable parameters.
```
jakobj, line 21:
```suggestion
See :ref:`sphx_glr_auto_examples_example_parametrized_nodes.py` for an example evolving an expression with numerical constants using custom defined operator node with parameters.
```

jakobj:
> I also added a sketch that compares normal evolution to evolution + local search. The figure can certainly be optimized, with regards to visual quality, but more importantly, I think it is a bit misleading that in the evolution we go from `f,g,h` --> `f',g',h' --> `f',g',h'`and then to`f,g` in the next generation. But I am not sure how to solve this, perhaps you have a good idea, @jakobj .

what about adding a generation label to the individuals? i agree that otherwise it is indeed a bit misleading since it either looks like the offsprings to not inherit their parent parameters or that the parameters are not transferred to the next generation. also i would add the (possible) effect that instead of `f,g` surviving, after local search `f,h` survive.

docs/details/local_search.rst
mschmidt87: I find constants here a bit misleading. "Constants" could be interpreted as if they are constant throughout the evolution.

docs/details/local_search.rst
mschmidt87: same here.

mschmidt87:
> > I also added a sketch that compares normal evolution to evolution + local search. The figure can certainly be optimized, with regards to visual quality, but more importantly, I think it is a bit misleading that in the evolution we go from `f,g,h` --> `f',g',h' --> `f',g',h'`and then to`f,g` in the next generation. But I am not sure how to solve this, perhaps you have a good idea, @jakobj .
> 
> what about adding a generation label to the individuals? i agree that otherwise it is indeed a bit misleading since it either looks like the offsprings to not inherit their parent parameters or that the parameters are not transferred to the next generation. also i would add the (possible) effect that instead of `f,g` surviving, after local search `f,h` survive.

Adding generation labels doesn't really solve the problem, right? I mean, what label would you put on the individuals in each of the three circles then?

docs/details/local_search.rst
jakobj: well, i actually meant constants here bc i though "an expression" relates to the target function :D maybe something like "free parameters"? i just don't like "learnable" in this context, my prior is too strong ;)

jakobj:
> Adding generation labels doesn't really solve the problem, right? I mean, what label would you put on the individuals in each of the three circles then? The only possibility I see is to add a fourth circle for the offspring population in the next generation which then contains f', g' (or h') + a new offspring.

i had almost the same in mind, except i wouldn't introduce an additional circle: you just move f,g to the bottom of the circle, add a subscript "g" and at the top of the circle you add the individuals from the next generation, e.g., f', g' with subscript "g+1". maybe you can additionaly include a horizontal line in this circle to separate the individuals of the two generations?

docs/details/local_search.rst
mschmidt87: Yes, I know what you mean but in a way the term "constant" is overloaded in this context:
- constant across the evolution
- constant = independent from the input to the computational graph.

So I'd rather not use that word here. Why not "learnable"?

jakobj: sorry, no, the term is _not_ overloaded with regards to the target function. that is a fixed function with some numerical values and those are generally referred to as constants. but i understand that it's more tricky if we talk about "the things in the expressions represented by individuals that are adapted by local search". so let's stick with our nomenclature and call it "parameter".

i would avoid "learnable" since for me this immediately sets the context to plasticity, brain, backprop, and so on. but these parameters can be adapted in whatever way the user wants. random search, grid search, really any continuous optimizer would be suitable and i would find it a stretch to call all of this "learning". what about "adjustable"?

mschmidt87: It is overloaded in the context of this document because just above we are referring to `ConstantNode` which is exactly what we don't mean here.

"Adjustable" is fine for me.

mschmidt87:
I updated the figure, please check, @jakobj .

docs/details/local_search.rst
```diff
@@ -0,0 +1,30 @@
+==========================================================
+Searching numerical values for parameters via local search
+==========================================================
+
+hal-cgp supports the use of numerical values in expression via nodes with input-independent output in order to represent expressions like `f(x) = x + c` with some fixed value `c`.
+
+While the `ConstantFloat` node has a fixed output value, the output value of a `Parameter` node is stored per individual and can be modified during evolution via local search. These parameters can be different for each position in the genome.
```
jakobj, line 7:
```suggestion
While the `ConstantFloat` node has a fixed output value, the output value of a `Parameter` node is stored per individual and can be modified during evolution via local search. These parameters can be different for each position in the genome. One can define custom nodes which contain an arbitrary number of parameters (see `example_parametrized_nodes.py`)
```
I think it would be worth making the link to the example here

docs/details/local_search.rst
```diff
@@ -0,0 +1,30 @@
...
+
+.. image:: local_search.svg
+
+The library provides a local search function (:meth:`cgp.local_search.gradient_based`) that performs stochastic gradient descent on the values of `Parameter` nodes (cf. Topchy & Punch, 2001; Izzo et al., 2017). Note that this method requires the definition of a differentiable loss function 
+
+--------
+Examples
+--------
+
+See :ref:`sphx_glr_auto_examples_example_differential_evo_regression.py` for an example evolving an expression containing learnable parameters.
```
jakobj, line 19:
```suggestion
See :ref:`sphx_glr_auto_examples_example_differential_evo_regression.py` for an example evolving an expression containing adjustable parameters.
```

docs/details/local_search.rst
```diff
@@ -0,0 +1,30 @@
...
+
+See :ref:`sphx_glr_auto_examples_example_parametrized_nodes.py` for an example evolving an expression containing operator nodes with learnable parameters.
```
jakobj, line 21:
```suggestion
See :ref:`sphx_glr_auto_examples_example_parametrized_nodes.py` for an example evolving an expression containing operator nodes with adjustable parameters.
```

mschmidt87:
Thanks, I improved the figure further. If you want the right part perfect, I fear that you'll have to do it, I can't figure it out properly in inkscape.

mschmidt87:
I was also wondering whether "Detailed documentation" is the best title we can come up with. Maybe just "Documentation" or "User documentation"?

jakobj:
> I was also wondering whether "Detailed documentation" is the best title we can come up with. Maybe just "Documentation" or "User documentation"? @jakobj

indeed "detailed" might be a bit misleading, since it's quite high-level. i think we can just go with "Documentation" :ok_hand:

jakobj:
> I was also wondering whether "Detailed documentation" is the best title we can come up with. Maybe just "Documentation" or "User documentation"? @jakobj

it's super pretty and round enough ;) one thing: maybe it would be nice to also add the generation index and $g_{i+1} = g_i$ to the left side of the figure?

mschmidt87:
> > I was also wondering whether "Detailed documentation" is the best title we can come up with. Maybe just "Documentation" or "User documentation"? @jakobj
> 
> it's super pretty and round enough ;) one thing: maybe it would be nice to also add the generation index and $g_{i+1} = g_i$ to the left side of the figure?

Thanks, done.

jakobj:
> > > I was also wondering whether "Detailed documentation" is the best title we can come up with. Maybe just "Documentation" or "User documentation"? @jakobj
> > 
> > 
> > it's super pretty and round enough ;) one thing: maybe it would be nice to also add the generation index and $g_{i+1} = g_i$ to the left side of the figure?
> 
> Thanks, done.

nice! i think it would be nice to also split the parent population on the left like you did on the right

jakobj:
> > I was also wondering whether "Detailed documentation" is the best title we can come up with. Maybe just "Documentation" or "User documentation"? @jakobj
> 
> indeed "detailed" might be a bit misleading, since it's quite high-level. i think we can just go with "Documentation" ok_hand

i think this is a todo that's still left

mschmidt87:
> > > > I was also wondering whether "Detailed documentation" is the best title we can come up with. Maybe just "Documentation" or "User documentation"? @jakobj
> > > 
> > > 
> > > it's super pretty and round enough ;) one thing: maybe it would be nice to also add the generation index and $g_{i+1} = g_i$ to the left side of the figure?
> > 
> > 
> > Thanks, done.
> 
> nice! i think it would be nice to also split the parent population on the left like you did on the right

Hmm, but since f_{i+1} = f_i, that's kind of pointless, isn't it? I'd say not splitting it underlines that point.

mschmidt87:
> > > I was also wondering whether "Detailed documentation" is the best title we can come up with. Maybe just "Documentation" or "User documentation"? @jakobj
> > 
> > 
> > indeed "detailed" might be a bit misleading, since it's quite high-level. i think we can just go with "Documentation" ok_hand
> 
> i think this is a todo that's still left

Thanks, done.

jakobj:
> > > > > I was also wondering whether "Detailed documentation" is the best title we can come up with. Maybe just "Documentation" or "User documentation"? @jakobj
> > > > 
> > > > 
> > > > it's super pretty and round enough ;) one thing: maybe it would be nice to also add the generation index and $g_{i+1} = g_i$ to the left side of the figure?
> > > 
> > > 
> > > Thanks, done.
> > 
> > 
> > nice! i think it would be nice to also split the parent population on the left like you did on the right
> 
> Hmm, but since f_{i+1} = f_i, that's kind of pointless, isn't it? I'd say not splitting it underlines that point.

ok, i don't have a strong feeling about this besides symmetry.

PR #222, opened 2020-07-31
Make sure user-set attributes of individuals are copied to offspring

jakobj:
In some cases it would be nice to record some historical information within the objective function. The easiest approach would be to create a objective-function closure, similar to the history callback, and let the objective directly write to the corresponding recording object. This works well will a single process, but since parallel processes are running in different interpreter instances, it does not work with multiprocessing (at least i didn't get it to work).

Here I'm making a suggestion based on discussions with @HenrikMettler: the user can set (new) attributes of individuals within the objective which could then be subsequently recorded outside the parallel context, similar to recording, e.g., fitness values. To make this work we need to make sure offspring inherit all these custom attributes from their parents, since some offspring are never evaluated (for example due to mutation in silent regions) and hence never have this attribute set.

Please regard this only as a suggestion, I'm happy to hear about other solutions!

### DIFF
--- a/cgp/individual.py
+++ b/cgp/individual.py
@@ -1,4 +1,5 @@
-from typing import Callable, List, Union
+import copy
+from typing import Callable, List, Set, Type, Union
 
 import numpy as np
 
@@ -25,6 +26,8 @@ class IndividualBase:
     """Base class for all individuals.
     """
 
+    __base_attrs__: Set[str]
+
     def __init__(self, fitness: Union[float, None]) -> None:
         """Init function.
 
@@ -32,11 +35,25 @@ class IndividualBase:
             Fitness of the individual.
         """
         self.fitness: Union[float, None] = fitness
-        self.idx: int
-        self.parent_idx: int
+        self.idx: Union[int, None] = None
+        self.parent_idx: Union[int, None] = None
+
+    def __init_subclass__(cls: Type) -> None:
+
+        # store the attributes present right after instance creation,
+        # i.e., all attributes not set by the user
+        cls.__base_attrs__ = set(cls(None, None).__dict__.keys())
 
     def clone(self):
         raise NotImplementedError()
+
+    def _copy_user_defined_attributes(self, other):
+        """Copy all attributes that are not defined in __init__ of the (sub
+        and super) class from self to other.
+        """
+        for attr in self.__dict__:
+            if attr not in self.__base_attrs__:
+                setattr(other, attr, copy.deepcopy(getattr(self, attr)))
 
     def mutate(self, mutation_rate, rng):
         raise NotImplementedError()
@@ -109,6 +126,7 @@ class IndividualSingleGenome(IndividualBase):
     def clone(self) -> "IndividualSingleGenome":
         ind = IndividualSingleGenome(self.fitness, self.genome.clone())
         ind.parent_idx = self.idx
+        self._copy_user_defined_attributes(ind)
         return ind
 
     def mutate(self, mutation_rate: float, rng: np.random.RandomState) -> None:
@@ -155,6 +173,7 @@ class IndividualMultiGenome(IndividualBase):
     def clone(self) -> "IndividualMultiGenome":
         ind = IndividualMultiGenome(self.fitness, [g.clone() for g in self.genome])
         ind.parent_idx = self.idx
+        self._copy_user_defined_attributes(ind)
         return ind
 
     def mutate(self, mutation_rate: float, rng: np.random.RandomState) -> None:
--- a/docs/documentation/population_individual.rst
+++ b/docs/documentation/population_individual.rst
@@ -35,6 +35,21 @@ There are two types of individuals:
 These classes are thin wrappers around the `Genome` (see :meth:`cgp.Genome`) and
 the `CartesianGraph` (see :meth:`cgp.CartesianGraph`) classes.
 
+Any custom properties of individuals that are set within the objective function are copied to their offspring.
+This, for example, allows recording of variables used during fitness evaluation:
+
+.. code-block:: python
+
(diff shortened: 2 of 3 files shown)


cgp/individual.py
```diff
@@ -31,12 +32,17 @@ def __init__(self, fitness: Union[float, None]) -> None:
...
-        self.idx: int
-        self.parent_idx: int
+        self.idx: Union[int, None] = None
+        self.parent_idx: Union[int, None] = None
 
     def clone(self):
         raise NotImplementedError()
 
+    def _copy_user_defined_attributes(self, other):
+        for attr in self.__dict__:
```
mschmidt87, line 42: If you copy everything that is in `self.__dict__`, then these attributes are not really "user-defined", are they?

jakobj: yeah, maybe there's a comment missing here. i iterate through `__dict__` but then exclude all members that would be present in a "clean" instance in the line below. so "user defined" means all members that are not defined at `__init__`.

mschmidt87: Yes, please add a comment. Is this the "proper" way to achieve this kind of behavior? It looks a bit "hacky" to me.

jakobj: hmm, i'm not sure what you mean by "proper", but it seems like a reasonable approach to me ;)

jakobj:
> Looks good to me. Just wondering if one ever wants to be able to clone without copying user defined attributes, so have a function `clone_without_user_attributes` (or something alike). But personally I don't think this is necessary

yes, it's a fundamental flaw of this approach that this is not easily possible. however, i would suggest to tackle this problem when it becomes an issue, not earlier ;)

cgp/individual.py
mschmidt87: is `self.__class__(None, None)` creating a new instance? If yes, you're creating a new instance in every function call? Isn't that a bit wasteful?

It seems that this feature could be better solved in another way with a member function that is executed by a callback that the user activates by some optional arguments in the high-level API?

jakobj: i agree, it's a bit wasteful, but clone is not called too often so i don't think this would be a big issue.

yes, you might be right. could you expand a bit on your suggestion? i will look more into multiprocessing since i'm not sure how easy it is to get variables back out of the parallel context. it is possible that this could only be done via return values.

mschmidt87: What I mean is that the `ea.MuPlusLambda._create_new_offspring_generation()` function could accept a callback that is called just after cloning: https://github.com/Happy-Algorithms-League/hal-cgp/blob/7932565d5112135a3c538d4ff6b948ba3020eb1b/cgp/ea/mu_plus_lambda.py#L136

This callback then sets attributes in the newly cloned individuals.
Alternatively, to be more restrictive: Just accept a set of attribute names as an arguments to the function which are then copied in the function from the parents to the cloned offspring.

jakobj: ok, made a suggestion, please have a look :)

cgp/individual.py
```diff
@@ -106,9 +107,14 @@ def __init__(self, fitness: Union[float, None], genome: Genome) -> None:
     def __repr__(self):
         return f"Individual(idx={self.idx}, fitness={self.fitness}, genome={self.genome}))"
 
-    def clone(self) -> "IndividualSingleGenome":
+    def clone(
+        self,
+        *,
```
mschmidt87, line 112: What does this `*` argument do here?

jakobj: this is just the notation for keyword-only arguments (https://www.python.org/dev/peps/pep-3102/). everything behind `*` needs to be passed like `argument_name=value`

jakobj:
yes, an example is a good idea. this will also allow us to evaluate whether we are really happy with this API, currently i'm still not certain.

jakobj:
i've added an example now. i have to say i don't see the benefit of forcing the user to define this custom callback, why not just copy all attributes to the offspring?

mschmidt87:
What about using your original solution but we define a `__base__attr__` member dict of the `Individual` class that stores the basic elementers and can be used instead of creating new instance every time.

Then you could do:
```python
        for attr in parent.__dict__:
            if attr not in parent.__base_attrs__:
                setattr(offspring, attr, copy.deepcopy(getattr(parent, attr)))
```

jakobj:
i've implemented your suggestion @mschmidt87, please have another look!

mschmidt87 (changes requested):
The implementation looks good to me.
However, adding an entire example that is basically identical with the regression example except for this one custom attribute is a bit of an overkill. What about adding some documentation of this feature to the documentation page about individuals, popuations etc., with some example code?

docs/documentation/population_individual.rst
```diff
@@ -35,6 +35,23 @@ There are two types of individuals:
 These classes are thin wrappers around the `Genome` (see :meth:`cgp.Genome`) and
 the `CartesianGraph` (see :meth:`cgp.CartesianGraph`) classes.
 
+Any custom properties of individuals that are set within the objective function are copied to their offspring.
+This, for example, allows recording of variables used during fitness evaluation:
+
+.. code-block:: python
+
+  def objective(individual):
+      individual.fitness = ...
```
mschmidt87, line 44: I would suggest to add some value here to obtain valid Python code. This will enable us to use `doctest` down the road to automatically test these snippets.

jakobj: ok, i will change it, but i would argue that it's pointless to test this snippet, isn't it?

mschmidt87 (approved):

mschmidt87:
Looks good to me, please squash before merging.

jakobj:
thanks! squashed and waiting for travis

PR #227, opened 2020-08-04
Add minimal regression example

### LINKED ISSUE #221: Simplify `example_evo_regression`
I keep going back to this example for trying out new stuff and whenever i do, i'm annoyed by it being so unnecessarily complicated. shouldn't we view this example as the "minimal beginner example" where everything is _as simple as possible_? this would mean reducing the example to a single target function, removing the division operator to be able to remove the ugly context manager catching warnings, not using numpy , only using one input node, etc pp. i think this make it much easier to understand for someone looking at the first script using hal-cgp. also i would remove the docstrings in favor of the sphinx documentation.

what do you think @mschmidt87 @HenrikMettler ?

jakobj:
please suggest any further simplification that comes to mind.

closes #221

### DIFF
--- a/examples/example_minimal.py
+++ b/examples/example_minimal.py
@@ -1,0 +1,139 @@
+"""
+Minimal example for evolutionary regression
+===========================================
+
+Example demonstrating the use of Cartesian genetic programming for
+a simple regression task.
+"""
+
+# The docopt str is added explicitly to ensure compatibility with
+# sphinx-gallery.
+docopt_str = """
+   Usage:
+     example_minimal.py [--max-generations=<N>]
+
+   Options:
+     -h --help
+     --max-generations=<N>  Maximum number of generations [default: 300]
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+import scipy.constants
+from docopt import docopt
+
+import cgp
+
+args = docopt(docopt_str)
+
+# %%
+# We first define a target function.
+
+
+def f_target(x):
+    return x[0] ** 2 + 1.0
+
+
+# %%
+# Then we define the objective function for the evolution. It uses
+# the mean-squared error between the output of the expression
+# represented by a given individual and the target function evaluated
+# on a set of random points.
+
+
+def objective(individual):
+    if individual.fitness is not None:
+        return individual
+
+    n_function_evaluations = 1000
+
+    np.random.seed(1234)
+
+    f = individual.to_func()
+    loss = 0
+    for x in np.random.uniform(-4, 4, n_function_evaluations):
+        # the callable returned from `to_func` accepts and returns
+        # lists; accordingly we need to pack the argument and unpack
+        # the return value
+        y = f([x])[0]
+        loss += (f_target([x]) - y) ** 2
+
+    individual.fitness = -loss / n_function_evaluations
+
+    return individual
+
+
+# %%
+# Next, we set up the evolutionary search. We first define the
+# parameters for the population, the genome of individuals, and the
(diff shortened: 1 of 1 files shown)


examples/example_minimal.py
```diff
@@ -0,0 +1,139 @@
...
+# %%
+# Next, we set up the evolutionary search. It comprises:
+#
+# - defining the parameters for the population, the genome of individuals,
+#   and the evolutionary algorithm.
+# - creating a Population instance and instantiating the evolutionary algorithm.
+# - defining a recording callback closure for bookkeeping of the progression of the evolution.
+#
+# Finally, we call the `evolve` method to perform the evolutionary search.
+
```
mschmidt87, line 74: This docstring could be broken up into more pieces to make this nicer in the sphinx docs.

examples/example_minimal.py
```diff
@@ -0,0 +1,139 @@
...
+    "n_rows": 1,
+    "levels_back": 5,
+    "primitives": (cgp.Add, cgp.Sub, cgp.Mul, cgp.ConstantFloat),
+}
+
+ea_params = {"n_offsprings": 4, "tournament_size": 2, "n_processes": 2}
+
+evolve_params = {"max_generations": int(args["--max-generations"]), "min_fitness": 0.0}
+
+# create population that will be evolved
```
mschmidt87, line 91: Turn into sphinx-compatible docstring.

examples/example_minimal.py
```diff
@@ -0,0 +1,139 @@
...
+pop = cgp.Population(**population_params, genome_params=genome_params)
+
+# create instance of evolutionary algorithm
```
mschmidt87, line 94: same here.

examples/example_minimal.py
```diff
@@ -0,0 +1,139 @@
...
+ea = cgp.ea.MuPlusLambda(**ea_params)
+
+# define callback for recording of fitness over generations
```
mschmidt87, line 97: and here.

examples/example_minimal.py
```diff
@@ -0,0 +1,139 @@
...
+history = {}
+history["fitness_champion"] = []
+
+
+def recording_callback(pop):
+    history["fitness_champion"].append(pop.champion.fitness)
+
+
+# Perform the evolution
```
mschmidt87, line 106: also here.

examples/example_minimal.py
jakobj: hmm, it's rendered as an unordered list, isn't that what we want here?

mschmidt87: What I mean is: Instead of describing all the steps in one block of comment, describe them just before you execute. Let me make some concrete suggestions.

examples/example_minimal.py
```diff
@@ -0,0 +1,144 @@
...
+
+# %%
+# Next, we set up the evolutionary search. It comprises:
+#
+# - defining the parameters for the population, the genome of individuals,
+#   and the evolutionary algorithm.
+# - creating a Population instance and instantiating the evolutionary algorithm.
+# - defining a recording callback closure for bookkeeping of the progression of the evolution.
+#
+# Finally, we call the `evolve` method to perform the evolutionary search.
```
mschmidt87, line 74:
```suggestion
# Next, we set up the evolutionary search. We first define the parameters for the population, the genome of individuals,
#   and the evolutionary algorithm.
```

mschmidt87:
Approved. Please squash before merging.

jakobj:
thanks. done, waiting for travis

PR #230, opened 2020-08-11
Add example for regression with multiple genomes

mschmidt87:
This PR adds an example for multiple regression. It is based on `example_minimal.py` and modifies the target function to be piecewise defined.

### DIFF
--- a/examples/example_multi_genome.py
+++ b/examples/example_multi_genome.py
@@ -1,0 +1,143 @@
+"""
+Example for evolutionary regression with multiple genomes
+=========================================================
+
+Example demonstrating the use of Cartesian genetic programming with multiple
+genomes per individual for a simple regression task with a piecewise
+target function.
+"""
+
+# The docopt str is added explicitly to ensure compatibility with
+# sphinx-gallery.
+docopt_str = """
+   Usage:
+     example_multi_genome.py [--max-generations=<N>]
+
+   Options:
+     -h --help
+     --max-generations=<N>  Maximum number of generations [default: 300]
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+import scipy.constants
+from docopt import docopt
+
+import cgp
+
+args = docopt(docopt_str)
+
+# %%
+# We first define a target function.  The function applies different
+# transformations to the input depending whether the input is less or greater
+# than or equal to zero. We thus need to fit two different functions.
+
+
+def f_target(x):
+    return np.select([x < 0, x >= 0], [-x, x ** 2 + 1.0])
+
+
+# %%
+# Then we define the objective function for the evolution. It uses the
+# mean-squared error between the output of the expression represented by a given
+# individual and the target function evaluated on a set of random points. We
+# here either evaluate the function represented by the first (``f[0]``) or the second
+# genome (``f[1]``), depending whether the input is less or greater than zero.
+
+
+def objective(individual):
+    if individual.fitness is not None:
+        return individual
+
+    n_function_evaluations = 1000
+
+    np.random.seed(1234)
+
+    # Note that f is now a list of functions because individual is an instance
+    # of `InvidividualMultiGenome`
+    f = individual.to_numpy()
+    x = np.random.uniform(-4, 4, (n_function_evaluations, 1))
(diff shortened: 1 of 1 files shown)


examples/example_multi_genome.py
```diff
@@ -0,0 +1,138 @@
+"""
+Example for evolutionary regression with multiple genomes
+=========================================================
+
+Example demonstrating the use of Cartesian genetic programming with multiple
+genomes per individual for a simple regression task with a piecewise-defined
```
jakobj, line 6:
```suggestion
genomes per individual for a simple regression task with a piecewise
```
but this is a matter of taste ;)

examples/example_multi_genome.py
```diff
@@ -0,0 +1,138 @@
...
+target function.
+"""
+
+# The docopt str is added explicitly to ensure compatibility with
+# sphinx-gallery.
+docopt_str = """
+   Usage:
+     example_minimal.py [--max-generations=<N>]
```
jakobj, line 14:
```suggestion
     example_multi_genome.py [--max-generations=<N>]
```

examples/example_multi_genome.py
```diff
@@ -0,0 +1,138 @@
...
+
+def objective(individual):
+    if individual.fitness is not None:
+        return individual
+
+    n_function_evaluations = 1000
+
+    np.random.seed(1234)
+
+    f = individual.to_numpy()
```
jakobj, line 56: maybe add a comment that `f` is now a list? otherwise line 58 is quite confusing. or maybe add an additional comment there?

examples/example_multi_genome.py
```diff
@@ -0,0 +1,138 @@
...
+    y = np.piecewise(x, [x[:, 0] < 0, x[:, 0] >= 0], f)[:, 0]
+    loss = np.sum((f_target(x[:, 0]) - y) ** 2)
+    individual.fitness = -loss / n_function_evaluations
+    return individual
+
+
+# %%
+# Next, we set up the evolutionary search. First, we define the parameters for
+# the population, the genomes of individuals, and the evolutionary
+# algorithm. Note how we define ``genome_params`` as a list of parameter
```
jakobj, line 67:
```suggestion
# algorithm. Note that we define ``genome_params`` as a list of parameter
```

examples/example_multi_genome.py
```diff
@@ -0,0 +1,138 @@
...
+# dictionaries which leads to the populations creating instances of
```
jakobj, line 68:
```suggestion
# dictionaries which causes the population to create instances of
```

examples/example_multi_genome.py
mschmidt87: Good idea, done.

examples/example_multi_genome.py
```diff
@@ -0,0 +1,140 @@
...
+from docopt import docopt
+
+import cgp
+
+args = docopt(docopt_str)
+
+# %%
+# We first define a target function.  The function applies different
+# transformations to the input depending whether the input is less or greater
+# than zero. We thus need to fit two different functions.
```
jakobj, line 33:
```suggestion
# transformations to the input depending whether the input is less than zero or not.
# We thus need to fit two different functions.
```
or some other formulation to include "equal to zero"

examples/example_multi_genome.py
```diff
@@ -0,0 +1,140 @@
...
+    history["fitness_champion"].append(pop.champion.fitness)
+
+
+# %%
+# and finally perform the evolution
+cgp.evolve(pop, objective, ea, **evolve_params, print_progress=True, callback=recording_callback)
+
+
+# %%
+# After finishing the evolution, we plot the result.
```
jakobj, line 113: maybe it would be nice here to print the expressions encoded in the individual?

examples/example_multi_genome.py
mschmidt87: Thanks done.

examples/example_multi_genome.py
mschmidt87: yes, thanks, added it.

Issue #93, opened 2020-05-02
Add more nodes


mschmidt87:
As an extension to the nodes currently implemented, add nodes that perform:
- comparisons:  less than (`<`) and greater than (`>`)
- if/else branches
- binary operators: AND, OR, NOT

jakobj:
i like the idea, in particular for the larger/smaller operations i can think of an application ;)

we need to carefully think about this since i think it's not as easy as just implementing some new nodes. what are the inputs for each operator, what is the output? can we still execute the graph on the fly (`res = graph(<inputs>)`) or do we always need to generate code like in `to_func` etc? i guess it would help to start with a small example where such functionality is necessary, then develop ideas from there

mschmidt87:
One simple application example would be to do regression with a target function that is defined differently for certain intervals, such that one needs if/else checks and "greater/less than" comparison operators.

One difficulty is that potentially the nodes would not necessarily accept each other as inputs any more. For instance, `if/else` should strictly accept only boolean as input. However, we can circumvent that as in Python, i.e. for instance cast integers to True except for 0 value.
The same applies to the binary operators.

jakobj:
I'm not sure about accepting only particular types as inputs. Wouldn't this break with ducktyping? if we nevertheless implement this, this could, for example, be explicitly considered during mutation: only those mutations are allowed that do not make the expression invalid. I'm now wondering how this works for general GP that, for examples generates code instead of mathematical expressions. Is the generated code syntax checked only during fitness evaluation? This seems like a waste, bc i would strongly suggest that most randomly mutated programs are invalid.

jakobj:
is there some progress on this issue @mschmidt87 ? if not i would rather close now and reopen again when this becomes relevant.

mschmidt87:
I haven't had time to work on this, but why closing the issue? It is marked for the future and as an enhancement so I don't see why we cannot just leave it open and, if someone wants to contribute, continue the discussion.

jakobj:
i'd just like to keep this clean by avoiding collecting issues that will not be addressed in the near future, i guess my feeling being that a project should have zero "issues" ;) can we find a different place to keep these "future enhancement ideas"?

mschmidt87:
I would argue that this is what the #thefuture tag is for.

mschmidt87:
[...]

PR #231, opened 2020-08-11
Example with comparison and if/else nodes

HenrikMettler:
Added an example with a composed target function, implements a comparison node and an if/else node. 
Minor change in `_extract_input_names_from_def_output` to only detect unique values, since `x_0, x_1` appear twice in `def_output` of `ComparisonOperator`. 

Note: The current seed finds a correct solution only very slowly. From an earlier version with vector inputs (x,0) I suppose that there should be configurations that work better.


examples/example_composed_target_function.py
```diff
@@ -0,0 +1,160 @@
+"""
+Example for evolutionary regression on a composed target function
```
mschmidt87, line 2: Instead of "composed target function", I would say "piecewise target function". Also in other comments below.

examples/example_composed_target_function.py
```diff
@@ -0,0 +1,160 @@
...
+    _def_output = "1.0 * ((x_0 - x_1)/max(abs(x_0 - x_1), 0.000001))"
+    _def_sympy_output = "1.0 * sign(x_0 - x_1)"
+    _def_torch_output = "1.0 * torch.sign(x_0 - x_1)"
+    _def_numpy_output = "1.0 * np.sign(x_0 - x_1)"
+
+
+def f_target(x):
+    y = np.zeros_like(x)
+    y[x <= 0] = x[x <= 0] ** 2
+    y[x > 0] = x[x > 0]
```
mschmidt87, line 57: In principle, we're solving the same task as in the example that I added in #230 . I would propose to use the same target function as in that example and keep everything as close as possible to that example, to make it easier to compare both examples.

examples/example_composed_target_function.py
```diff
@@ -0,0 +1,160 @@
...
+        else:
+            self._output = inputs[2]
+
+    _def_output = "x_1 if x_0 >= 0 else x_2"
+    _def_numpy_output = "np.piecewise(x_0, [x_0>= 0, x_0<0], [x_1 [x_0>= 0] , x_2[x_0<0]])"
+    _def_sympy_output = "Piecewise((x_1, x_0 >= 0), (x_2, x_0 < 0))"
+    _def_torch_output = "torch.where(x_0 >= 0, x_1, x_2)"
+
+
+class ComparisonOperator(cgp.OperatorNode):
```
mschmidt87, line 43: I would define this operator such that it implements the mathematical symbol `<=`, i.e. it does not output three different output values, but only two. One simple reason for this is that this way we can use the symbol as a string representation when we want to print the expression.
It is also easier to interpret. 
In that case, I would output `0` if the condition fails and `1` if it is true and adapt the IfElseOperator accordingly.

examples/example_composed_target_function.py
```diff
@@ -0,0 +1,160 @@
+"""
```
mschmidt87, line 1: The example is lacking the CLI of the other examples (you can just copy the code) and explanatory docstrings that can be picked up by sphinx.

examples/example_composed_target_function.py
```diff
@@ -0,0 +1,160 @@
...
+    # Perform the evolution
+    cgp.evolve(pop, obj, ea, **evolve_params, print_progress=True, callback=recording_callback)
+    return history, pop.champion
+
+
+if __name__ == "__main__":
+    seed = 1234  # Todo: find better seed
+    rng = np.random.RandomState(seed=seed)
+
+    history, champion = evolution(f_target, rng)
```
mschmidt87, line 159: Can you add a plot here, which can then be rendered in the documentation?
And can you also print the champion's expression in the end?

mschmidt87:
One more thing: when I do `print(champion.to_sympy())` in the end, this is the expression I am getting:
```
In [8]: champion.to_sympy()                                                                          
Out[8]: [Piecewise((1.0*x_0, x_0 >= 0), (x_0**2, True))]
```
This looks a bit strange to me. Why does it say `True` for the second condition?

jakobj:
> One more thing: when I do `print(champion.to_sympy())` in the end, this is the expression I am getting:
> 
> ```
> In [8]: champion.to_sympy()                                                                          
> Out[8]: [Piecewise((1.0*x_0, x_0 >= 0), (x_0**2, True))]
> ```
> 
> This looks a bit strange to me. Why does it say `True` for the second condition?

i think this is due to sympy's simplification combined with the evaluation order: going from left to right it first checks that `x_0` is larger than or equal to zero, and if this condition fails it moves on to the next one. now `x_0` is definitely smaller than zero, so the condition `x_0 < 0` always evaluates to true. but yeah, it's a bit confusing to read

mschmidt87:
This PR will close #93 .

HenrikMettler:
@mschmidt87 Thanks for the feedback & sorry for the delay, I will try to address your points today.

HenrikMettler:
> > One more thing: when I do `print(champion.to_sympy())` in the end, this is the expression I am getting:
> > ```
> > In [8]: champion.to_sympy()                                                                          
> > Out[8]: [Piecewise((1.0*x_0, x_0 >= 0), (x_0**2, True))]
> > ```
> > 
> > 
> > This looks a bit strange to me. Why does it say `True` for the second condition?
> 
> i think this is due to sympy's simplification combined with the evaluation order: going from left to right it first checks that `x_0` is larger than or equal to zero, and if this condition fails it moves on to the next one. now `x_0` is definitely smaller than zero, so the condition `x_0 < 0` always evaluates to `True`. but yeah, it's a bit confusing to read

Maybe I should adjust `_def_sympy_output = "Piecewise((x_1, x_0 >= 0), (x_2, x_0 < 0))"` to the way sympy simplification treats the if/else condition?

HenrikMettler:
I am facing the problem, that solutions do not converge to the target function, when using the same target as in `example_multi_genome`, even for `n_generations = 5000` (for several different seeds). Not sure there if there is an error in the implementation. Maybe the problem because too difficult, due to the many different function available operator genes (6 available operators)?

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,185 @@
+"""
+Example for evolutionary regression on a piecewise target function with one single genome
+=========================================================================================
+
+Example demonstrating the use of Cartesian genetic programming for
+regression on a piecewise target function.
+Demonstrates the use of if/else branching operator gene.
```
jakobj, line 7:
```suggestion
Example demonstrating the use of Cartesian genetic programming for
regression on a piecewise target function by using the conditional (if/else) operator.
```

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,185 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+from docopt import docopt
+
+import cgp
+
+args = docopt(docopt_str)
+
+
+# %%
```
jakobj, line 31: this operator should be defined in `node_impl.py`

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,185 @@
...
+     and outputs input1 if true and input2 otherwise """
+
+    _arity = 3
+
+    def __init__(self, idx, inputs):
+        super().__init__(idx, inputs)
+        if inputs[0] > 0:
+            self._output = inputs[1]
+        else:
+            self._output = inputs[2]
```
jakobj, line 44: what is this for?

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,185 @@
...
+import cgp
+
+args = docopt(docopt_str)
+
+
+# %%
+# First we define the new operator node: If/ElseOperator for conditional outputs
+class IfElseOperator(cgp.OperatorNode):
+    """A node that checks its input0 is bigger than 0
+     and outputs input1 if true and input2 otherwise """
```
jakobj, line 35:
```suggestion
    """A node that outputs the value of its second input if its first input is non-negative, and the value of its third input otherwise."""
```

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,185 @@
...
+    _arity = 3
+
+    def __init__(self, idx, inputs):
+        super().__init__(idx, inputs)
+        if inputs[0] > 0:
+            self._output = inputs[1]
+        else:
+            self._output = inputs[2]
+
+    _def_output = "x_1 if x_0 > 0 else x_2"
```
jakobj, line 46:
```suggestion
    _def_output = "x_1 if x_0 >= 0 else x_2"
```
this is how i would define it, but if you feel strongly about it, i can live with `x_0 > 0` ;)

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,185 @@
...
+    _def_torch_output = "torch.where(x_0 > 0, x_1, x_2)"
+
+
+# %%
+# We define a piecewise target function.  The function applies different
+# transformations to the input depending whether the input is less or greater
+# than zero. Thus to achieve high fitness,
+# an individual must make use of the if/else operator.
+def f_target(x):
+    return np.select([x <= 0, x > 0], [-x, x ** 2 + 1.0])
```
jakobj, line 58:
```suggestion
    return np.select([x > 0, x <= 0], [x ** 2 + 1.0, -x])
```
for consistency with order in operator definition

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,185 @@
...
+    -------
+    Individual
+        Modified individual with updated fitness value.
+    """
+    if individual.fitness is not None:
+        return individual
+
+    n_function_evaluations = 1000
+
+    f_graph = individual.to_func()
```
jakobj, line 86:
```suggestion
    f = individual.to_func()
```

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,185 @@
...
+    y = np.empty(n_function_evaluations)
+    x = rng.uniform(-5, 5, size=n_function_evaluations)
+    for i, x_i in enumerate(x):
+        y[i] = f_graph([x_i])[0]
```
jakobj, line 90:
```suggestion
        y[i] = f([x_i])[0]
```

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,185 @@
...
+
+    loss = np.mean((target_function(x) - y) ** 2)
+    individual.fitness = -loss
+
+    return individual
+
+
+seed = 999  # Todo: find better seed
```
jakobj, line 98: i would move this below to where you actually use `rng`

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,185 @@
...
+history = {}
+history["fitness_parents"] = []
+history["champion_sympy"] = []
+
+
+def recording_callback(pop):
+    history["fitness_parents"].append(pop.fitness_parents())
+    try:
+        sympy_expression = pop.champion.to_sympy()
+    except TypeError:
```
jakobj, line 143: why is this necessary?

test/test_node.py
```diff
@@ -413,6 +413,197 @@ class DoubleParameter(cgp.OperatorNode):
     assert y == pytest.approx(p + q)
 
 
+# Node definition currently within test,
```
jakobj, line 416: this definition can be removed once the operator is defined in `node_impl.py`

test/test_node.py
```diff
@@ -413,6 +413,197 @@ class DoubleParameter(cgp.OperatorNode):
...
+        else:
+            self._output = inputs[2]
+
+    _def_output = "x_1 if x_0 > 0 else x_2"
+    _def_numpy_output = "np.piecewise(x_0, [x_0> 0, x_0<=0], [x_1 [x_0> 0] , x_2[x_0<=0]])"
+    _def_sympy_output = "Piecewise((x_1, x_0 > 0), (x_2, x_0 <= 0))"
+    _def_torch_output = "torch.where(x_0 > 0, x_1, x_2)"
+
+
+# Node definition currently within test,
```
jakobj, line 437: please remove this node also here

test/test_node.py
```diff
@@ -413,6 +413,197 @@ class DoubleParameter(cgp.OperatorNode):
...
+    genome_params = {
+        "n_inputs": 3,
+        "n_outputs": 1,
+        "n_columns": 1,
+        "n_rows": 1,
+        "levels_back": None,
+    }
+    primitives = (IfElseOperator,)
+    genome = cgp.Genome(**genome_params, primitives=primitives)
+    genome.dna = [
```
jakobj, line 500: please add a comment indicating which function this genome encode like `# f(x) = ...`

test/test_node.py
```diff
@@ -413,6 +413,197 @@ class DoubleParameter(cgp.OperatorNode):
...
+    x_1 = [0.0, 10.0, -20.0]
+    y_target_1 = [-20.0]
+    _test_graph_call_and_to_x_compilations(genome, x_1, y_target_1)
+
+    x_2 = [-1.0, 10.0, -20.0]
+    y_target_2 = [-20.0]
+    _test_graph_call_and_to_x_compilations(genome, x_2, y_target_2)
+
+
+def test_genome_for_piecewise_target_function():  # Todo: test doesn't really fit into test_node
```
jakobj, line 536: what exactly does this test that isn't covered by the test above? maybe the answer would give a hint on where to put this test ;)

jakobj: also, the `ComparisonOperator` should be removed here

jakobj (changes requested):
nice work :ok_hand: 

i've added a few inline comments that should be addressed before merging.

examples/example_piecewise_target_function.py
HenrikMettler: `simpy.simplify` throws an error when `Piecewise` has the same expression for multiple conditions. This happens sometimes in the early generations of the algorithm.

test/test_node.py
HenrikMettler: Actually this test is redundant now I think, since you correctly pointed out that only the if/else operator is sufficient, for a piecewise target. Should I just remove it?

jakobj: if it's redundant, then yes :)

examples/example_piecewise_target_function.py
jakobj: i see. is there any way to avoid this without the try-except construct? i guess this will also happen in user code, so it would be nice to catch these problems within the library? maybe we could move this try except to the `to_sympy` function and emit a warning if it's not possible?

jakobj:
looks like travis fails bc of the import order:
```bash
$ isort --check-only cgp examples test

ERROR: /home/travis/build/Happy-Algorithms-League/hal-cgp/cgp/__init__.py Imports are incorrectly sorted and/or formatted.

ERROR: /home/travis/build/Happy-Algorithms-League/hal-cgp/examples/example_piecewise_target_function.py Imports are incorrectly sorted and/or formatted.

The command "isort --check-only cgp examples test" exited with 1.
```

cgp/node_impl.py
```diff
@@ -59,3 +59,25 @@ class Parameter(OperatorNode):
...
+    is non-negative, and the value of its third input otherwise."""
+
+    _arity = 3
+    _def_output = "x_1 if x_0 >= 0 else x_2"
+    _def_numpy_output = "np.piecewise(x_0, [x_0 >= 0, x_0 < 0], [x_1 [x_0 >= 0] , x_2[x_0 < 0]])"
+    _def_sympy_output = "Piecewise((x_1, x_0 >= 0), (x_2, x_0 < 0))"
+    _def_torch_output = "torch.where(x_0 >= 0, x_1, x_2)"
+
+
+class Comparison(OperatorNode):
```
jakobj, line 75: we decided to remove this, didn't we?

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,160 @@
+"""
+Example for evolutionary regression on a piecewise target function with one single genome
```
jakobj, line 2:
```suggestion
Example for evolutionary regression on a piecewise target function
```

cgp/node_impl.py
HenrikMettler: Ah sorry, my bad I thought you meant to just remove it from the example

examples/example_piecewise_target_function.py
HenrikMettler: At the moment 
https://github.com/Happy-Algorithms-League/hal-cgp/blob/90afc8dcadcff8366b4f3641c47ea494f126bc51/cgp/individual.py#L179

Defaults to 'true', so the 'CartesianGraph'  instance simplify unless the user specifies for the individual not to. An option would be do move this try structure into 'CartesianGraph.to_sympy'. I will create a seperate PR with this

jakobj: sounds good. but we really should emit a warning, to make the user aware that something went wrong

mschmidt87:
> As suggested by @jakobj I removed the comparison operator from the example, since it is not necessary for the function discontinuity at `0`. (Actually also for other the discontinuities the if/else operator could build its own condition via Add/Sub)
> The example now passes, however the expression is not fully correct at `x=0`. I will try to find a seed that results in the precise expression.
> Please already have another look @mschmidt87

It's a pity that we didn't keep the more general form of the IfElseOperator, but I'm fine with it, if you can't make the example converge (although this should just be a question of optimizing hyperparameters, shouldn't it?)

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,160 @@
...
+ea = cgp.ea.MuPlusLambda(**ea_params)
+
+# define callback for recording of fitness over generations
+history = {}
+history["fitness_parents"] = []
+history["champion_sympy"] = []
+
+
+def recording_callback(pop):
+    history["fitness_parents"].append(pop.fitness_parents())
```
jakobj, line 113: i would suggest to just record the fitness of the champion here for efficiency reasons

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,160 @@
...
+plt.xlabel("x")
+plt.ylabel("y")
+plt.title("Target function")
+plt.legend(["target"])
+plt.subplot(122)
+plt.plot(x, champion_numpy(x.reshape(x.size, 1)), "r")
+plt.xlabel("x")
+plt.ylabel("y")
+plt.title("Evolved function")
+plt.legend(["champion"])
```
jakobj, line 160: please make sure the created plot is saved ;)

jakobj:
> > As suggested by @jakobj I removed the comparison operator from the example, since it is not necessary for the function discontinuity at `0`. (Actually also for other the discontinuities the if/else operator could build its own condition via Add/Sub)
> > The example now passes, however the expression is not fully correct at `x=0`. I will try to find a seed that results in the precise expression.
> > Please already have another look @mschmidt87
> 
> It's a pity that we didn't keep the more general form of the IfElseOperator, but I'm fine with it, if you can't make the example converge (although this should just be a question of optimizing hyperparameters, shouldn't it?)

the version that was removed was actually _less_ general than this operator (currently it's `x_1 if x_0 >= 0.0 else x_2`). what do you mean by more general? would you prefer an operator with four inputs like  `x_2 if x_0 >= x_1 else x_3`?

HenrikMettler:
Saving the figure I noticed, that the current setup does not converge to the desired target function, but to `(2.0, Eq(x_0, 0)), (-x_0, x_0 <= 0), (x_0**2 + 1.0, x_0 >= 0), (x_0**2 - x_0 - 1.0, True)` which is incorrect at `x=0`. I will test some different population initiation seeds and values for `n_columns` to get a correct solution.

mschmidt87:
> > > As suggested by @jakobj I removed the comparison operator from the example, since it is not necessary for the function discontinuity at `0`. (Actually also for other the discontinuities the if/else operator could build its own condition via Add/Sub)
> > > The example now passes, however the expression is not fully correct at `x=0`. I will try to find a seed that results in the precise expression.
> > > Please already have another look @mschmidt87
> > 
> > 
> > It's a pity that we didn't keep the more general form of the IfElseOperator, but I'm fine with it, if you can't make the example converge (although this should just be a question of optimizing hyperparameters, shouldn't it?)
> 
> the version that was removed was actually _less_ general than this operator (currently it's `x_1 if x_0 >= 0.0 else x_2`). what do you mean by more general? would you prefer an operator with four inputs like `x_2 if x_0 >= x_1 else x_3`?

No, the original version that I proposed and that I believe was first implemented, was an if/else-operator that just takes three inputs:
- x_0, a boolean or integer or float
- x_1, x_2

and execute `if x_0: x_1 else x_2`. And then you need a second operator in this example that computes `x_0 > x_1` where x_0, x_1 are its inputs. See my description in the issue #93

jakobj:
> > the version that was removed was actually _less_ general than this operator (currently it's `x_1 if x_0 >= 0.0 else x_2`). what do you mean by more general? would you prefer an operator with four inputs like `x_2 if x_0 >= x_1 else x_3`?
> 
> No, the original version that I proposed and that I believe was first implemented, was an if/else-operator that just takes three inputs:
> 
>     * x_0, a boolean or integer or float
> 
>     * x_1, x_2
> 
> 
> and execute `if x_0: x_1 else x_2`. And then you need a second operator in this example that computes `x_0 > x_1` where x_0, x_1 are its inputs. See my description in the issue #93

ah i see. indeed this would be more general at the expense of requiring more operators and changing the type of data flowing between nodes (floats __&__ bool). so for simplicity reasons i would argue in favor of keeping only this node. if it turns out to not be enough, we can still implement your suggestions at a later time, ok?

mschmidt87:
> > > the version that was removed was actually _less_ general than this operator (currently it's `x_1 if x_0 >= 0.0 else x_2`). what do you mean by more general? would you prefer an operator with four inputs like `x_2 if x_0 >= x_1 else x_3`?
> > 
> > 
> > No, the original version that I proposed and that I believe was first implemented, was an if/else-operator that just takes three inputs:
> > ```
> > * x_0, a boolean or integer or float
> > 
> > * x_1, x_2
> > ```
> > 
> > 
> > and execute `if x_0: x_1 else x_2`. And then you need a second operator in this example that computes `x_0 > x_1` where x_0, x_1 are its inputs. See my description in the issue #93
> 
> ah i see. indeed this would be more general at the expense of requiring more operators and changing the type of data flowing between nodes (floats **&** bool). so for simplicity reasons i would argue in favor of keeping only this node. if it turns out to not be enough, we can still implement your suggestions at a later time, ok?

Fine for me. But for the record: no need to control types of data flowing across operators, as explained in the issue #93 .

mschmidt87:
Any luck in getting the right expression in the example, @HenrikMettler ?

jakobj:
i'm bumping this to v0.3.0, as we are close to finishing v0.2.0. please let me know if you disagree.

mschmidt87:
@HenrikMettler did you pursue this PR any further, i.e. did you try different seeds or configurations to reach the target expression?

HenrikMettler:
Really sorry about the long delay, the error was in using a mutation rate that was approx 1 order of magnitude to small.

HenrikMettler:
Not sure why coverage decreases in `cartesian_graph`, there are no changes to it  within this PR.

test/test_node.py
```diff
@@ -413,6 +413,55 @@ class DoubleParameter(cgp.OperatorNode):
...
+        0,  # first input gene is address of first input node
+        1,  # second input gene is address of second input node
+        2,  # third input gene is address of third input node
+        ID_OUTPUT_NODE,
+        3,
+        ID_NON_CODING_GENE,
+        ID_NON_CODING_GENE,
+    ]
+
+    # f(x_0, x_1, x_2) = {x_1 if x_0 >= 0, x_2 otherwise}
```
jakobj, line 450:
```suggestion
```

test/test_node.py
```diff
@@ -413,6 +413,55 @@ class DoubleParameter(cgp.OperatorNode):
...
+    genome_params = {
+        "n_inputs": 3,
+        "n_outputs": 1,
+        "n_columns": 1,
+        "n_rows": 1,
+        "levels_back": None,
+    }
+    primitives = (cgp.IfElse,)
+    genome = cgp.Genome(**genome_params, primitives=primitives)
+    genome.dna = [
```
jakobj, line 427:
```suggestion
    # f(x_0, x_1, x_2) = {x_1 if x_0 >= 0, x_2 otherwise}
    genome.dna = [
```

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,162 @@
...
+# an individual must make use of the if/else operator.
+def f_target(x):
+    return np.select([x >= 0, x < 0], [x ** 2 + 1.0, -x])
+
+
+# %%
+# Then we define the objective function for the evolution. It uses the
+# mean-squared error between the output of the expression represented by a given
+# individual and the target function evaluated on a set of pseudo-random points.
+def objective(individual, target_function, rng):
```
jakobj, line 44:
```suggestion
def objective(individual, rng):
```
i don't think we need this argument, there is only one target function anyway, isn't there?

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,162 @@
...
+    if individual.fitness is not None:
+        return individual
+
+    n_function_evaluations = 1000
+
+    f = individual.to_func()
+    y = np.empty(n_function_evaluations)
+    x = rng.uniform(-5, 5, size=n_function_evaluations)
+    for i, x_i in enumerate(x):
+        y[i] = f([x_i])[0]
```
jakobj, line 69:
```suggestion
    f = individual.to_numpy()
    x = rng.uniform(-5, 5, size=(n_function_evaluations, 1))
    y = f(x)
```
since you anyway use numpy in this script i'd use `to_numpy` version as evaluation is faster

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,162 @@
...
+# create population that will be evolved
+pop = cgp.Population(**population_params, genome_params=genome_params)
+
+# create instance of evolutionary algorithm
+ea = cgp.ea.MuPlusLambda(**ea_params)
+
+# define callback for recording of fitness over generations
+history = {}
+history["fitness_champion"] = []
+history["champion_sympy"] = []
```
jakobj, line 109:
```suggestion
history["expr_champion"] = []
```

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,162 @@
...
+    except TypeError:
+        sympy_expression = pop.champion.to_sympy(simplify=False)
+
+    history["champion_sympy"].append(sympy_expression)
+
+
+# the objective passed to evolve should only accept one argument,
+# the individual
+seed = 1000
+rng = np.random.RandomState(seed=seed)
```
jakobj, line 125:
```suggestion
rng = np.random.RandomState(seed=population_params["seed"])
```
just to avoid introducing multiple seeds

cgp/node_impl.py
```diff
@@ -59,3 +59,14 @@ class Parameter(OperatorNode):
...
     _def_torch_output = "torch.ones(1).expand(x.shape[0]) * <p>"
+
+
+class IfElse(OperatorNode):
+    """A node that outputs the value of its second input if its first input
+    is non-negative, and the value of its third input otherwise."""
+
+    _arity = 3
+    _def_output = "x_1 if x_0 >= 0 else x_2"
+    _def_numpy_output = "np.piecewise(x_0, [x_0 >= 0, x_0 < 0], [x_1 [x_0 >= 0] , x_2[x_0 < 0]])"
```
jakobj, line 70:
```suggestion
    _def_numpy_output = "np.piecewise(x_0, [x_0 >= 0, x_0 < 0], [x_1[x_0 >= 0] , x_2[x_0 < 0]])"
```

jakobj (changes requested):
great thanks for updating this! :ok_hand: 

i've left a few inline comments that should be addressed before merging.

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,162 @@
...
+    return individual
+
+
+# %%
+# Next, we set up the evolutionary search. First, we define the parameters for
+# the population, the genomes of individuals, and the evolutionary
+# algorithm.
+population_params = {
+    "n_parents": 1,
+    "mutation_rate": 0.05,
```
jakobj, line 83:
```suggestion
    "mutation_rate": 0.03,
```
this makes the example pass after 1590 generations on my machine.

examples/example_piecewise_target_function.py
```diff
@@ -0,0 +1,157 @@
...
+plt.ylabel("y")
+plt.title("Target function")
+plt.legend(["target"])
+plt.subplot(122)
+plt.plot(x, champion_numpy(x.reshape(x.size, 1)), "r")
+plt.xlabel("x")
+plt.ylabel("y")
+plt.title("Evolved function")
+plt.legend(["champion"])
+fig.savefig("example_piecewise_function_plot.pdf", dpi=300)
```
jakobj, line 157:
```suggestion
fig.savefig("example_piecewise_target_function.pdf", dpi=300)
```

jakobj (approved):
this looks great, thanks a lot! :+1:

i've added one brief inline comment. also please rewrite history into nice commits before we merge.

HenrikMettler:
I squashed the history to the major development steps within this PR. Should I squash it more to "remove" the history of the comparison operator?

jakobj:
yes, no need to keep this history, the important points are documented in the PR discussion. i'd go for three commits again: implementation, test, example. of course this could also be one big one, but these seem to me like nice self-contained chunks.

jakobj:
awesome, great job! and merging.

PR #233, opened 2020-08-13
Fix bug in Node.__repr__

### LINKED ISSUE #232: __repr__ of CartesianGraph is broken by InputNode
This code 
```python
import cgp
import numpy as np
rng = np.random.RandomState()

genome_params = {"n_inputs": 2,
        "n_outputs": 1,
        "n_columns": 3,
        "n_rows": 3,
        "levels_back": 2,
        "primitives": (cgp.Add, cgp.Sub, cgp.ConstantFloat),
    }
genome.randomize(rng)
g = cgp.CartesianGraph(genome)
print(g)
```
leads to
```python
AttributeError                            Traceback (most recent call last)
<ipython-input-20-e1cdc681402c> in <module>
----> 1 print(g)

~/opt/hal-cgp/cgp/cartesian_graph.py in __repr__(self)
     53 
     54     def __repr__(self) -> str:
---> 55         return "CartesianGraph(" + str(self._nodes) + ")"
     56 
     57     def print_active_nodes(self) -> str:

~/opt/hal-cgp/cgp/node.py in __repr__(self)
     79     def __repr__(self) -> str:
     80         return (
---> 81             f"{self.__class__.__name__}(idx: {self.idx}, active: {self._active}, "
     82             f"arity: {self._arity}, inputs {self._inputs}, output {self._output})"
     83         )

AttributeError: 'InputNode' object has no attribute '_output'
```

mschmidt87:
This PR fixes #232 by following the suggestions of @jakobj in the ticket:

- Remove _output from repr string of Node class
- Rename _inputs to _input_nodes for better clarity
- Add test of CartesianGraph.__repr__ function
- Add test for __repr__ of 3 example nodes

### DIFF
--- a/cgp/cartesian_graph.py
+++ b/cgp/cartesian_graph.py
@@ -173,7 +173,7 @@ class CartesianGraph:
                 continue
             active_nodes_by_hidden_column_idx[self._hidden_column_idx(node.idx)].add(node)
 
-            for i in node.inputs:
+            for i in node.input_nodes:
                 nodes_to_process.append(self._nodes[i])
 
         return active_nodes_by_hidden_column_idx
--- a/cgp/node.py
+++ b/cgp/node.py
@@ -33,7 +33,7 @@ class Node:
 
     _arity: int
     _active: bool = False
-    _inputs: List[int]
+    _input_nodes: List[int]
     _output: float
     _output_str: str
     _idx: int
@@ -49,7 +49,7 @@ class Node:
             List of integers specifying the input nodes to this node.
         """
         self._idx = idx
-        self._inputs = inputs[: self._arity]
+        self._input_nodes = inputs[: self._arity]
 
         assert idx not in inputs
 
@@ -66,11 +66,11 @@ class Node:
 
     @property
     def max_arity(self) -> int:
-        return len(self._inputs)
+        return len(self._input_nodes)
 
     @property
-    def inputs(self) -> List[int]:
-        return self._inputs
+    def input_nodes(self) -> List[int]:
+        return self._input_nodes
 
     @property
     def idx(self) -> int:
@@ -79,7 +79,7 @@ class Node:
     def __repr__(self) -> str:
         return (
             f"{self.__class__.__name__}(idx: {self.idx}, active: {self._active}, "
-            f"arity: {self._arity}, inputs {self._inputs}, output {self._output})"
+            f"arity: {self._arity}, input_nodes {self._input_nodes}"
         )
 
     def pretty_str(self, n: int) -> str:
@@ -102,7 +102,7 @@ class Node:
             if self._arity > 0:
                 s += "("
                 for i in range(self._arity):
-                    s += f"{self._inputs[i]:02d},"
+                    s += f"{self._input_nodes[i]:02d},"
                 s = s[:-1]
                 s += ")"
                 for i in range(self.max_arity - self._arity):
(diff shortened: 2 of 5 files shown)


cgp/node.py
```diff
@@ -33,7 +33,7 @@ class Node:
 
     _arity: int
     _active: bool = False
-    _inputs: List[int]
+    _input_nodes: List[int]
```
jakobj, line 36: hmm, shouldn't this be renamed in many other places as well? like in `Node.__init__`?

cgp/node.py
```diff
@@ -79,7 +79,7 @@ def idx(self) -> int:
     def __repr__(self) -> str:
         return (
             f"{self.__class__.__name__}(idx: {self.idx}, active: {self._active}, "
-            f"arity: {self._arity}, inputs {self._inputs}, output {self._output})"
+            f"arity: {self._arity}, inputs {self._inputs}"
```
jakobj, line 82: e.g., here

test/test_cartesian_graph.py
```diff
@@ -602,3 +602,11 @@ def test_pretty_str_with_unequal_inputs_rows_outputs():
...
 """
     assert graph.pretty_str() == expected_pretty_str
+
+
+def test_repr(rng, genome_params):
+    genome = cgp.Genome(**genome_params)
+    genome.randomize(rng)
+    graph = cgp.CartesianGraph(genome)
+    # Assert that the CartesianGraph.__repr__ doesn't raise an error
+    _ = str(graph)
```
jakobj, line 612:
```suggestion
    # Assert that the CartesianGraph.__repr__ doesn't raise an error
    str(cgp.CartesianGraph(genome))
```

test/test_node.py
```diff
@@ -448,3 +448,23 @@ class CustomAdd(cgp.OperatorNode):
             _arity = 2
             _def_output = "x_0 + x_1"
             _def_sympy_output = "x_0 +/ x_1"
+
+
+def test_repr():
+    idx = 0
+    inputs = [1, 2, 3, 4]
```
jakobj, line 455:
```suggestion
    input_nodes = [1, 2, 3, 4]
```

cgp/node.py
mschmidt87: Yes, thanks, I changed this in the other places.

cgp/cartesian_graph.py
```diff
@@ -173,7 +173,7 @@ def _determine_active_nodes(self) -> Dict[int, Set[Node]]:
                 continue
             active_nodes_by_hidden_column_idx[self._hidden_column_idx(node.idx)].add(node)
 
-            for i in node.inputs:
+            for i in node.input_nodes:
```
jakobj, line 176: when reading this line, i would prefer something like `for input_idx in node.input_node_indices`, i.e., renaming `input_nodes` -> `input_node_indices` or this overdoing it? ;)

test/test_node.py
```diff
@@ -12,16 +12,16 @@ def test_inputs_are_cut_to_match_arity():
 
     """
     idx = 0
```
jakobj, line 14:
```suggestion
    node_idx = 0
```

cgp/cartesian_graph.py
mschmidt87: I think this is overdoing it. ;-)

test/test_node.py
mschmidt87: I think this is unrelated to this PR and the class member is called `idx` anyway.

PR #234, opened 2020-08-13
Add documentation page about nodes

mschmidt87:
This PR adds a page to the documentation about nodes. It explains the basics of nodes and delivers two simple examples of custom nodes.

Note: with `doctest` and its sphinx extension, we can actually test those snippets to make sure that they run, but I am still figuring out how to do that. I will create a separate PR for that.

Related to #174


docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
+=====
+Nodes
+=====
+
+Nodes are the buildings blocks of the mathematical expressions represented by an individual.
+They are selected in the invididual's genome (see :ref:`pop_ind`).
```
jakobj, line 6:
```suggestion
Nodes define the buildings blocks of the mathematical expressions represented by an individual.
They are represented in the invididual's genome (see :ref:`pop_ind`).
```

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
...
+
+A node takes a certain number of inputs (specified by its `arity` which can also be 0) and transforms them into one output value. A priori, there are no limitations regarding the type of transformation, as long as it can be expressed in pure Python, SymPy_, PyTorch_, and NumPy_.
```
jakobj, line 8:
```suggestion
A node receives data from a certain number of input nodes (specified by its `arity` which can be any non-zero number) and transforms them into an output value. A priori, there are no limitations regarding the type of transformation, as long as it can be expressed in pure Python, SymPy_, PyTorch_, and NumPy_.
```
i would avoid saying "one output" since the numpy functions operate on batches

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
...
+
+----------------------------
+Implemented nodes in hal-cgp
```
jakobj, line 11:
```suggestion
Default nodes in hal-cgp
```
maybe?

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
...
+----------------------------
+
+hal-cgp comes with a set of implemented nodes:
+
+- The `InputNode` and `OutputNode` are special nodes that represent the input to and output from the computational graph.
+- A set of basic mathematical operations is implemented as operator nodes: addition (``Add``), subtraction (``Sub``), multiplication (``Mul``), division (``Div``) and power (``Pow``).
```
jakobj, line 17:
```suggestion
- A set of basic mathematical operations is implemented as `OoperatorNodes`: addition (``Add``), subtraction (``Sub``), multiplication (``Mul``), division (``Div``) and power (``Pow``).
```
could we link to the docstring for `OperatorNodes`?

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
...
+- A `ConstantFloat` that represents a constant in mathematical expressions.
+- A `ParameterNode` that holds adjustable parameter (see :ref:`local_search`).
```
jakobj, line 19:
```suggestion
- A `ParameterNode` that represents an adjustable parameter (see :ref:`local_search`).
```

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
...
+
+------------
+Custom nodes
+------------
+
+In hal-cgp, it is straightforward to implement new, custom nodes.
+The custom nodes needs to be implemented as a subclass of the `OperatorNode` (:meth:`cgp.node.OperatorNode`) and define its ``arity`` as well as its transformation, defined by the ``_def_output`` member of the class. The ``_def_output`` string defines the transformation in pure Python and is used to define the transformation in SymPy_, PyTorch_, and Numpy_.
```
jakobj, line 26:
```suggestion
The custom nodes needs to be implemented as a subclass of the `OperatorNode` (:meth:`cgp.node.OperatorNode`) and define its ``arity`` as well as its transformation, defined by the ``_def_output`` member of the class. The ``_def_output`` string defines the transformation in pure Python and is used to automatically define the transformation in SymPy_, PyTorch_, and Numpy_. In this string `x_i` refers to the `i`th input to the node.
```

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
...
+
+
+As a simple example, we implement a custom node that doubles its input:
+
+  .. code-block:: python
+
+		  
+     import cgp
+     import math
```
jakobj, line 35:
```suggestion
```

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
...
+
+     
+     class Double(cgp.OperatorNode): 
+         """ A node that calculates the exponential of its input.
```
jakobj, line 39:
```suggestion
         """ A node that doubles its input. 
```

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
...
+     import math
+
+     
+     class Double(cgp.OperatorNode): 
+         """ A node that calculates the exponential of its input. 
+	 """ 
+	 _arity = 1 
+	 _def_output = "2*x_0"
+
+If the representation of the transformation in SymPy_, PyTorch_, or NumPy_ differs from pure Python, we need to define dedicated expressions with additional class members. For instance, to implement a node that computes the exponential of an input, we define the same operation in four different expressions:
```
jakobj, line 44:
```suggestion
If the representation of the transformation in SymPy_, PyTorch_, or NumPy_ differs from pure Python, we need to define dedicated expressions via the `_def_X_output` members. For instance, to implement a node that computes the exponential of an input, we define the same operation in four different expressions:
```

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
...
+	 
+  .. code-block:: python
+
+		  
+     import cgp
+     import math
```
jakobj, line 50: either remove math or add sympy numpy and torch

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,71 @@
...
+	 """ 
+	 _arity = 1 
+	 _def_output = "math.exp(x_0)" 
+	 _def_numpy_output = "np.exp(x_0)" 
+	 _def_torch_output = "torch.exp(x_0)"
+	 _def_sympy_output = "exp(x_0)"
+
+	
+	 
+For a complete example that uses custom nodes, please refer to :ref:`sphx_glr_auto_examples_example_parametrized_nodes.py`.
```
jakobj, line 64:
```suggestion
For a complete example that uses custom nodes with adjustable parameters, please refer to :ref:`sphx_glr_auto_examples_example_parametrized_nodes.py`.
```

docs/documentation/nodes.rst
mschmidt87: your change is not correct, I think, because the arity can be zero.

docs/documentation/nodes.rst
mschmidt87: Don't we need to import this package if we use it in the string?

mschmidt87: How does this work if we don't import it?

docs/documentation/nodes.rst
mschmidt87: how does this work if we don't import math?

docs/documentation/nodes.rst
jakobj: good point, of course i meant "non-negative" ;)

docs/documentation/nodes.rst
jakobj: man you're asking questions i can't answer ;) i think it's important that the packages are available when the function is compiled, i.e., in `cartesian_graph.py`. since the expression is not evaluated here i don't think you need the imports, could you give that a try please?

i wanted to look into this problem more deeply bc if we can provide a "compilation context" (context managers ftw?) for nodes, we might be able to define and use arbitrary functions in `_def_output`, sort of like:
```python
def myifelse(x_0, x_1, x_2):
    if x_0 < 0:
        return x_1
    else:
        return x_2

class MyIfElseNode(cgp.OperatorNode):
    _arity = 3
    _def_output = "myifelse(x_0, x_1, x_2)"
```

this would provide great flexibility:tm:

jakobj: ah, and for completeness, in this node you don't use `math` ;)

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,91 @@
...
+	 _arity = 1 
+	 _def_output = "math.exp(x_0)" 
+	 _def_numpy_output = "np.exp(x_0)" 
+	 _def_torch_output = "torch.exp(x_0)"
+	 _def_sympy_output = "exp(x_0)"
+
+	
+
+We can make this custom mode more flexible by adding an adjustable Parameter for the scale of the exponential.
+To this end, we add the ``<scale>`` term in the expressions and define its initial value in the ``_initial_values`` class member:
```
jakobj, line 65: interesting `<scale>` actually works as a name? nice!

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,91 @@
...
+	 
+  .. code-block:: python
+
+		  
+     import cgp
+     import math
+
+     
+     class Exp(cgp.OperatorNode): 
+         """ A node that calculates the exponential of its input.
```
jakobj, line 75:
```suggestion
         """ A node that calculates the exponential of its input. An adjustable parameter governs the scale.
```

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,91 @@
...
+	 """ 
+	 _arity = 1 
+	 _def_output = "math.exp(x_0)" 
+	 _def_numpy_output = "np.exp(x_0)" 
+	 _def_torch_output = "torch.exp(x_0)"
+	 _def_sympy_output = "exp(x_0)"
+
+	
+
+We can make this custom mode more flexible by adding an adjustable Parameter for the scale of the exponential.
```
jakobj, line 64:
```suggestion
We can make this custom mode more flexible by adding an adjustable parameter for the scale of the exponential. Parameter names are enclosed by angle brackets in the expressions. Furthermore, each parameter requires a function that returns its initial values to be defined in the `_initial_values` dict.
```

docs/documentation/nodes.rst
jakobj, line 65:
```suggestion
We add the ``<scale>`` term in the expressions and define its initial value in the ``_initial_values`` class member:
```
(only if you use the changes above)

docs/documentation/nodes.rst
mschmidt87: Right, yes. Okay, the import is not required, indeed.
But yes, we have to find a solution for this, because relying on imports in `cartesian_graph.py` is not a good solution. Can you create an issue for that?

docs/documentation/nodes.rst
mschmidt87: okay, impemented.

docs/documentation/nodes.rst
mschmidt87: see above

docs/documentation/nodes.rst
mschmidt87: Running this code snippet doesn't fail at least, so it passes the validation.

mschmidt87:
Updated the PR. @jakobj

docs/documentation/nodes.rst
jakobj: done

docs/documentation/nodes.rst
```diff
@@ -0,0 +1,85 @@
...
+A node receives data from a certain number of input nodes (specified by its `arity` which can be any non-negative number) and transforms them into one output value. A priori, there are no limitations regarding the type of transformation, as long as it can be expressed in pure Python, SymPy_, PyTorch_, and NumPy_.
+
+----------------------------
+Default nodes in hal-cgp
+----------------------------
+
+hal-cgp comes with a set of implemented nodes:
+
+- The `InputNode` and `OutputNode` are special nodes that represent the input to and output from the computational graph.
+- A set of basic mathematical operations is implemented as `OoperatorNodes`: addition (``Add``), subtraction (``Sub``), multiplication (``Mul``), division (``Div``) and power (``Pow``).
```
jakobj, line 17:
```suggestion
- A set of basic mathematical operations is implemented as `OperatorNodes`: addition (``Add``), subtraction (``Sub``), multiplication (``Mul``), division (``Div``) and power (``Pow``).
```

Issue #224, opened 2020-08-03
Controlling the Wallclock time by limiting the Number of individual evaluations


HenrikMettler:
Currently the evolve function is limited by the number of generations or attaining a fitness goal (`min_fitness`). However the number of generations is only loosely related to the number of fitness evaluations, since only individuals with a novel phenotype (new configuration of active genes) are evaluated. In a qualitative comparison study for cgp's evolutionary mechanisms (https://ieeexplore.ieee.org/document/6815728/?arnumber=6815728) Goldman et al. benchmark their parameter tuning on the number of fitness evaluations. I would argue that this could be a reasonable approach for us as well, since presumably in most of the implementations: 
`T(evaluating_individual) = big`
`T(calculating_genome_output) = small`
Therefore allowing the user to define a maximal number of fitness evaluations might simplify the control of a maximal simulation duration.

mschmidt87:
Makes sense to me. This could be an additional keyword argument for the `evolve` function which can be set instead of max_generations.

jakobj:
an additional keyword makes sense, but i wouldn't make the two exclusive: just break whenever one of the two reaches it's maximal value.

PR #238, opened 2020-08-24
Limiting fitness evaluations

HenrikMettler:
This PR draft is a suggestion for a solution to #224 
It works by counting the calls of the objective function using a decorator (as suggested eg: https://stackoverflow.com/questions/21716940/is-there-a-way-to-track-the-number-of-times-a-function-is-called). I am not sure if this is a reasonable approach, mainly two things bother me: 

- the subtraction if fitness is not reevalutated (necessary since we don't want to count those)
- if a caching decorator is used for the inner objective, the counts of the objective function calls might not correspond to full function evaluations anymore

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -52,7 +52,7 @@ class MuPlusLambda:
         self.local_search = local_search
         self.k_local_search = k_local_search
 
-        self.n_objective_calls = 0
+        self.n_objective_calls: int = 0
 
     def initialize_fitness_parents(
         self, pop: Population, objective: Callable[[IndividualBase], IndividualBase]
@@ -104,8 +104,6 @@ class MuPlusLambda:
         # population instead of the other way around
         combined = offsprings + pop.parents
 
-        self.update_n_objective_calls(combined)
-
         # we follow a two-step process for selection of new parents:
         # we first determine the fitness for all individuals, then, if
         # applicable, we apply local search to the k_local_search
@@ -150,6 +148,8 @@ class MuPlusLambda:
         self, combined: List[IndividualBase], objective: Callable[[IndividualBase], IndividualBase]
     ) -> List[IndividualBase]:
 
+        self.update_n_objective_calls(combined)
+
         # computes fitness on all individuals, objective functions
         # should return immediately if fitness is not None
         if self.n_processes == 1:
@@ -193,5 +193,5 @@ class MuPlusLambda:
          i.e., for which the objective function will be evaluated.
         """
         for individual in combined:
-            if individual.fitness is not None:
+            if individual.fitness is None:
                 self.n_objective_calls += 1
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -52,6 +52,8 @@ class MuPlusLambda:
         self.local_search = local_search
         self.k_local_search = k_local_search
 
+        self.n_objective_calls = 0
+
     def initialize_fitness_parents(
         self, pop: Population, objective: Callable[[IndividualBase], IndividualBase]
     ) -> None:
@@ -102,6 +104,8 @@ class MuPlusLambda:
         # population instead of the other way around
         combined = offsprings + pop.parents
 
+        self.update_n_objective_calls(combined)
+
         # we follow a two-step process for selection of new parents:
         # we first determine the fitness for all individuals, then, if
         # applicable, we apply local search to the k_local_search
@@ -145,6 +149,7 @@ class MuPlusLambda:
     def _compute_fitness(
         self, combined: List[IndividualBase], objective: Callable[[IndividualBase], IndividualBase]
     ) -> List[IndividualBase]:
+
         # computes fitness on all individuals, objective functions
         # should return immediately if fitness is not None
         if self.n_processes == 1:
@@ -182,3 +187,11 @@ class MuPlusLambda:
 
         """
         return combined[:n_parents]
+
+    def update_n_objective_calls(self, combined: List[IndividualBase]) -> None:
+        """Increase n_objective_calls by the number of individuals with fitness=None,
+         i.e., for which the objective function will be evaluated.
+        """
+        for individual in combined:
+            if individual.fitness is not None:
+                self.n_objective_calls += 1
--- a/cgp/hl_api.py
+++ b/cgp/hl_api.py
@@ -11,8 +11,9 @@ def evolve(
     pop: Population,
     objective: Callable[[IndividualBase], IndividualBase],
     ea: MuPlusLambda,
-    max_generations: int,
     min_fitness: float,
+    max_generations: int = np.inf,
+    max_objective_calls: int = np.inf,
     print_progress: Optional[bool] = False,
     callback: Optional[Callable[[Population], None]] = None,
     n_processes: int = 1,
@@ -31,10 +32,15 @@ def evolve(
     ea : EA algorithm instance
         The evolution algorithm. Needs to be a class instance with an
         `initialize_fitness_parents` and `step` method.
-    max_generations : int
-        Maximum number of generations.
     min_fitness : float
         Minimum fitness at which the evolution is stopped.
+    max_generations : int
+        Maximum number of generations.
+        Defaults to positive infinity.
+        Either this or `max_objective_calls` needs to be set to a finite value.
+    max_objective_calls: int
+        Maximum number of function evaluations.
+        Defaults to positive infinity.
     print_progress : boolean, optional
         Switch to print out the progress of the algorithm. Defaults to False.
     callback :  callable, optional
@@ -49,6 +55,8 @@ def evolve(
     -------
     None
     """
+    if np.isinf(max_generations) and np.isinf(max_objective_calls):
+        raise ValueError("Either max_generations or max_objective_calls must be finite.")
 
     ea.initialize_fitness_parents(pop, objective)
     if callback is not None:
@@ -57,7 +65,7 @@ def evolve(
     # perform evolution
     max_fitness = np.finfo(np.float).min
     # Main loop: -1 offset since the last loop iteration will still increase generation by one
-    while pop.generation < max_generations - 1:
+    while pop.generation < max_generations - 1 and ea.n_objective_calls < max_objective_calls:
 
         pop = ea.step(pop, objective)
 
@@ -69,11 +77,21 @@ def evolve(
             max_fitness = pop.champion.fitness
 
         if print_progress:
-            print(
-                f"\r[{pop.generation + 1}/{max_generations}] max fitness: {max_fitness}\033[K",
-                end="",
-                flush=True,
-            )
+            if np.isfinite(max_generations):
+                print(
+                    f"\r[{pop.generation + 1}/{max_generations}] max fitness: {max_fitness}\033[K",
+                    end="",
+                    flush=True,
+                )
+            elif np.isfinite(max_objective_calls):
+                print(
+                    f"\r[{ea.n_objective_calls}/{max_objective_calls}] "
+                    f"max fitness: {max_fitness}\033[K",
+                    end="",
+                    flush=True,
+                )
+            else:
+                assert False  # should never be reached
 
         if callback is not None:
             callback(pop)
--- a/test/test_ea_mu_plus_lambda.py
+++ b/test/test_ea_mu_plus_lambda.py
@@ -221,3 +221,43 @@ def test_create_new_parent_population(population_params, genome_params, ea_param
     # we picked the first three individuals
     new_parents = ea._create_new_parent_population(3, pop.parents)
     assert new_parents == pop.parents[:3]
+
+
+def test_update_n_objective_calls(population_params, genome_params, ea_params):
+    def objective(individual):
+        individual.fitness = float(individual.idx)
+        return individual
+
+    n_objective_calls_expected = 0
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    ea = cgp.ea.MuPlusLambda(**ea_params)
+    assert ea.n_objective_calls == n_objective_calls_expected
+
+    ea.initialize_fitness_parents(pop, objective)
+    n_objective_calls_expected = population_params["n_parents"]
+    assert ea.n_objective_calls == n_objective_calls_expected
+
+    n_generations = 100
+    for _ in range(n_generations):
+        offsprings = ea._create_new_offspring_generation(pop)
+        combined = offsprings + pop.parents
+        n_objective_calls_expected += sum([1 for ind in combined if ind.fitness is None])
+        combined = ea._compute_fitness(combined, objective)
+        assert n_objective_calls_expected == ea.n_objective_calls
+
+
+def test_update_n_objective_calls_mutation_rate_one(population_params, genome_params, ea_params):
+    def objective(individual):
+        individual.fitness = float(individual.idx)
+        return individual
+
+    population_params["mutation_rate"] = 1.0
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    ea = cgp.ea.MuPlusLambda(**ea_params)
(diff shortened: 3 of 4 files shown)


jakobj:
thanks for working on this! didn't have time yet to check it in detail, but as i already mentioned we could think about introducing a general "objective" decorator, that would allow us to do such things like counting the number of objective evaluations in a simpler fashion. this would also remove the need to manually add the `if individual.fitness is None` statement to every objective. it would look something like:
```python
@cgp.objective
def objective(individual):
    ....
    return individual
```
what do you think @mschmidt87?

mschmidt87:
Totally agree, a general objective wrapper sounds awesome. So, what you're saying is that within this PR, we would use @HenrikMettler 's code and
- add the `if individual.fitness is None` check to it
- move the decorator from the example to `utils.py`
, right?

mschmidt87:
Little activity on this PR, @HenrikMettler are you working on this PR?

HenrikMettler:
I moved the wrapper into `utils.py`.  Since I am not sure this is the way you were thinking @jakobj I didn't adapt all the examples. Also many tests in `test_hl_api.py` and `test_utils.py` fail under the current, because their `objectives` have no attribute `calls`

cgp/hl_api.py
```diff
@@ -31,10 +32,12 @@ def evolve(
...
         The evolution algorithm. Needs to be a class instance with an
         `initialize_fitness_parents` and `step` method.
-    max_generations : int
-        Maximum number of generations.
     min_fitness : float
         Minimum fitness at which the evolution is stopped.
+    max_generations : int
+        Maximum number of generations. Defaults to inf
+    max_evaluations: int
+        Maximum number of function evaluations. Defaults to inf
```
jakobj, line 40:
```suggestion
    max_generations : int
        Maximum number of generations. Defaults to positive infinity. Either this or `max_evaluations` need to be set to a finite value.
    max_evaluations: int
        Maximum number of function evaluations. Defaults to positive infinity.
```

cgp/hl_api.py
```diff
@@ -49,6 +52,12 @@ def evolve(
     -------
     None
     """
+    if max_generations == np.inf and max_evaluations == np.inf:
```
jakobj, line 55:
```suggestion
    if np.isinf(max_generations) and np.isinf(max_evaluations):
```

cgp/hl_api.py
```diff
@@ -49,6 +52,12 @@ def evolve(
...
+        raise (
+            "Faulty configuration of evolve parameters: "
+            "Either max_generations or max_evaluation must be finite, "
+            "to avoid non terminal evolutionary process"
+        )
```
jakobj, line 60:
```suggestion
        raise ValueError("Either `max_generations` or `max_evaluation` must be finite.")
```

cgp/hl_api.py
```diff
@@ -69,11 +78,20 @@ def evolve(
             max_fitness = pop.champion.fitness
 
         if print_progress:
-            print(
-                f"\r[{pop.generation + 1}/{max_generations}] max fitness: {max_fitness}\033[K",
-                end="",
-                flush=True,
-            )
+            if not max_generations == np.inf:
```
jakobj, line 81:
```suggestion
            if np.isfinite(max_generations):
```

cgp/hl_api.py
```diff
@@ -69,11 +78,20 @@ def evolve(
...
+                print(
+                    f"\r[{pop.generation + 1}/{max_generations}] max fitness: {max_fitness}\033[K",
+                    end="",
+                    flush=True,
+                )
+            elif not max_evaluations == np.inf:
```
jakobj, line 87:
```suggestion
            elif np.isfinite(max_evaluations):
```

cgp/hl_api.py
```diff
@@ -69,11 +78,20 @@ def evolve(
...
+                print(
+                    f"\r[{objective.calls}/{max_evaluations}] max fitness: {max_fitness}\033[K",
```
jakobj, line 89: doesn't this also need a `+1` like in the other if branch?

cgp/utils.py
```diff
@@ -201,3 +201,26 @@ def primitives_from_class_names(primitives_str: Tuple[str, ...]) -> Tuple[Type[N
         primitives.append(primitives_dict[s])
 
     return tuple(primitives)
+
+
+def objective(f: Callable[[IndividualBase], IndividualBase]):
```
jakobj, line 206:
```suggestion
def objective(f: Callable[[IndividualBase], IndividualBase]) -> Callable[[IndividualBase], IndividualBase]:
```

cgp/utils.py
```diff
@@ -201,3 +201,26 @@ def primitives_from_class_names(primitives_str: Tuple[str, ...]) -> Tuple[Type[N
...
+    """ Wrapper for the objective function.
+
+    Wraps the objective function and counts the calls of the function.
```
jakobj, line 209:
```suggestion
    """ Decorator for the objective function.
    
    Skips evaluations of individuals with fitness != None and counts number of objective calls.
```

cgp/utils.py
```diff
@@ -201,3 +201,26 @@ def primitives_from_class_names(primitives_str: Tuple[str, ...]) -> Tuple[Type[N
...
+
+    Parameters
+    ----------
+    f: Callable
+        The objective function.
+    """
+
+    @functools.wraps(f)
+    def wrapped(individual: IndividualBase, *args, **kwargs):
```
jakobj, line 218:
```suggestion
    def wrapped(individual: IndividualBase, *args, **kwargs) -> IndividualBase:
```

cgp/utils.py
```diff
@@ -201,3 +201,26 @@ def primitives_from_class_names(primitives_str: Tuple[str, ...]) -> Tuple[Type[N
...
+        if individual.fitness is None:
+            wrapped.calls += 1
+            return f(individual, *args, **kwargs)
+        else:
+            return individual
+
+    wrapped.calls = 0
```
jakobj, line 225:
```suggestion
    wrapped.calls: int = 0
```

cgp/utils.py
```diff
@@ -201,3 +201,26 @@ def primitives_from_class_names(primitives_str: Tuple[str, ...]) -> Tuple[Type[N
...
+
+def objective(f: Callable[[IndividualBase], IndividualBase]):
+    """ Wrapper for the objective function.
+
+    Wraps the objective function and counts the calls of the function.
+
+    Parameters
+    ----------
+    f: Callable
+        The objective function.
```
jakobj, line 214:
```suggestion
    f : Callable
        The objective function.
```

examples/example_minimal.py
```diff
@@ -41,9 +41,8 @@ def f_target(x):
 # on a set of random points.
 
 
+@cgp.utils.objective
```
jakobj, line 44: as we'll be using this in every script, maybe pull it up into the cgp namespace by importing it in `cgp/__init__.py`? that would allows us to write `@cgp.objective` which seems nicer.

jakobj (changes requested):
thanks for the changes :ok_hand: i've left some inline comments that should be addressed.

cgp/hl_api.py
HenrikMettler: I don't think so, since the `objective.calls` has no offset to the number of evaluations. What I just realized could happen though, is that there are too many `evaluations`. Since `ea.step` (https://github.com/HenrikMettler/hal-cgp/blob/212bdea9dd53207ba0f0ec097d2596d0bd2f9767/cgp/hl_api.py#L71) can call the `objective` function more than once (for multiple offspring with different phenotype). I don't know if there is an easy fix to this without moving the while loop for `objective.calls` into the `ea.step`

HenrikMettler:
Should I introduce the objective decorator for all examples and tests within this PR, or in a separate PR?

jakobj:
> Should I introduce the objective decorator for all examples and tests within this PR, or in a separate PR?

let's first figure out a suitable implementation that works with more than one process. we might end up with a different solution than the decorator after all.

HenrikMettler:
As suggested by @jakobj the counter for objective function calls is now moved to the evolutionary algorithm. Removing it from the `objective` decorator enables using it with multiple processes running in parallel. I left everything for the `objective` decorator within this PR, even though it doesn't really belong to this topic anymore

cgp/ea/mu_plus_lambda.py
```diff
@@ -52,6 +52,8 @@ def __init__(
         self.local_search = local_search
         self.k_local_search = k_local_search
 
+        self.objective_calls = 0
```
jakobj, line 55:
```suggestion
        self.n_objective_calls = 0
```
to indicate that this is a number. we use it in a couple of places, so why not here as well as it can increase readability.

cgp/ea/mu_plus_lambda.py
```diff
@@ -102,6 +104,9 @@ def step(
         # population instead of the other way around
         combined = offsprings + pop.parents
 
+        # before computing the fitness update the count of objective function calls
+        self.update_objective_counts(combined)
```
jakobj, line 108:
```suggestion
        self.update_n_objective_calls(combined)
```

cgp/ea/mu_plus_lambda.py
```diff
@@ -102,6 +104,9 @@ def step(
         # population instead of the other way around
         combined = offsprings + pop.parents
 
+        # before computing the fitness update the count of objective function calls
```
jakobj, line 107:
```suggestion
```

cgp/ea/mu_plus_lambda.py
```diff
@@ -182,3 +188,8 @@ def _create_new_parent_population(
 
         """
         return combined[:n_parents]
+
+    def update_objective_counts(self, combined: List[IndividualBase]) -> None:
```
jakobj, line 192:
```suggestion
    def update_objective_counts(self, combined: List[IndividualBase]) -> None:
        """Increase n_objective_calls by the number of individuals with fitness=None, i.e., for which the objective function will be evaluated.
        """ 
```

cgp/hl_api.py
```diff
@@ -31,10 +32,15 @@ def evolve(
...
         The evolution algorithm. Needs to be a class instance with an
         `initialize_fitness_parents` and `step` method.
-    max_generations : int
-        Maximum number of generations.
     min_fitness : float
         Minimum fitness at which the evolution is stopped.
+    max_generations : int
+        Maximum number of generations.
+        Defaults to positive infinity.
+        Either this or `max_evaluations` need to be set to a finite value.
```
jakobj, line 40:
```suggestion
        Either this or `max_objective_calls` needs to be set to a finite value.
```

cgp/utils.py
```diff
@@ -201,3 +201,25 @@ def primitives_from_class_names(primitives_str: Tuple[str, ...]) -> Tuple[Type[N
         primitives.append(primitives_dict[s])
 
     return tuple(primitives)
+
+
+def objective(
```
jakobj, line 206: i would remove this from this PR, we don't use it for keeping track of the number of objective function evaluations, do we?

examples/example_minimal.py
```diff
@@ -82,7 +80,8 @@ def objective(individual):
 
 ea_params = {"n_offsprings": 4, "tournament_size": 2, "n_processes": 2}
 
-evolve_params = {"max_generations": int(args["--max-generations"]), "min_fitness": 0.0}
+evolve_params = {"max_generations": np.inf, "min_fitness": 0.0, "max_objective_calls": 10000}
```
jakobj, line 83: is this change intentional or remaining from testing? if intentional, what's the motivation?

jakobj (changes requested):
thanks for (re)implementing this so quickly! :rocket: 

i've added a few inline comments. also please add a unit test.

examples/example_minimal.py
HenrikMettler: It is still from testing. Should the new functionality be in a separate example (or just available and covered by tests)?

jakobj: i'd suggest to just have it available, documented and covered by tests. no need for an additional example.

test/test_ea_mu_plus_lambda.py
```diff
@@ -221,3 +221,46 @@ def test_create_new_parent_population(population_params, genome_params, ea_param
...
+def test_update_n_objective_calls(population_params, genome_params, ea_params):
+    def objective(individual):
+        individual.fitness = float(individual.idx)
+        return individual
+
+    # create a population and an instance of the evolutionary algorithm and
+    # check that number of objective calls is initially zero
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    ea = cgp.ea.MuPlusLambda(**ea_params)
+    assert ea.n_objective_calls == 0
```
jakobj, line 235:
```suggestion
    # check that number of objective calls is initially zero
    n_objective_calls_expected = 0
    pop = cgp.Population(**population_params, genome_params=genome_params)
    ea = cgp.ea.MuPlusLambda(**ea_params)
    assert ea.n_objective_calls == n_objective_calls_expected
```
i would introduce this `_expected` variable and set it always according to what you expect such that you can use it in the `assert`s. i think this would make the code clearer and also the comments obsolete.

test/test_ea_mu_plus_lambda.py
```diff
@@ -221,3 +221,46 @@ def test_create_new_parent_population(population_params, genome_params, ea_param
...
+    n_objective_calls = ea.n_objective_calls
+    n_generations = 100
+    for _ in range(n_generations):
+        offsprings = ea._create_new_offspring_generation(pop)
+        combined = offsprings + pop.parents
+        n_objective_calls += sum([1 for ind in combined if ind.fitness is None])
+        combined = ea._compute_fitness(combined, objective)
+        assert n_objective_calls == ea.n_objective_calls
+
+    # create a population with mutation rate one and check that the number
```
jakobj, line 253: nice idea for a second test! but i'd suggest to move it to a separate `test_` function with an appropriate name.

test/test_hl_api.py
```diff
@@ -135,6 +135,22 @@ def f1(x):
     assert abs(pop.champion.fitness) == pytest.approx(0.0)
 
 
+def test_finite_evolve(population_params, genome_params, ea_params):
```
jakobj, line 138:
```suggestion
def test_finite_max_generations_or_max_objective_calls(population_params, genome_params, ea_params):
```

test/test_hl_api.py
```diff
@@ -135,6 +135,22 @@ def f1(x):
...
+    def objective(individual):
+        individual.fitness = float(individual.idx)
+        return individual
+
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    ea = cgp.ea.MuPlusLambda(**ea_params)
+    evolve_params = {
+        "max_generations": np.inf,
+        "min_fitness": -1e-12,
```
jakobj, line 147:
```suggestion
        "min_fitness": 0.0,
```
the loop exit criterion takes floating point imprecision into account, no need for such funky fitness values ;)

jakobj (changes requested):
thanks for the update! :ok_hand: 

some more inline comments, then it's ready to merge!

test/test_ea_mu_plus_lambda.py
```diff
@@ -228,39 +228,37 @@ def objective(individual):
...
-    # to the number of parents
     ea.initialize_fitness_parents(pop, objective)
-    assert ea.n_objective_calls == population_params["n_parents"]
+    n_objective_calls_expected = population_params["n_parents"]
+    assert ea.n_objective_calls == n_objective_calls_expected
 
-    # count the number of objective calls by adding the number of offspring with fitness = None
-    # at each generation
-    n_objective_calls = ea.n_objective_calls
+    n_objective_calls_expected = ea.n_objective_calls
```
jakobj, line 240:
```suggestion
```
this is not necessary as you've just checked that they are equal, right?

jakobj (approved):
excellent, thanks a lot! :+1:

as always, please reorganize the history before merging.

Issue #225, opened 2020-08-03
Reorder genome to reduce limitations of gene connections


HenrikMettler:
As stated by Goldman et al (https://ieeexplore.ieee.org/document/6815728?arnumber=6815728) the ordering of nodes in cgp adds artificial limitations to node connections. Nodes that could be connected without forming a cycle will never be connected due to their position and adjacent nodes (assuming `n_rows = 1`) can never be connect through an intermediary node. Furthermore the probability of a node being active depends largely on its position in the genome. 
A proposed method to attenuate these problems is called _Reorder_. Roughly speaking this method shuffles the genome, by assigning a new position to a node, once all the nodes it takes input from are placed earlier in the genome. 
Since their results suggest that _Reorder_ helps to reduce the number of required function evaluations (tested on some boolean problems), I propose to add it to our library.

mschmidt87:
Could you explain the two issues ("connections of nodes are artificially restricted", "nodes being active depends on position") in more detail. It is not clear to me why that is the case.

HenrikMettler:
- _Connections of nodes are artificially restriced_ (paraphrasing from the paper): Consider two nodes X,Y which have no dependent relationship, where X preceeds Y. Assume that X taking input from Y would lead to a good solution. Without Reorder X can never take input from Y and would therefore need "node reinvention" of a node identical to Y appearing before X (or the other way around). With Reorder the different orderings can be examined without the need for such a reinvention. 

- _Nodes being active depends on their position_: (For simplicity let's assume `levels_back = n_columns` and `n_outputs =1`) For every node that is behind the node which connects to the output, the probability of being active is 0, as long as the input gene of the output node is not mutated. Whereas for every node on the left, there is still a chance of being connected to this internal node. Generally speaking, the more to the left (closer to the input) a node is the higher its chance of being active via an (or several) indermediates (Remember that all internal nodes have the same probability of being the input of the output node). In the paper they derived the formula for the probability of a gene to be active. Where I think there should be a "+" instead of "." before the factorial.


![Goldmann_equation](https://user-images.githubusercontent.com/32355155/89186263-d184a600-d59b-11ea-878d-7c4d5e2e2be1.png)

mschmidt87:
Thanks for the detailed explanation. That makes sense to me. How would you implement this? As an option for the evolve function?

HenrikMettler:
[...]

PR #239, opened 2020-08-25
Dev/genome reorder

HenrikMettler:
Implements a first solution to #225 
Reorders the nodes in the parents genome(s) before creating offspring, without violating any node dependencies, as suggested in Goldman 2015. Works only with `levels_back = n_columns` and `n_rows = 1` since possible issue with limitations in levels_back and more than 1 row are not accounted for. 
Currently I did not add a new example for this functionality but changed `levels_back = n_columns` in `example_minimal.py` and set the boolean for doing reorder to true. 
Testing currently only covers repeated application of reorder on two examples genomes, while asserting no changes in the resulting sympy expression.

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -25,6 +25,7 @@ class MuPlusLambda:
         n_processes: int = 1,
         local_search: Callable[[IndividualBase], None] = lambda combined: None,
         k_local_search: Union[int, None] = None,
+        reorder_genome: bool = False,
     ):
         """Init function
 
@@ -44,6 +45,13 @@ class MuPlusLambda:
         k_local_search : int
             Number of individuals in the whole population (parents +
             offsprings) to apply local search to.
+       reorder_genome: bool, optional
+            Whether genome reordering should be applied.
+            Reorder shuffles the genotype of an individual without changing its phenotype,
+            thereby contributing to neutral drift through the genotypic search space.
+            If True, reorder is applied to each parents genome at every generation
+            before creating offsprings.
+            Defaults to True.
         """
         self.n_offsprings = n_offsprings
 
@@ -51,6 +59,7 @@ class MuPlusLambda:
         self.n_processes = n_processes
         self.local_search = local_search
         self.k_local_search = k_local_search
+        self.reorder_genome = reorder_genome
 
     def initialize_fitness_parents(
         self, pop: Population, objective: Callable[[IndividualBase], IndividualBase]
@@ -71,7 +80,7 @@ class MuPlusLambda:
         pop._parents = self._compute_fitness(pop.parents, objective)
 
     def step(
-        self, pop: Population, objective: Callable[[IndividualBase], IndividualBase]
+        self, pop: Population, objective: Callable[[IndividualBase], IndividualBase],
     ) -> Population:
         """Perform one step in the evolution.
 
@@ -89,6 +98,10 @@ class MuPlusLambda:
         Population
             Modified population with new parents.
         """
+
+        if self.reorder_genome:
+            pop.reorder_genome()
+
         offsprings = self._create_new_offspring_generation(pop)
 
         # we want to prefer offsprings with the same fitness as their
--- a/cgp/genome.py
+++ b/cgp/genome.py
@@ -1,4 +1,4 @@
-from typing import Dict, Generator, List, Optional, Tuple, Type, Union
+from typing import Dict, Generator, List, Optional, Set, Tuple, Type, Union
 
 import numpy as np
 
@@ -138,7 +138,7 @@ class Genome:
         return s
 
     def determine_permissible_values_per_gene(self, gene_idx: int) -> np.ndarray:
-        region_idx = gene_idx // self._length_per_region
+        region_idx = self._get_region_idx(gene_idx)
 
         if self._is_input_region(region_idx):
             return self._determine_permissible_values_input_region(gene_idx)
@@ -266,6 +266,147 @@ class Genome:
 
         # accept generated dna if it is valid
         self.dna = dna
+
+    def reorder(self, rng: np.random.RandomState) -> None:
+        """Reorder the genome
+
+        Shuffle node ordering of internal (hidden) nodes in genome without changing the phenotype.
+        (Goldman 2015, DOI: 10.1109/TEVC.2014.2324539)
+
+        During reordering, inactive genes, e.g., input genes of nodes with arity zero,
+        are not taken into account and can hence have invalid values after reordering.
+        These invalid values are replaced by random values
+        for the respective gene after reordering.
+
+        Parameters
+        ----------
+        rng : numpy.RandomState
+            Random number generator instance.
+
+        Returns
+        ----------
+        None
+        """
+        if (self._n_rows != 1) or (self._levels_back != self._n_columns):
+            raise ValueError(
+                "Genome reordering is only implemented for n_rows=1" " and levels_back=n_columns"
+            )
+
+        dna = self._dna.copy()
+
+        node_dependencies: Dict[int, Set[int]] = self._determine_node_dependencies()
+
+        addable_nodes: Set[int] = self._get_addable_nodes(node_dependencies)
+
+        new_node_idx: int = self._n_inputs  # First position to be placed is after inputs
+        used_node_indices: List[int] = []
+
+        while len(addable_nodes) > 0:
+
+            old_node_idx = rng.choice(list(addable_nodes))
+
+            dna = self._copy_dna_segment(dna, old_node_idx=old_node_idx, new_node_idx=new_node_idx)
+
+            for dependencies in node_dependencies.values():
+                dependencies.discard(old_node_idx)
+
+            used_node_indices.append(old_node_idx)
+            addable_nodes = self._get_addable_nodes(node_dependencies, used_node_indices)
+            new_node_idx += 1
+
+        self._update_input_genes(dna, used_node_indices)
+        self._replace_invalid_input_alleles(dna, rng)
+
+        self.dna = dna
+
+    def _copy_dna_segment(self, dna: List[int], old_node_idx: int, new_node_idx: int) -> List[int]:
+        """ Copy a nodes dna segment from its old node location to a new location. """
+
+        dna[
+            new_node_idx * self._length_per_region : (new_node_idx + 1) * self._length_per_region
+        ] = self._dna[
+            old_node_idx * self._length_per_region : (old_node_idx + 1) * self._length_per_region
+        ]
+
+        return dna
+
+    def _update_input_genes(self, dna: List[int], used_node_indices: List[int]) -> None:
+        """Update input genes of all nodes from old node indices to new node indices"""
+        for gene_idx, gene_value in enumerate(dna):
+            region_idx = self._get_region_idx(gene_idx)
+            if self._is_hidden_input_gene(gene_idx, region_idx) or self._is_output_input_gene(
+                gene_idx
+            ):
+                if gene_value >= self._n_inputs:
+                    gene_value = self._n_inputs + used_node_indices.index(gene_value)
+            dna[gene_idx] = gene_value
+
+    def _replace_invalid_input_alleles(self, dna: List[int], rng: np.random.RandomState) -> None:
+        """Replace invalid alleles for unused input genes of all nodes
+        by random permissible values.
+        WARNING: Works only if self.n_rows==1.
+        """
+        assert self._n_rows == 1
+
+        for gene_idx, gene_value in enumerate(dna):
+            region_idx = self._get_region_idx(gene_idx)
+            if self._is_hidden_input_gene(gene_idx, region_idx) and gene_value > region_idx:
+                permissible_values = self.determine_permissible_values_per_gene(gene_idx)
+                gene_value = rng.choice(permissible_values)
+                dna[gene_idx] = gene_value
+
+    def _get_addable_nodes(
+        self, node_dependencies: Dict[int, Set[int]], used_node_indices: List[int] = [],
+    ) -> Set[int]:
+        """ Get the set of addable nodes,
+         nodes which have no dependencies and were not already used.
+        """
+        addable_nodes = set(
+            idx for idx, dependencies in node_dependencies.items() if len(dependencies) == 0
+        )
+        return addable_nodes.difference(used_node_indices)
+
+    def _get_region_idx(self, gene_idx: int) -> int:
+
+        return gene_idx // self._length_per_region
+
+    def _determine_node_dependencies(self) -> Dict[int, Set[int]]:
+        """ Determines the set of node indices on which each node depends.
+            Unused input genes are ignored.
+
+        Returns
+        ----
+        dependencies: Dict[int, Set[int]]
+            Dictionary containing for every node the set of active input genes
+
+        """
+        dependencies: Dict[int, Set[int]] = {}
+        for region_idx, _ in self.iter_hidden_regions():
+
+            current_node_dependencies: Set[int] = set()
+
+            operator_idx: int = region_idx * self._length_per_region
+
+            current_arity: int = self._determine_operator_arity(operator_idx)
+
+            for idx_gene in range(
+                1, current_arity + 1
+            ):  # shift by 1 since first gene is the operator gene
+                input_node_idx = self._dna[operator_idx + idx_gene]
+                if not self._is_input_region(
+                    input_node_idx
+                ):  # not necessary to add input regions, since their positions remain fixed
+                    current_node_dependencies.add(input_node_idx)
+
+            dependencies[region_idx] = current_node_dependencies
+
+        return dependencies
+
+    def _determine_operator_arity(self, gene_idx: int) -> int:
+
+        assert self._is_function_gene(gene_idx)
+
+        return self._primitives[self._dna[gene_idx]]._arity
 
     def _permissible_inputs(self, region_idx: int) -> List[int]:
 
@@ -403,6 +544,11 @@ class Genome:
     def _is_hidden_input_gene(self, gene_idx: int, region_idx: int) -> bool:
         return self._is_hidden_region(region_idx) and (not self._is_function_gene(gene_idx))
 
+    def _is_output_input_gene(self, gene_idx: int) -> bool:
+        return (
+            self._is_gene_in_output_region(gene_idx) and gene_idx % self._length_per_region == 1
+        )  # assumes 2nd gene in output is coding for input
+
     def _select_gene_indices_for_mutation(
         self, mutation_rate: float, rng: np.random.RandomState
     ) -> List[int]:
@@ -450,7 +596,7 @@ class Genome:
 
         for (gene_idx, allele) in zip(selected_gene_indices, np.array(dna)[selected_gene_indices]):
 
-            region_idx = gene_idx // self._length_per_region
+            region_idx = self._get_region_idx(gene_idx)
 
             permissible_values = self._permissible_values[gene_idx]
             permissible_alternative_values = permissible_values[permissible_values != allele]
--- a/cgp/hl_api.py
+++ b/cgp/hl_api.py
@@ -45,6 +45,7 @@ def evolve(
         parallel evaluation of the objective is supported. Currently
         not implemented. Defaults to 1.
 
+
     Returns
     -------
     None
--- a/cgp/individual.py
+++ b/cgp/individual.py
@@ -61,6 +61,9 @@ class IndividualBase:
     def randomize_genome(self, rng):
         raise NotImplementedError()
 
+    def reorder_genome(self, rng):
+        raise NotImplementedError()
+
     def to_func(self):
         raise NotImplementedError()
 
@@ -83,6 +86,10 @@ class IndividualBase:
     @staticmethod
     def _randomize_genome(genome: Genome, rng: np.random.RandomState) -> None:
         genome.randomize(rng)
+
+    @staticmethod
+    def _reorder_genome(genome: Genome, rng: np.random.RandomState) -> None:
+        genome.reorder(rng)
 
     @staticmethod
     def _to_func(genome: Genome) -> Callable[[List[float]], List[float]]:
@@ -137,6 +144,9 @@ class IndividualSingleGenome(IndividualBase):
     def randomize_genome(self, rng: np.random.RandomState) -> None:
         self._randomize_genome(self.genome, rng)
 
+    def reorder_genome(self, rng: np.random.RandomState) -> None:
+        self._reorder_genome(self.genome, rng)
+
     def to_func(self) -> Callable[[List[float]], List[float]]:
         return self._to_func(self.genome)
 
@@ -185,6 +195,10 @@ class IndividualMultiGenome(IndividualBase):
     def randomize_genome(self, rng: np.random.RandomState) -> None:
         for g in self.genome:
             self._randomize_genome(g, rng)
+
+    def reorder_genome(self, rng: np.random.RandomState) -> None:
+        for g in self.genome:
+            self._reorder_genome(g, rng)
 
     def to_func(self) -> List[Callable[[List[float]], List[float]]]:
         return [self._to_func(g) for g in self.genome]
--- a/cgp/population.py
+++ b/cgp/population.py
@@ -133,3 +133,13 @@ class Population:
             List of fitness values for all parents.
         """
         return [ind.fitness for ind in self._parents]
+
+    def reorder_genome(self) -> None:
+        """ Reorders the genome for all parents in the population
+
+        Returns
+        ---------
+        None
+        """
+        for parent in self.parents:
+            parent.reorder_genome(self.rng)
--- a/examples/example_reorder.py
+++ b/examples/example_reorder.py
@@ -1,0 +1,161 @@
+"""
+Example for evolutionary regression with genome reordering
+===========================================================
+
+Example demonstrating the effect of genome reordering.
+
+References
+-----------
+Goldman B.W., Punch W.F. (2014): Analysis of Cartesian Genetic Programming’s
+Evolutionary Mechanisms
+DOI: 10.1109/TEVC.2014.2324539
+"""
+
+# The docopt str is added explicitly to ensure compatibility with
+# sphinx-gallery.
+docopt_str = """
+   Usage:
+     example_reorder.py [--max-generations=<N>]
+
+   Options:
+     -h --help
+     --max-generations=<N>  Maximum number of generations [default: 300]
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+import scipy.constants
+from docopt import docopt
+
+import cgp
+
+args = docopt(docopt_str)
+
+# %%
+# We first define a target function.
+
+
+def f_target(x):
+    return x[0] ** 2 + 1.0
+
+
+# %%
+# Then we define the objective function for the evolution. It uses
+# the mean-squared error between the output of the expression
+# represented by a given individual and the target function evaluated
+# on a set of random points.
+
+
+def objective(individual):
+    if individual.fitness is not None:
+        return individual
+
+    n_function_evaluations = 1000
+
+    np.random.seed(1234)
+
+    f = individual.to_func()
+    loss = 0
+    for x in np.random.uniform(-4, 4, n_function_evaluations):
+        # the callable returned from `to_func` accepts and returns
+        # lists; accordingly we need to pack the argument and unpack
+        # the return value
+        y = f([x])[0]
+        loss += (f_target([x]) - y) ** 2
+
+    individual.fitness = -loss / n_function_evaluations
+
+    return individual
+
+
+# %%
+# Next, we set up the evolutionary search. We first define the
+# parameters for the population, the genome of individuals, and two
+# evolutionary algorithms without (default) and with genome reordering.
+
+
+population_params = {"n_parents": 1, "mutation_rate": 0.03, "seed": 818821}
+
+genome_params = {
+    "n_inputs": 1,
+    "n_outputs": 1,
+    "n_columns": 12,
+    "n_rows": 1,
+    "levels_back": None,
+    "primitives": (cgp.Add, cgp.Sub, cgp.Mul, cgp.ConstantFloat),
+}
+
+ea_params = {"n_offsprings": 4, "tournament_size": 2, "n_processes": 2}
+ea_params_with_reorder = {
+    "n_offsprings": 4,
+    "tournament_size": 2,
+    "n_processes": 2,
+    "reorder_genome": True,
+}
+
+evolve_params = {"max_generations": int(args["--max-generations"]), "min_fitness": 0.0}
+
+# %%
+# We create two populations that will be evolved
+pop = cgp.Population(**population_params, genome_params=genome_params)
+pop_with_reorder = cgp.Population(**population_params, genome_params=genome_params)
+
+# %%
+# and two instances of the (mu + lambda) evolutionary algorithm
+ea = cgp.ea.MuPlusLambda(**ea_params)
+ea_with_reorder = cgp.ea.MuPlusLambda(**ea_params_with_reorder)
+
+# %%
+# We define two callbacks for recording of fitness over generations
+history = {}
+history["fitness_champion"] = []
+
+
+def recording_callback(pop):
+    history["fitness_champion"].append(pop.champion.fitness)
+
+
+history_with_reorder = {}
+history_with_reorder["fitness_champion"] = []
+
+
+def recording_callback_with_reorder(pop):
+    history_with_reorder["fitness_champion"].append(pop.champion.fitness)
+
+
+# %%
+# and finally perform the evolution of the two populations
+cgp.evolve(
+    pop, objective, ea, **evolve_params, print_progress=True, callback=recording_callback,
+)
+
+cgp.evolve(
+    pop_with_reorder,
+    objective,
+    ea_with_reorder,
+    **evolve_params,
+    print_progress=True,
+    callback=recording_callback_with_reorder,
+)
+
+
+# %%
+# After finishing the evolution, we plot the evolution of the fittest individual
+# with and without genome reordering
+
+
+width = 9.0
+fig = plt.figure(1, figsize=[width, width / scipy.constants.golden])
+
+plt.plot(history["fitness_champion"], label="Champion")
+plt.plot(history_with_reorder["fitness_champion"], label="Champion with reorder")
+
+plt.xlabel("Generation")
+plt.ylabel("Fitness")
+plt.legend(["Champion", "Champion with reorder"])
+
+plt.yscale("symlog")
+plt.ylim(-1.0e2, 0.1)
+plt.axhline(0.0, color="0.7")
+
+fig.savefig("example_reorder.pdf", dpi=300)
--- a/test/test_genome.py
+++ b/test/test_genome.py
@@ -548,3 +548,102 @@ def test_mutate_does_not_reinitialize_parameters(genome_params, rng, mutation_ra
     assert genome._parameter_names_to_values["<p2>"] == pytest.approx(
         parameter_names_to_values_before["<p2>"]
     )
+
+
+def test_genome_reordering_empirically(rng):
+    # empirically test that reordering does not change the output function of a genome
+
+    pytest.importorskip("sympy")
+
+    genome_params = {
+        "n_inputs": 2,
+        "n_outputs": 1,
+        "n_columns": 10,
+        "n_rows": 1,
+        "levels_back": None,
+        "primitives": (cgp.Mul, cgp.Sub, cgp.Add, cgp.ConstantFloat),
+    }
+
+    genome = cgp.Genome(**genome_params)
+
+    # f(x_0, x_1) = x_0 ** 2 - x_1 + 1
+    dna_fixed = [
+        ID_INPUT_NODE,
+        ID_NON_CODING_GENE,
+        ID_NON_CODING_GENE,
+        ID_INPUT_NODE,
+        ID_NON_CODING_GENE,
+        ID_NON_CODING_GENE,
+        0,  # Mul ->  x_0^2 (address 2)
+        0,  # x
+        0,  # x
+        1,  # Sub -> x_0^2 - x_1 (address 3)
+        2,  # x^2
+        1,  # y
+        1,  # Sub ->  0 (address 4)
+        0,  # x
+        0,  # x
+        3,  # const ->  1 (address 5)
+        2,  # address of x_0^2 (unused)
+        3,  # address of 0 (unused)
+        3,  # const -> outs 1 (address 6)
+        0,
+        0,
+        2,  # Add -> x_0^2 - x_1 + 1 (address 7)
+        3,  # x_0^2 - x_1
+        5,  # 1
+        3,  # const (address 8)
+        0,
+        1,
+        3,  # const (address 9)
+        0,
+        1,
+        3,  # const (address 10)
+        0,
+        1,
+        3,  # const (address 11)
+        0,
+        1,
+        ID_OUTPUT_NODE,
+        7,
+        ID_NON_CODING_GENE,
+    ]
+
+    genome.dna = dna_fixed
+    sympy_expression = cgp.CartesianGraph(genome).to_sympy()
+    n_reorderings = 100
+    for _ in range(n_reorderings):
+        genome.reorder(rng)
+        sympy_expression_after_reorder = cgp.CartesianGraph(genome).to_sympy()
+        assert sympy_expression_after_reorder == sympy_expression
+
+
+def test_genome_reordering_parameterization_consistency(rng):
+
+    genome_params = {
+        "n_inputs": 2,
+        "n_outputs": 1,
+        "n_columns": 10,
+        "n_rows": 2,
+        "levels_back": None,
+        "primitives": (cgp.Mul, cgp.Sub, cgp.Add, cgp.ConstantFloat),
+    }
+
+    genome = cgp.Genome(**genome_params)
+
+    with pytest.raises(ValueError):
+        genome.reorder(rng)
+
+    genome_params = {
+        "n_inputs": 2,
+        "n_outputs": 1,
+        "n_columns": 10,
+        "n_rows": 1,
+        "levels_back": 5,
+        "primitives": (cgp.Mul, cgp.Sub, cgp.Add, cgp.ConstantFloat),
+    }
+
+    genome = cgp.Genome(**genome_params)
+
+    with pytest.raises(ValueError):
+        genome.reorder(rng)


cgp/genome.py
```diff
@@ -267,6 +267,172 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                # to only consider genes which are used in current node
+                operator_idx = region_idx * self._length_per_region
+                operator_value = self._dna[operator_idx]
+                current_arity = self._primitives[operator_value]._arity
+
+                # append each input and its respective dependencies to the dependencies
+                for idx_gene in range(
+                    1, current_arity + 1
+                ):  # shift by 1 since first gene is operator
+                    input_region_idx = self._dna[operator_idx + idx_gene]
```
jakobj, line 409: the dna actually does not store a region index, but a node index, they are just identical. not sure we should mix semantics here :thinking:

cgp/genome.py
```diff
@@ -267,6 +267,172 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                # replace value that is too large by smaller value
+                permissible_values = self.determine_permissible_values_per_gene(gene_idx)
+                gene_value = rng.choice(permissible_values)
+                dna_for_reorder[gene_idx] = gene_value
+
+        self.dna = dna_for_reorder
+
+    def _determine_dependencies(self) -> List[set]:
+
+        dependencies = []
```
jakobj, line 387: since you iterate over the regions in order you can use a list here and the index it with `input_region_idx` below. just to be a bit more transparent, maybe use a dict here, mapping region indices to their dependencies?

cgp/genome.py
```diff
@@ -267,6 +267,172 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                        current_node_dependencies.add(input_region_idx)  # append the input node &
+                        current_node_dependencies.update(
+                            dependencies[input_region_idx]
+                        )  # the dependencies of the input node to the current node dependencies
+                        # Todo: can be removed?
+                dependencies.append(current_node_dependencies)
+
+            elif self._is_output_region(
+                region_idx
+            ):  # Todo: actually it is not necessary to calculate those,
```
jakobj, line 422: judging by your comment, a `continue` would also do here? ;) if so, you could combine it with the `continue` for input regions

cgp/hl_api.py
```diff
@@ -44,7 +45,9 @@ def evolve(
         Number of parallel processes to be used. If greater than 1,
         parallel evaluation of the objective is supported. Currently
         not implemented. Defaults to 1.
-
+    do_reorder: float, optional
```
jakobj, line 48:
```suggestion
    reorder_genome: bool, optional
```

cgp/hl_api.py
```diff
@@ -44,7 +45,9 @@ def evolve(
...
+        Determines whether genome reordering should be applied.
```
jakobj, line 49:
```suggestion
        Whether genome reordering should be applied.
```

cgp/hl_api.py
```diff
@@ -59,6 +62,17 @@ def evolve(
     # Main loop: -1 offset since the last loop iteration will still increase generation by one
     while pop.generation < max_generations - 1:
 
+        # reorder the genome of the parents before creating offspring
```
jakobj, line 65: i feel like this would fit better into the population class rather than the `hl_api`, don't you think so?

cgp/hl_api.py
```diff
@@ -59,6 +62,17 @@ def evolve(
...
+        if do_reorder:
+            reorder_possible = check_if_reorder_possible((pop).parents[0])
+            if reorder_possible:
+                pop.reorder_genome()
+            else:
+                print(
+                    "Reorder the genome not possible. Currently only available if "
+                    "levels_back = n_columns and n_rows = 1"
+                )
```
jakobj, line 74: i would hide this check directly in `reorder_genome`

test/test_genome.py
```diff
@@ -548,3 +548,129 @@ def test_mutate_does_not_reinitialize_parameters(genome_params, rng, mutation_ra
...
+
+    genome.dna = dna_fixed
+    sympy_expression = cgp.CartesianGraph(genome).to_sympy()
+    n_reorderings = 100
+    for _ in range(n_reorderings):
+        genome.reorder(rng)
+        sympy_expression_after_reorder = cgp.CartesianGraph(genome).to_sympy()
+        assert sympy_expression_after_reorder == sympy_expression
+
+    # target: f(x_0,x_1) = x_0 ** 2 - x_1 + 1
```
jakobj, line 616: what's different to the check above? maybe move to separate function

examples/example_minimal.py
```diff
@@ -76,7 +76,7 @@ def objective(individual):
     "n_outputs": 1,
     "n_columns": 12,
     "n_rows": 1,
-    "levels_back": 5,
+    "levels_back": None,
```
jakobj, line 79: please undo this change again before merging

cgp/population.py
```diff
@@ -128,3 +128,13 @@ def fitness_parents(self) -> List[Union[None, float]]:
...
         return [ind.fitness for ind in self._parents]
+
+    def reorder_genome(self) -> None:
+        """ Reorders the genome for all parents in the population
+
+        Returns
+        ---------
+        None
+        """
+        for parent in self.parents:
```
jakobj, line 139: as discussed this reorders the parents forever, rather than only before creating offsprings. it hence adds an additional (neutral) drift through the search space. could you check again please how this is handled in the original paper and (if we keep it like this) make sure to document it appropriately in the doc for the `do_reorder` argument?

jakobj (changes requested):
awesome, thanks for working on this! :1st_place_medal: 

i've added a few inline comments that should be addressed

jakobj:
also: mypy reports issues, please fix them :)

cgp/genome.py
```diff
@@ -267,6 +267,172 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+            current_node_dependencies = set()
+
+            if self._is_input_region(region_idx):
+                dependencies.append(
+                    current_node_dependencies
+                )  # input regions have no dependencies
+
+            elif self._is_hidden_region(
+                region_idx
+            ):  # can assume dependencies is not empty anymore
```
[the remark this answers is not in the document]
HenrikMettler: because first empty sets for the input regions were added to `dependencies`. Anyway this is redundant now, if i change `dependencies` to a dict, as you suggested above

cgp/hl_api.py
HenrikMettler: What do you think about checking in `genome`? And if reorder is not possible, simply pass (maybe with a print or warning)?

cgp/population.py
HenrikMettler: In the paper it says: "Reorder is used once each generation to shuffle the nodes of the parent’s genome". This is not fully explicit, but if you look at the source code: 
https://github.com/brianwgoldman/Analysis-of-CGPs-Mechanisms/blob/master/evolution.py#L591

Reorder is applied to the parent even if the parent is not replaced at the end of a generation.

examples/example_minimal.py
HenrikMettler: Yes, I will do. Should I add an example showing the use of `reorder`?
Or should I integrate it within an existing example (not sure it fits somewhere...)?

test/test_genome.py
HenrikMettler: One genome has only a single input and the other has two. But i agree it might not be necessary to test the same for two similar `genome_params` configurations. Should I remove one of them?

HenrikMettler:
Should I add more tests for the auxiliary functions used in `reorder_genome` or is the general empirical test sufficient?

cgp/genome.py
```diff
@@ -267,6 +267,172 @@ def randomize(self, rng: np.random.RandomState) -> None:
         # accept generated dna if it is valid
         self.dna = dna
 
+    def reorder(self, rng: np.random.RandomState) -> None:
+        """Reorder the genome
+
+        Shuffle node ordering of internal (hidden) nodes in genome without changing node behavior.
+        (Goldman 2015, DOI: 10.1109/TEVC.2014.2324539)
+
+        First calculates a list of sets of dependencies on earlier nodes
```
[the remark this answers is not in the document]
HenrikMettler: I shortened the description but kept a brief overview over how the function operates. Please tell me if this is not necessary or should be improved.

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        From a set of nodes with no dependencies (addable_nodes)
+        a (pseudo-) random node is placed at the first internal position.
+        After updating the set of addable nodes,
+        a (pseudo-) random node for the next position is picked.
+        This procedure is repeated until all nodes are placed.
+
+        Gene values are updated to correspond to the new address of the corresponding node.
+
+        Invalid gene values (from unused arities of eg. constant operator nodes)
+        are replaced by (pseudo-) random picks of permissible values for that position.
```
jakobj, line 285:
```suggestion
        During reordering, inactive genes, e.g., input genes of nodes with arity zero, are not taken
        into account and can hence have invalid values after reordering.
        These invalid values are replaced by random values for the respective gene after reordering.
```
it's important to note this, but the rest of the explanation can be removed i think

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        ----------
+        rng : numpy.RandomState
+            Random number generator instance.
+
+        Returns
+        ----------
+        None
+        """
+        if not self._n_rows == 1:
+            Warning("n_rows must be 1, reorder is not applied")
```
jakobj, line 297: shouldn't this result in an exception? if this happens the user should fix their config, right?

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                dependencies, used_idx_list, idx_used=current_idx
+            )
+            node_idx += 1
+
+        self._update_gene_values(dna_for_reorder, used_idx_list)
+        self._replace_invalid_gene_values(dna_for_reorder, rng)
+
+        self.dna = dna_for_reorder
+
+    def _exchange_dna_segments(
```
jakobj, line 333: this is not so much an exchange as a `copy`, isn't it?

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        self, dna_new: List[int], idx_old_location: int, idx_new_location: int
```
jakobj, line 334:
```suggestion
        self, dna: List[int], idx_old_location: int, idx_new_location: int
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        return dna_new
+
+    def _update_addable_nodes(
+        self, dependencies: Dict, used_idx_list: List[int], idx_used: int
+    ) -> set:
+        """ Update the set of addable nodes,
+        by reevaluating which nodes haves no dependencies
+        and remove nodes which were already placed from the list.
+        """
+        addable_nodes = set(int(idx_str) for idx_str, item in dependencies.items() if not item)
```
jakobj, line 358: you use this construction twice, maybe refactor into small helper function

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                for idx_gene in range(
+                    1, current_arity + 1
+                ):  # shift by 1 since first gene is the operator gene
+                    input_node_idx = self._dna[operator_idx + idx_gene]
+                    if not self._is_input_region(
+                        input_node_idx
+                    ):  # not necessary to add input regions, since they are first anyway
+                        current_node_dependencies.add(input_node_idx)  # append the input node &
+
+                dependencies[str(region_idx)] = current_node_dependencies
```
jakobj, line 432: why `str`?

cgp/hl_api.py
jakobj: i'm not sure i understand your question? i would just remove all traces of `reorder` from this function and rather manage this decision in the population class

test/test_genome.py
jakobj: yeah in that case i would just keep the more general test case

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        if not self._n_rows == 1:
+            Warning("n_rows must be 1, reorder is not applied")
+            return
+        if not self._levels_back == self._n_columns:
+            Warning("levels_back must be equal n_columns, reorder is not applied")
+            return
+
+        dna_for_reorder = self._dna.copy()
+
+        dependencies = self._determine_dependencies()
```
jakobj, line 305:
```suggestion
        node_dependencies = self._determine_node_dependencies()
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        addable_nodes = set(int(idx_str) for idx_str, item in dependencies.items() if not item)
```
jakobj, line 307:
```suggestion
        addable_nodes = set(idx for idx, dependencies in node_dependencies.items() if len(dependencies) == 0)
```
i would be explicit here that we check the length of the dependencies, but it's a matter of taste

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        node_idx = self._n_inputs  # First position to be placed is after inputs
+        used_idx_list: List = []
+
+        while len(addable_nodes) > 0:
+            # pick one of the addable nodes
+            current_idx = rng.choice(list(addable_nodes))
```
jakobj, line 314:
```suggestion
            current_node_idx = rng.choice(addable_nodes)
```
do you really need this conversion?

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        dependencies = self._determine_dependencies()
+
+        addable_nodes = set(int(idx_str) for idx_str, item in dependencies.items() if not item)
+
+        node_idx = self._n_inputs  # First position to be placed is after inputs
+        used_idx_list: List = []
+
+        while len(addable_nodes) > 0:
+            # pick one of the addable nodes
```
jakobj, line 313:
```suggestion
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+            return
+
+        dna_for_reorder = self._dna.copy()
+
+        dependencies = self._determine_dependencies()
+
+        addable_nodes = set(int(idx_str) for idx_str, item in dependencies.items() if not item)
+
+        node_idx = self._n_inputs  # First position to be placed is after inputs
+        used_idx_list: List = []
```
jakobj, line 310:
```suggestion
        used_node_indices: List[int] = []
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        None
+        """
+        if not self._n_rows == 1:
+            Warning("n_rows must be 1, reorder is not applied")
+            return
+        if not self._levels_back == self._n_columns:
+            Warning("levels_back must be equal n_columns, reorder is not applied")
+            return
+
+        dna_for_reorder = self._dna.copy()
```
jakobj, line 303:
```suggestion
        dna = self._dna.copy()
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        while len(addable_nodes) > 0:
+            # pick one of the addable nodes
+            current_idx = rng.choice(list(addable_nodes))
+
+            dna_for_reorder = self._exchange_dna_segments(
+                dna_for_reorder, idx_old_location=current_idx, idx_new_location=node_idx
+            )
+
+            for _, current_dependencies in dependencies.items():
```
jakobj, line 320:
```suggestion
            for dependencies in node_dependencies.values():
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        dna_new[
+            idx_new_location
+            * self._length_per_region : (idx_new_location + 1)
+            * self._length_per_region
+        ] = node_dna
+
+        return dna_new
+
+    def _update_addable_nodes(
+        self, dependencies: Dict, used_idx_list: List[int], idx_used: int
```
jakobj, line 352:
```suggestion
        self, node_dependencies: Dict, used_node_indices: List[int], current_idx: int
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+    ) -> set:
+        """ Update the set of addable nodes,
+        by reevaluating which nodes haves no dependencies
+        and remove nodes which were already placed from the list.
+        """
+        addable_nodes = set(int(idx_str) for idx_str, item in dependencies.items() if not item)
+        used_idx_list.append(idx_used)
+        addable_nodes = addable_nodes.difference(used_idx_list)
+
+        return addable_nodes
```
jakobj, line 362:
```suggestion
        return addable_nodes, used_idx_list
```
since you modify both of these lists, i would return both of them. i'm not sure what the most pythonic way is though. @mschmidt87 any input?

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+    def _update_gene_values(self, dna_for_reorder: List[int], used_idx_list: List[int]) -> None:
```
jakobj, line 364:
```suggestion
    def _update_gene_values(self, dna: List[int], used_node_indices: List[int]) -> None:
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        """ Update gene values to correspond to the new position of the corresponding node"""
+        for gene_idx, gene_value in enumerate(dna_for_reorder):
+            region_idx = gene_idx // self._length_per_region
```
jakobj, line 367: this operation appears often, doesn't it? maybe this should also be a helper function

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+            if self._is_hidden_input_gene(gene_idx, region_idx) or self._is_output_input_gene(
+                gene_idx
+            ):
+                try:
+                    gene_value = self._n_inputs + used_idx_list.index(gene_value)
+                except ValueError:  # means the input is from an input node -> do nothing
+                    pass
```
jakobj, line 374:
```suggestion
            if gene_value >= self._n_inputs:
                    gene_value = self._n_inputs + used_idx_list.index(gene_value)
```
isn't this clearer? if so, maybe we should have a helper function for this? `is_input_node`?

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+    def _replace_invalid_gene_values(
+        self, dna_for_reorder: List[int], rng: np.random.RandomState
+    ) -> None:
+        """ Replace gene values of unused arities
+        by (pseudo-) random picks of permissible values for that node
+        """
+        for gene_idx, gene_value in enumerate(dna_for_reorder):
+            region_idx = gene_idx // self._length_per_region
+            if self._is_hidden_input_gene(gene_idx, region_idx) and gene_value > region_idx:
```
jakobj, line 385: maybe add comment that this logic only work for `n_rows = 1`?

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                # replace value that is too large by smaller value
```
jakobj, line 386:
```suggestion
                # replace value that is too large by valid value
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                permissible_values = self.determine_permissible_values_per_gene(gene_idx)
+                gene_value = rng.choice(permissible_values)
+                dna_for_reorder[gene_idx] = gene_value
+
+    def _determine_dependencies(self) -> Dict:
```
jakobj, line 391:
```suggestion
    def _determine_node_dependencies(self) -> Dict:
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+                # append each input to the dependencies
+                for idx_gene in range(
+                    1, current_arity + 1
+                ):  # shift by 1 since first gene is the operator gene
+                    input_node_idx = self._dna[operator_idx + idx_gene]
+                    if not self._is_input_region(
+                        input_node_idx
+                    ):  # not necessary to add input regions, since they are first anyway
+                        current_node_dependencies.add(input_node_idx)  # append the input node &
```
jakobj, line 430:
```suggestion
                        current_node_dependencies.add(input_node_idx)
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                continue
+
+            elif self._is_hidden_region(region_idx):
+
+                operator_idx = region_idx * self._length_per_region
+
+                # only consider genes which are used in current node
+                current_arity = self._determine_operator_arity(operator_idx)
+
+                # append each input to the dependencies
```
jakobj, line 422:
```suggestion
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+            current_node_dependencies: set = set()
+
+            if self._is_input_region(region_idx) or self._is_output_region(region_idx):
+                continue
+
+            elif self._is_hidden_region(region_idx):
+
+                operator_idx = region_idx * self._length_per_region
+
+                # only consider genes which are used in current node
```
jakobj, line 419:
```suggestion
```

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                        current_node_dependencies.add(input_node_idx)  # append the input node &
+
+                dependencies[str(region_idx)] = current_node_dependencies
+
+            else:
+                assert False  # should never be reached
+
+        return dependencies
+
+    def _determine_operator_arity(self, operator_idx):
```
jakobj, line 439: missing type annotation

cgp/genome.py
```diff
@@ -267,6 +267,181 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        operator_value = self._dna[operator_idx]
+        arity = self._primitives[operator_value]._arity
+        return arity
```
jakobj, line 443:
```suggestion
        return self._primitives[operator_value]._arity
```

cgp/hl_api.py
```diff
@@ -44,7 +45,12 @@ def evolve(
         Number of parallel processes to be used. If greater than 1,
         parallel evaluation of the objective is supported. Currently
         not implemented. Defaults to 1.
-
+    reorder_genome: bool, optional
+        Whether genome reordering should be applied.
+        If True, reorder is applied to the parents (genome) before creating offspring.
+        Since the parents genome changes to a reordered genome at every generation,
+        this creates a neutral drift through the search space.
```
jakobj, line 52:
```suggestion
        If True, reorder is applied to the parents genomes at every generation before creating offspring. Reorder randomizes the genotype of an individual without changing the phenotype.
```

jakobj: i remove the "neutral drift" part not bc it's wrong, but i think it's not really helpful here. maybe this should be added somewhere in the documentation, maybe of the population class?

jakobj (changes requested):
awesome, this is already much easier to understand, isn't it? :ok_hand: 

i've added a few more inline comment

cgp/genome.py
HenrikMettler: i thought only `str` could be used as keys in dict. Removed it.

cgp/genome.py
HenrikMettler: yes, `rng.choice` can't be applied to sets (https://numpy.org/doc/stable/reference/random/generated/numpy.random.RandomState.choice.html)

cgp/hl_api.py
HenrikMettler: Added a comment in `Population.reorder_genome`

cgp/hl_api.py
HenrikMettler: Ah sorry, I thought this comment was referring to the check if reorder is possible.

cgp/genome.py
jakobj: hmm, if that use case arises we can implement it. i wouldn't consider it for now

cgp/genome.py
```diff
@@ -267,6 +267,183 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+            Warning("levels_back must be equal n_columns, reorder is not applied")
+            return
+
+        dna = self._dna.copy()
+
+        node_dependencies = self._determine_node_dependencies()
+
+        addable_nodes = self._get_addable_nodes(node_dependencies)
+
+        node_idx = self._n_inputs  # First position to be placed is after inputs
```
jakobj, line 303:
```suggestion
        new_node_idx = self._n_inputs  # First position to be placed is after inputs
```

cgp/genome.py
```diff
@@ -267,6 +267,183 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        used_node_indices: List[int] = []
+
+        while len(addable_nodes) > 0:
+
+            current_node_idx = rng.choice(list(addable_nodes))
```
jakobj, line 308:
```suggestion
            old_node_idx = rng.choice(list(addable_nodes))
```

cgp/genome.py
```diff
@@ -267,6 +267,183 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+            dna = self._copy_dna_segment(
+                dna, idx_old_location=current_node_idx, idx_new_location=node_idx
+            )
+
+            for dependencies in node_dependencies.values():
+                dependencies.discard(current_node_idx)
+
+            [addable_nodes, used_node_indices] = self._update_addable_nodes(
```
jakobj, line 317:
```suggestion
            addable_nodes, used_node_indices = self._update_addable_nodes(
```
i think these are not needed

cgp/genome.py
```diff
@@ -267,6 +267,183 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        self._update_gene_values(dna, used_node_indices)
+        self._replace_invalid_gene_values(dna, rng)
+
+        self.dna = dna
+
+    def _copy_dna_segment(
+        self, dna: List[int], idx_old_location: int, idx_new_location: int
+    ) -> List[int]:
+        """ Moves a nodes dna from its old node location to a new location. """
```
jakobj, line 330:
```suggestion
        """ Copy a nodes dna from its old node location to a new location. """
```

cgp/genome.py
```diff
@@ -267,6 +267,184 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        ] = node_dna
+
+        return dna
+
+    def _update_addable_nodes(
+        self, node_dependencies: Dict, used_node_indices: List[int], current_idx: int
+    ) -> [set, List[int]]:
+        """ Update the set of addable nodes,
+        by reevaluating which nodes haves no dependencies
+        and remove nodes which were already placed from the list.
```
jakobj, line 350:
```suggestion
        and remove nodes which were already placed from the dependencies.
```
?

cgp/genome.py
```diff
@@ -267,6 +267,184 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+    def _replace_invalid_gene_values(
+        self, dna_for_reorder: List[int], rng: np.random.RandomState
+    ) -> None:
+        """ Replace gene values of unused arities
+        by (pseudo-) random picks of permissible values for that node.
+        Works only in self.n_rows==1
+        """
+        if not self._n_rows == 1:
+            raise ValueError("Replacing invalid gene values only implemented for n_rows = 1")
```
jakobj, line 377: isn't this already caught at a higher level? would this ever evaluate to true? if it's just for reminding us that this holds only for `n_rows = 1`, then i would prefer just a comment in the code

cgp/genome.py
```diff
@@ -267,6 +267,184 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        for gene_idx, gene_value in enumerate(dna_for_reorder):
+            region_idx = self._get_region_idx(gene_idx)
+            if self._is_hidden_input_gene(gene_idx, region_idx) and gene_value > region_idx:
+                # replace value that is too large by valid value
+                permissible_values = self.determine_permissible_values_per_gene(gene_idx)
+                gene_value = rng.choice(permissible_values)
+                dna_for_reorder[gene_idx] = gene_value
+
+    def _get_addable_nodes(self, node_dependencies: Dict) -> set:
```
jakobj, line 387:
```suggestion
    def _get_addable_nodes(self, node_dependencies: Dict[int,List[int]]) -> Set[int]:
```
i'm not sure about the dict annotation, you should check, but i think it would be good to specify the types as detailed as possible

cgp/genome.py
```diff
@@ -267,6 +267,184 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        return set(
+            idx for idx, dependencies in node_dependencies.items() if len(dependencies) == 0
+        )
+
+    def _get_region_idx(self, gene_idx):
```
jakobj, line 393: missing type annotation

cgp/genome.py
```diff
@@ -267,6 +267,184 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        return gene_idx // self._length_per_region
+
+    def _determine_node_dependencies(self) -> Dict:
```
jakobj, line 397: more detailed type annotation please

cgp/genome.py
```diff
@@ -267,6 +267,184 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        """ Determines for every node a set of dependencies.
+
+            The set of dependencies for a node is given by the indices of the nodes,
+            which explicitly must occur earlier in the genome.
```
jakobj, line 401:
```suggestion
            The set of dependencies for a node is given by the indices of the nodes,
            on which its inputs depend.
```

cgp/genome.py
```diff
@@ -267,6 +267,184 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        Returns
+        ----
+        dependencies: Dict
+            Dictionary containing for every node a set of internal nodes dependencies
+
+        """
+        dependencies = {}
+        for region_idx in range(self._n_regions):
+            current_node_dependencies: set = set()
```
jakobj, line 416: you could move this in the `elif` block

jakobj (changes requested):
nice, we're getting there :) some more inline suggestions

examples/example_minimal.py
jakobj: i think it would be good idea to provide a separate example, maybe comparing the time-to-solution with and without reordering

mschmidt87:
Are you still working on this, @HenrikMettler ?

HenrikMettler:
Yes, was on holiday last week and this week I have some other priorities. Will probably wrap this up next week.

examples/example_minimal.py
HenrikMettler: I added an example which compares reordered and non-reordered history on the same target function as in minimal example. Please tell me if this is suited.

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        by reevaluating which nodes haves no dependencies
+        and remove nodes which were already placed from the dependencies.
+        """
+        addable_nodes = self._get_addable_nodes(node_dependencies)
+        used_node_indices.append(current_idx)
+        addable_nodes = addable_nodes.difference(used_node_indices)
+
+        return addable_nodes, used_node_indices
+
+    def _update_gene_values(self, dna: List[int], used_node_indices: List[int]) -> None:
```
jakobj, line 354:
```suggestion
    def _update_input_genes(self, dna: List[int], used_node_indices: List[int]) -> None:
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        """ Update gene values to correspond to the new position of the corresponding node"""
```
jakobj, line 355:
```suggestion
        """Update input genes of all nodes from old node indices to new node indices"""
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        for gene_idx, gene_value in enumerate(dna):
+            region_idx = self._get_region_idx(gene_idx)
+            if self._is_hidden_input_gene(gene_idx, region_idx) or self._is_output_input_gene(
+                gene_idx
+            ):
+                if gene_value >= self._n_inputs:
+                    gene_value = self._n_inputs + used_node_indices.index(gene_value)
+            dna[gene_idx] = gene_value
+
+    def _replace_invalid_gene_values(
```
jakobj, line 365:
```suggestion
    def _replace_invalid_input_genes(
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        self, dna_for_reorder: List[int], rng: np.random.RandomState
+    ) -> None:
+        """ Replace gene values of unused arities
+        by (pseudo-) random picks of permissible values for that node.
+        Works only if self.n_rows==1
+        """
```
jakobj, line 371:
```suggestion
        """Replace invalid alleles for unused input genes of all nodes by random permissible values.
        WARNING: Works only if self.n_rows==1.
        """
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
```
jakobj, line 372:
```suggestion
        assert self.n_rows == 1
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+            region_idx = self._get_region_idx(gene_idx)
+            if self._is_hidden_input_gene(gene_idx, region_idx) or self._is_output_input_gene(
+                gene_idx
+            ):
+                if gene_value >= self._n_inputs:
+                    gene_value = self._n_inputs + used_node_indices.index(gene_value)
+            dna[gene_idx] = gene_value
+
+    def _replace_invalid_gene_values(
+        self, dna_for_reorder: List[int], rng: np.random.RandomState
```
jakobj, line 366:
```suggestion
        self, dna: List[int], rng: np.random.RandomState
```

cgp/genome.py
jakobj: does the warning only apply to this or also the previous function?

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                # replace value that is too large by valid value
+                permissible_values = self.determine_permissible_values_per_gene(gene_idx)
+                gene_value = rng.choice(permissible_values)
+                dna_for_reorder[gene_idx] = gene_value
+
+    def _get_addable_nodes(self, node_dependencies: Dict[int, set]) -> Set[int]:
+
+        return set(
+            idx for idx, dependencies in node_dependencies.items() if len(dependencies) == 0
+        )
```
jakobj, line 385: :heart_eyes:

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+    def _get_region_idx(self, gene_idx: int) -> int:
+
+        return gene_idx // self._length_per_region
+
+    def _determine_node_dependencies(self) -> Dict[int, set]:
+        """ Determines for every node a set of dependencies.
```
jakobj, line 392:
```suggestion
        """ Determines the set of node indices on which each node depends.
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+            The set of dependencies for a node is given by the indices of the nodes,
+            on which its inputs depend.
```
jakobj, line 395:
```suggestion
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+            The arity of the operator gene is taken into account, only nodes which
+            are used by the operator are seen as dependency.
```
jakobj, line 397:
```suggestion
            Unused input genes are ignored.
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+            Implicit dependencies are not added, since they are covered via the
+            set of explicit dependencies of the respective node.
```
jakobj, line 399:
```suggestion
```
i'm not sure we need this explanation here, do we?

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+
+        Returns
+        ----
+        dependencies: Dict
```
jakobj, line 404:
```suggestion
        dependencies: Dict[int, Set[int]]
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        return set(
+            idx for idx, dependencies in node_dependencies.items() if len(dependencies) == 0
+        )
+
+    def _get_region_idx(self, gene_idx: int) -> int:
+
+        return gene_idx // self._length_per_region
+
+    def _determine_node_dependencies(self) -> Dict[int, set]:
```
jakobj, line 391:
```suggestion
    def _determine_node_dependencies(self) -> Dict[int, Set[int]]:
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+            The arity of the operator gene is taken into account, only nodes which
+            are used by the operator are seen as dependency.
+            Implicit dependencies are not added, since they are covered via the
+            set of explicit dependencies of the respective node.
+
+
+        Returns
+        ----
+        dependencies: Dict
+            Dictionary containing for every node a set of internal nodes dependencies
```
jakobj, line 405:
```suggestion
            Dictionary containing for every node the set of active input genes.
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+        """
+        dependencies = {}
```
jakobj, line 408:
```suggestion
        dependencies: Dict[int, Set[int]] = {}
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        for region_idx in range(self._n_regions):
+
+            if self._is_input_region(region_idx) or self._is_output_region(region_idx):
+                continue
+
+            elif self._is_hidden_region(region_idx):
```
jakobj, line 414: couldn't this be simplified by using the `_iter_hidden_regions` iterator?

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+                current_node_dependencies: set = set()
```
jakobj, line 416:
```suggestion
                current_node_dependencies: Set[int] = set()
```

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+            else:
+                assert False  # should never be reached
+
+        return dependencies
+
+    def _determine_operator_arity(self, operator_idx: int) -> int:
+
+        operator_value = self._dna[operator_idx]
+        return self._primitives[operator_value]._arity
```
jakobj, line 441:
```suggestion
    def _determine_operator_arity(self, gene_idx: int) -> int:
        assert self._is_function_gene(gene_idx)
        return self._primitives[self._dna[gene_idx]]._arity
```

cgp/genome.py
```diff
@@ -403,6 +576,11 @@ def _is_function_gene(self, gene_idx: int) -> bool:
     def _is_hidden_input_gene(self, gene_idx: int, region_idx: int) -> bool:
         return self._is_hidden_region(region_idx) and (not self._is_function_gene(gene_idx))
 
+    def _is_output_input_gene(self, gene_idx: int) -> bool:
+        return (
+            self._is_gene_in_output_region(gene_idx) and gene_idx % self._length_per_region == 1
+        )  # assumes 2nd gene in output is address
```
jakobj, line 582:
```suggestion
        )  # assumes 2nd gene in output is coding for input
```

cgp/hl_api.py
```diff
@@ -59,6 +64,9 @@ def evolve(
     # Main loop: -1 offset since the last loop iteration will still increase generation by one
     while pop.generation < max_generations - 1:
 
+        if reorder_genome:
+            pop.reorder_genome()
```
jakobj, line 68: i feel like this call should be done from `MuPlusLambda.step` right before creating a new offspring generation. seems more natural to me than here.

cgp/individual.py
```diff
@@ -67,6 +70,10 @@ def _mutate_genome(genome: Genome, mutation_rate: float, rng: np.random.RandomSt
     def _randomize_genome(genome: Genome, rng: np.random.RandomState) -> None:
         genome.randomize(rng)
 
+    @staticmethod
+    def _reorder_genome(genome: Genome, rng: np.random.RandomState):
```
jakobj, line 74:
```suggestion
    def _reorder_genome(genome: Genome, rng: np.random.RandomState) -> None:
```

cgp/population.py
```diff
@@ -128,3 +128,16 @@ def fitness_parents(self) -> List[Union[None, float]]:
             List of fitness values for all parents.
         """
         return [ind.fitness for ind in self._parents]
+
+    def reorder_genome(self) -> None:
+        """ Reorders the genome for all parents in the population
+
+        Creates a neutral drift though the search space,
+        by reordering parents and not offspring.
```
jakobj, line 136: i'm not sure we should add that information here. in general, reordering should be documented at one point, not in different parts, like `hl_genome` and `Population`

examples/example_reorder.py
```diff
@@ -0,0 +1,156 @@
+"""
+Example for evolutionary regression with genome reordering
+===========================================================
+
+Example demonstrating the shuffling of internal node ordering in genome
+and comparison of performance to non-reordered genome.
```
jakobj, line 6:
```suggestion
Example demonstrating the effect of genome reordering.
```

test/test_genome.py
```diff
@@ -548,3 +548,72 @@ def test_mutate_does_not_reinitialize_parameters(genome_params, rng, mutation_ra
...
         parameter_names_to_values_before["<p2>"]
     )
+
+
+def test_genome_reordering_empirically(rng):
+    # empirically test that reordering doesn't change the output function of a genome
+
+    pytest.importorskip("sympy")
+
+    # target: f(x_0,x_1) = x_0 ** 2 - x_1 + 1
```
jakobj, line 558:
```suggestion
```

test/test_genome.py
```diff
@@ -548,3 +548,72 @@ def test_mutate_does_not_reinitialize_parameters(genome_params, rng, mutation_ra
     assert genome._parameter_names_to_values["<p2>"] == pytest.approx(
         parameter_names_to_values_before["<p2>"]
     )
+
+
+def test_genome_reordering_empirically(rng):
+    # empirically test that reordering doesn't change the output function of a genome
```
jakobj, line 554:
```suggestion
    # empirically test that reordering does not change the output function of a genome
```

test/test_genome.py
```diff
@@ -548,3 +548,72 @@ def test_mutate_does_not_reinitialize_parameters(genome_params, rng, mutation_ra
...
+        "n_outputs": 1,
+        "n_columns": 10,
+        "n_rows": 1,
+        "levels_back": None,
+        "primitives": (cgp.Mul, cgp.Sub, cgp.Add, cgp.ConstantFloat),
+    }
+
+    genome = cgp.Genome(**genome_params)
+
+    dna_fixed = [
```
jakobj, line 571:
```suggestion
    # f(x_0, x_1) = x_0 ** 2 - x_1 + 1
    dna_fixed = [
```

test/test_genome.py
```diff
@@ -548,3 +548,72 @@ def test_mutate_does_not_reinitialize_parameters(genome_params, rng, mutation_ra
...
+        ID_INPUT_NODE,
+        ID_NON_CODING_GENE,
+        ID_NON_CODING_GENE,
+        ID_INPUT_NODE,
+        ID_NON_CODING_GENE,
+        ID_NON_CODING_GENE,
+        0,  # Mul -> outs x_0² (address 2)
```
jakobj, line 578:
```suggestion
        0,  # Mul -> x_0^2 (address 2)
```
please use `^` instead of actual superscripts ;)

test/test_genome.py
```diff
@@ -548,3 +548,72 @@ def test_mutate_does_not_reinitialize_parameters(genome_params, rng, mutation_ra
...
+        0,  # x
+        0,  # x
+        1,  # Sub -> x_0² - x_1 (address 3)
+        2,  # x²
+        1,  # y
+        1,  # Sub -> outs 0 (address 4)
```
jakobj, line 584:
```suggestion
        1,  # Sub -> 0 (address 4)
```
etc.

test/test_genome.py
```diff
@@ -548,3 +548,72 @@ def test_mutate_does_not_reinitialize_parameters(genome_params, rng, mutation_ra
...
+        3,  # const (address 11)
+        0,
+        1,
+        ID_OUTPUT_NODE,
+        7,
+        ID_NON_CODING_GENE,
+    ]
+
+    genome.dna = dna_fixed
+    sympy_expression = cgp.CartesianGraph(genome).to_sympy()
```
jakobj, line 614: maybe use `simplify = False` in `to_sympy` to make sure it's really the same expression not just after simplifying it with sympy?

test/test_genome.py
```diff
@@ -548,3 +548,72 @@ def test_mutate_does_not_reinitialize_parameters(genome_params, rng, mutation_ra
...
+    n_reorderings = 100
```
jakobj, line 615: if it's not taking too long you can increase this number to 1000, just to be a bit safer

jakobj (changes requested):
awesome work! :1st_place_medal:

i've added some inline comments that should be addressed before merging

cgp/genome.py
```diff
@@ -267,6 +267,179 @@ def randomize(self, rng: np.random.RandomState) -> None:
         # accept generated dna if it is valid
         self.dna = dna
 
+    def reorder(self, rng: np.random.RandomState) -> None:
+        """Reorder the genome
+
+        Shuffle node ordering of internal (hidden) nodes in genome without changing node behavior.
```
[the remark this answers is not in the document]
HenrikMettler: Thanks, this is much better. Node behavior was pretty much "Goldman-wording"

cgp/genome.py
HenrikMettler: Not sure what you mean. It also applies to `reorder` in general, but not to any of the internal subfunctions. I added the warning here again, because one  might be tempted to reuse this function within genome.py

cgp/genome.py
HenrikMettler: :blush:

cgp/genome.py
HenrikMettler: Thanks! This description is so much better like this!

cgp/genome.py
HenrikMettler: Yes, makes it much simpler to understand!

cgp/hl_api.py
HenrikMettler: I agree, moved it there

cgp/population.py
HenrikMettler: I think you are right.  I moved the hl-description into `MuPlusLambda.step` and the details are still in `genome.reorder`, please tell me if think this is appropriate.

test/test_genome.py
HenrikMettler: with `n_reorderings = 1000` all tests in `test_genome.py` take ~4x as long (18s vs 4s on my laptop). Isn't this a bit too much?

test/test_genome.py
HenrikMettler: The problem is that sympy sometimes treats `x_0*x_0` as `x_0^2` even without simplifying the expression.

test/test_genome.py
jakobj: yes, way too long. keep it at 100 :)

test/test_genome.py
jakobj: ouch. ok, let's keep it like this for now.

cgp/genome.py
jakobj: i was wondering whether this comment applies also to `_update_input_genes` since it looked similar. i think you answer cleared it up, thanks :)

test/test_genome.py
```diff
@@ -548,3 +548,71 @@ def test_mutate_does_not_reinitialize_parameters(genome_params, rng, mutation_ra
     assert genome._parameter_names_to_values["<p2>"] == pytest.approx(
         parameter_names_to_values_before["<p2>"]
     )
+
+
+def test_genome_reordering_empirically(rng):
```
jakobj, line 553: just to be safe, can you add tests that check that an error is raised when one is trying to apply reorder to a genome where `n_rows != 1` and/or `n_levels_back != n_columns`? you can do that with the `pytest.raises(...)` context manager

cgp/genome.py
```diff
@@ -267,6 +267,162 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        used_node_indices: List[int],
+        old_node_idx: int,
+    ) -> Tuple[set, List[int]]:
+        """ Update the set of addable nodes,
+        by reevaluating which nodes haves no dependencies
+        and remove nodes which were already placed from the dependencies.
+        """
+        addable_nodes = self._get_addable_nodes(node_dependencies)
+        used_node_indices.append(old_node_idx)
+        addable_nodes = addable_nodes.difference(used_node_indices)
```
jakobj, line 346: logically speaking, the only thing that actually happens here is that `old_node_idx` is removed from the addable nodes, right? bc all other entries in used_node_indices were already removed before? if this is the case, maybe we can simplify the code to express this logic?

cgp/genome.py
```diff
@@ -267,6 +267,162 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        for gene_idx, gene_value in enumerate(dna):
+            region_idx = self._get_region_idx(gene_idx)
+            if self._is_hidden_input_gene(gene_idx, region_idx) or self._is_output_input_gene(
+                gene_idx
+            ):
+                if gene_value >= self._n_inputs:
+                    gene_value = self._n_inputs + used_node_indices.index(gene_value)
+            dna[gene_idx] = gene_value
+
+    def _replace_invalid_input_gene(self, dna: List[int], rng: np.random.RandomState) -> None:
```
jakobj, line 361:
```suggestion
    def _replace_invalid_input_alleles(self, dna: List[int], rng: np.random.RandomState) -> None:
```

cgp/genome.py
```diff
@@ -267,6 +267,162 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        """Replace invalid alleles for unused input genes of all nodes
+        by random permissible values.
+        WARNING: Works only if self.n_rows==1.
+        """
+        assert self._n_rows == 1
+
+        for gene_idx, gene_value in enumerate(dna):
+            region_idx = self._get_region_idx(gene_idx)
+            if self._is_hidden_input_gene(gene_idx, region_idx) and gene_value > region_idx:
+                # replace value that is too large by valid value
```
jakobj, line 371:
```suggestion
```
i think the code is explicit enough. if not, adapt the code ;)

cgp/genome.py
```diff
@@ -267,6 +267,162 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+                permissible_values = self.determine_permissible_values_per_gene(gene_idx)
+                gene_value = rng.choice(permissible_values)
+                dna[gene_idx] = gene_value
+
+    def _get_addable_nodes(self, node_dependencies: Dict[int, set]) -> Set[int]:
```
jakobj, line 376:
```suggestion
    def _get_addable_nodes(self, node_dependencies: Dict[int, Set[int]]) -> Set[int]:
```

cgp/genome.py
```diff
@@ -267,6 +267,162 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+    def _get_region_idx(self, gene_idx: int) -> int:
+
+        return gene_idx // self._length_per_region
+
+    def _determine_node_dependencies(self) -> Dict[int, Set[int]]:
+        """ Determines the set of node indices on which each node depends.
+
+            Unused input genes are ignored.
+
```
jakobj, line 390:
```suggestion
```

cgp/genome.py
```diff
@@ -267,6 +267,162 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+        )
+
+    def _get_region_idx(self, gene_idx: int) -> int:
+
+        return gene_idx // self._length_per_region
+
+    def _determine_node_dependencies(self) -> Dict[int, Set[int]]:
+        """ Determines the set of node indices on which each node depends.
+
+            Unused input genes are ignored.
```
jakobj, line 389:
```suggestion
        """ Determines the set of node indices on which each node depends. Unused input genes are ignored.
```

cgp/hl_api.py
```diff
@@ -16,6 +16,7 @@ def evolve(
     print_progress: Optional[bool] = False,
     callback: Optional[Callable[[Population], None]] = None,
     n_processes: int = 1,
+    reorder_genome: bool = False,
```
jakobj, line 19:
```suggestion
```

cgp/hl_api.py
```diff
@@ -44,6 +45,9 @@ def evolve(
         Number of parallel processes to be used. If greater than 1,
         parallel evaluation of the objective is supported. Currently
         not implemented. Defaults to 1.
+    reorder_genome: bool, optional
+        Whether genome reordering should be applied.
+        Defaults to False.
```
jakobj, line 50:
```suggestion
```

cgp/hl_api.py
```diff
@@ -59,7 +63,7 @@ def evolve(
     # Main loop: -1 offset since the last loop iteration will still increase generation by one
     while pop.generation < max_generations - 1:
 
-        pop = ea.step(pop, objective)
+        pop = ea.step(pop, objective, reorder_genome=reorder_genome)
```
jakobj, line 66:
```suggestion
        pop = ea.step(pop, objective)
```

cgp/ea/mu_plus_lambda.py
```diff
@@ -71,7 +71,10 @@ def initialize_fitness_parents(
         pop._parents = self._compute_fitness(pop.parents, objective)
 
     def step(
-        self, pop: Population, objective: Callable[[IndividualBase], IndividualBase]
+        self,
+        pop: Population,
+        objective: Callable[[IndividualBase], IndividualBase],
+        reorder_genome: bool = False,
```
jakobj, line 77:
```suggestion
```
please make `_reorder_genome` a member of the `MuPlusLambda` class which is an argument in `MuPlusLambda.__init__` and defaults to `True`

cgp/ea/mu_plus_lambda.py
```diff
@@ -83,12 +86,23 @@ def step(
             An objective function used for the evolution. Needs to take an
             individual (IndividualBase) as input parameter and return
             a modified individual (with updated fitness).
+        reorder_genome: bool, optional
```
jakobj, line 89: this docstring can then be moved up to the constructor.

jakobj (changes requested):
nice work! :rocket: 

i've left a few more inline comments, i'd say one more round and we're ready to merge.

cgp/genome.py
HenrikMettler: No, because the implementation of `_get_addable_nodes` is returning all nodes with no dependencies. I think it makes sense to change that implementation to return only nodes which have no dependencies _and_ are not used yet

HenrikMettler: and then replace this function

cgp/ea/mu_plus_lambda.py
HenrikMettler: What should I do with the other examples and the tests? Add an additional argument `genome_reorder = False` to `ea_params`?

cgp/ea/mu_plus_lambda.py
```diff
@@ -25,6 +25,7 @@ def __init__(
         n_processes: int = 1,
         local_search: Callable[[IndividualBase], None] = lambda combined: None,
         k_local_search: Union[int, None] = None,
+        reorder_genome: bool = True,
```
jakobj, line 28: @mschmidt87 any strong feelings about changing this default behaviour? for 1.0 i'd anyway like to use this as the default, so we might as well change it now, right?

cgp/ea/mu_plus_lambda.py
```diff
@@ -44,13 +45,21 @@ def __init__(
         k_local_search : int
             Number of individuals in the whole population (parents +
             offsprings) to apply local search to.
+        reorder_genome: bool, optional
+            Whether genome reordering should be applied.
+            If True, reorder is applied to the parents genomes
+            at every generation before creating offspring.
+            Thereby creating a neutral drift though the search space.
+            Reorder randomizes the genotype of an individual without changing the phenotype.
+            Defaults to True.
```
jakobj, line 54:
```suggestion
        reorder_genome: bool, optional
            Whether genome reordering should be applied.
            Reorder shuffles the genotype of an individual without changing its phenotype, thereby contributing to neutral drift through the genotypic search space.
            If True, reorder is applied to each parents genome at every generation before creating offsprings.
            Defaults to True.
```

cgp/genome.py
```diff
@@ -267,6 +267,146 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+
+            current_arity: int = self._determine_operator_arity(operator_idx)
+
+            for idx_gene in range(
+                1, current_arity + 1
+            ):  # shift by 1 since first gene is the operator gene
+                input_node_idx = self._dna[operator_idx + idx_gene]
+                if not self._is_input_region(
+                    input_node_idx
+                ):  # not necessary to add input regions, since they are first anyway
```
jakobj, line 397:
```suggestion
                ):  # not necessary to add input regions, since their position remains fixed
```

examples/example_reorder.py
```diff
@@ -0,0 +1,157 @@
...
+history_with_reorder["fitness_champion"] = []
+
+
+def recording_callback_with_reorder(pop):
+    history_with_reorder["fitness_champion"].append(pop.champion.fitness)
+
+
+# %%
+# and finally perform the evolution of the two populations
+print("Population without genome reordering:")
```
jakobj, line 121: not sure we need these print statements here

test/test_genome.py
jakobj: (ping)

jakobj (changes requested):
thanks for the update!

a bit more nitpicking and please implement the test that raise exceptions as my inline comment indicates. otherwise this looks awesome!

test/test_genome.py
HenrikMettler: sorry, pushed the wrong commit yesterday, it will be in the reworked version

jakobj (commented):
awesome stuff! :+1:

just one more question and I'd like to hear @mschmidt87's thoughts on changing the default behavior. if the default is changed, i guess we should adopt all tests and examples accordingly.

also, before merging it would be good to rewrite commit history such that each commit applies a logical change.

cgp/genome.py
```diff
@@ -267,6 +267,149 @@ def randomize(self, rng: np.random.RandomState) -> None:
...
+            region_idx = self._get_region_idx(gene_idx)
+            if self._is_hidden_input_gene(gene_idx, region_idx) and gene_value > region_idx:
+                permissible_values = self.determine_permissible_values_per_gene(gene_idx)
+                gene_value = rng.choice(permissible_values)
+                dna[gene_idx] = gene_value
+
+    def _get_addable_nodes(
+        self,
+        node_dependencies: Dict[int, Set[int]],
+        used_node_indices: Sequence[Union[int, None]] = [],
```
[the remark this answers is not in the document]
HenrikMettler: my bad, didn't know an empty list can have type annotation `List[int]`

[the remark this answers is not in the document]
jakobj: not sure, can it? ;)

jakobj (approved):
awesome, great work! :+1:

before merging, the history still needs some cleaning up imho.

cgp/ea/mu_plus_lambda.py
HenrikMettler: Should I set this default behavior to `False` for now?

HenrikMettler:
Is it ok, if I squash everything in one commit? Due to the previous squashing it is difficult to separate implementation, examples and test development (but I will try to keep this better seperated in future PR's)

cgp/ea/mu_plus_lambda.py
jakobj: ok, do it :) we can still change it at a later point in time.

HenrikMettler:
The build fail seems a bit odd. Travis complains about section titles in the docstring of lines that are not concerned with this PR. Has something changed in the travis config?

HenrikMettler:
@jakobj travis found a logic mistake in the example. I fixed it and while doing also adapted some docstrings to make more sense. Could you quickly look at them before merging?

jakobj:
looks great, thanks! :+1: travis is also happy! please squash the new commit, then it's ready to be merged. :tada:

PR #240, opened 2020-09-02
Catch type error in simplification

HenrikMettler:
As discussed in #231 we should catch errors thrown by sympy when calling `expr.simplify`. 
When implementing this PR I noticed that the option of the boolean `simplify = False` is never used. Maybe we could remove it altogether?

### DIFF
--- a/cgp/cartesian_graph.py
+++ b/cgp/cartesian_graph.py
@@ -412,7 +412,10 @@ class _C(torch.nn.Module):
 
         if not simplify:
             return sympy_exprs
-        else:  # simplify expression if desired
+        else:  # simplify expression if desired and possible
             for i, expr in enumerate(sympy_exprs):
-                sympy_exprs[i] = expr.simplify()
+                try:
+                    sympy_exprs[i] = expr.simplify()
+                except TypeError:
+                    RuntimeWarning(f"SymPy could not simplify expression: {expr}")
             return sympy_exprs


cgp/cartesian_graph.py
```diff
@@ -412,7 +412,10 @@ def to_sympy(self, simplify: Optional[bool] = True) -> List["sympy_expr.Expr"]:
...
         if not simplify:
             return sympy_exprs
-        else:  # simplify expression if desired
+        else:  # simplify expression if desired and possible
             for i, expr in enumerate(sympy_exprs):
-                sympy_exprs[i] = expr.simplify()
+                try:
+                    sympy_exprs[i] = expr.simplify()
+                except TypeError:
+                    Warning(f"Could not simplify expression: {expr}")
```
jakobj, line 420:
```suggestion
                    RuntimeWarning(f"SymPy could not simplify the expression: {expr}")
```
like with exceptions, please be more specific

jakobj: ~and shouldn't you nevertheless set `sympy_exprs[i] = expr`?~ nope ;)

jakobj (changes requested):
nice, thanks! just a minor comment

jakobj:
> When implementing this PR I noticed that the option of the boolean `simplify = False` is never used. Maybe we could remove it altogether?

i'm not sure what you mean. never used in the tests and examples? wouldn't this rather imply we should add a test and not remove the option? i've used it already quite a bit during development so it should certainly remain ;)

cgp/cartesian_graph.py
```diff
@@ -412,7 +412,10 @@ def to_sympy(self, simplify: Optional[bool] = True) -> List["sympy_expr.Expr"]:
...
         if not simplify:
             return sympy_exprs
-        else:  # simplify expression if desired
+        else:  # simplify expression if desired and possible
             for i, expr in enumerate(sympy_exprs):
-                sympy_exprs[i] = expr.simplify()
+                try:
+                    sympy_exprs[i] = expr.simplify()
+                except TypeError:
+                    RuntimeWarning(f"Sympy could not simplify expression: {expr}")
```
jakobj, line 420:
```suggestion
                    RuntimeWarning(f"SymPy could not simplify expression: {expr}")
```

jakobj (approved):
nice :+1: approving conditional on fixing the typo and travis being happy.

please squash commits before merging

jakobj:
great, thanks! :rocket:

PR #241, opened 2020-10-21
Make sure cache consistency uses finite return values

jakobj:
So far, always the first entry in the cache was used to compare the return values of different functions sharing the same cache. If this values happens to be non-finite, e.g., `np.inf`, objectives which in general have different return values, but both return non-finite value for this specific input, can be determined to be consistent.

This PR makes sure this does not happen by looking for cache entries with finite values when checking cache consistency.

### DIFF
--- a/cgp/utils.py
+++ b/cgp/utils.py
@@ -11,22 +11,43 @@ from .node import Node, primitives_dict
 
 
 def __check_cache_consistency(fn: str, func: Callable[..., float]) -> None:
+    """Retrieve an entry from the cache, execute the callable with the
+    cached arguments and check whether the return value matches the
+    cached result.
+
+    WARNING: consistency is only checked when a _finite_ return value
+    can be found in the cache.
+
+    """
+    cached_item: Union[Dict[str, Any], None] = __find_item_with_finite_return_value(fn)
+    if cached_item is None:
+        return
+
+    return_value: float = func(*cached_item["args"], **cached_item["kwargs"])
+
+    if not np.isclose(return_value, cached_item["return_value"]):
+        raise RuntimeError(
+            "inconsistent return values"
+            " -- are different functions using the same cache file?"
+            " please use different cache files for different functions"
+        )
+
+
+def __find_item_with_finite_return_value(fn: str) -> Union[Dict[str, Any], None]:
+    """Find a cache entry which has a finite return value."""
     if os.path.isfile(fn):
         with open(fn, "rb") as f:
-            try:
-                cursor: Dict[str, Any] = pickle.load(f)
-            except EOFError:
-                return  # no entry yet, so not possible to check
+            while True:
+                try:
+                    cursor: Dict[str, Any] = pickle.load(f)
+                except EOFError:
+                    return None  # no entry yet, so not possible to check
 
-            cached_item: Dict[str, Any] = list(cursor.values())[0]
-            return_value: float = func(*cached_item["args"], **cached_item["kwargs"])
-
-            if not np.isclose(return_value, cached_item["return_value"]):
-                raise RuntimeError(
-                    "inconsistent return values"
-                    " -- are different functions using the same cache file?"
-                    " please use different cache files for different functions"
-                )
+                cached_item: Dict[str, Any] = list(cursor.values())[0]
+                if np.isfinite(cached_item["return_value"]):
+                    return cached_item
+    else:
+        return None
 
 
 def __compute_key_from_args(*args: Any, **kwargs: Any) -> str:
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -217,6 +217,42 @@ def test_cache_decorator_consistency():
         return x
 
 
+def test_cache_decorator_does_not_compare_infinite_return_values():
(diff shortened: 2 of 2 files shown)


jakobj:
just fyi: the second commit is independent of this PR, it's just to make (the new version of) mypy happy

cgp/utils.py
```diff
@@ -11,22 +11,39 @@
 
 
 def __check_cache_consistency(fn: str, func: Callable[..., float]) -> None:
+    """Retrieve an entry from the cache, execute the callable with the
+    cached arguments and check whether the return value matches the
```
HenrikMettler, line 15:
```suggestion
    cached arguments and check whether the return value matches the finite
```
?

test/test_utils.py
```diff
@@ -217,6 +217,36 @@ def objective_h(x):
...
+    # first call produces infinite return value, identical to
+    # objective_g although in general their return values are
+    # different
+    objective_f(0.0)
+    # second call produces a finite return value which should be used
+    # to check consistency
+    objective_f(2.0)
+
+    # since the consistency check uses the finite return value it
+    # should detect that the two objectives are indeed different
```
HenrikMettler, line 239:
```suggestion
```

test/test_utils.py
```diff
@@ -217,6 +217,36 @@ def objective_h(x):
...
+
+    # first call produces infinite return value, identical to
+    # objective_g although in general their return values are
+    # different
+    objective_f(0.0)
+    # second call produces a finite return value which should be used
+    # to check consistency
+    objective_f(2.0)
+
+    # since the consistency check uses the finite return value it
```
HenrikMettler, line 238:
```suggestion
```

test/test_utils.py
HenrikMettler: I think this is clear without the comment

HenrikMettler (approved):
Just some minor comments, on your discretion whether they have to be addressed before merging

cgp/population.py
```diff
@@ -55,7 +55,12 @@ def __init__(
     def champion(self) -> IndividualBase:
         """Return parent with the highest fitness.
         """
-        return max(self._parents, key=lambda ind: ind.fitness)
+
+        def key(ind: IndividualBase) -> float:
```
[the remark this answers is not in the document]
jakobj: good point. will revert this and create a separate PR

cgp/utils.py
jakobj: good point. i've added a warning to this docstring highlighting this.

cgp/utils.py
```diff
@@ -11,22 +11,39 @@
...
+    cached result.
+
+    """
+    cached_item = __find_item_with_finite_return_value(fn)
```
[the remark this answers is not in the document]
jakobj: yes, included.

test/test_utils.py
jakobj: i agree, but in tests i tend to favor overdocumentation. not sure whether this is best practice though.

jakobj:
thanks @HenrikMettler for your comments. i've addressed all of them, please have another look.

before merging, i'll squash these commits into one logical unit.

HenrikMettler (approved):
Lgtm

jakobj:
thanks! squashed and waiting for travis before merging.

cgp/utils.py
```diff
@@ -11,22 +11,42 @@
...
+            while True:
+                try:
+                    cursor: Dict[str, Any] = pickle.load(f)
+                except EOFError:
+                    return None  # no entry yet, so not possible to check
+
+                cached_item: Dict[str, Any] = list(cursor.values())[0]
+                if np.isfinite(cached_item["return_value"]):
+                    return cached_item
+    return None
```
mschmidt87, line 49: coverage declines because of this line. By itself, that's not a big deal, but do we really need this line? Is it ever executed? 
If this is to catch the case that `os.path.isfile(fn) = False`, then we should make it explicit and put an `else:` statement here. And in this case, we should also have a small test for this.

mschmidt87 (changes requested):
Asked a question about the line that makes coverage decline.

cgp/utils.py
jakobj: this line should be executed whenever the decorator is called for the first time as the cache file does not exist yet, so i'm surprised that it's never encountered :thinking: in any case, i agree an explicit `else` would make this intention clearer.

mschmidt87: I wouldn't trust coveralls here a 100 % to be honest. But in any case, why don't you add a simple one-liner test case so got that covered explicitly?

jakobj: indeed, `mkstemp` already creates the file so this line was never hit. added a test now.

mschmidt87: `black` is complaining in test suite.

Why don't you use pre-commit hooks ...
Why don't you use pre-commit hooks ...
Why don't you use pre-commit hooks ...
...
...

jakobj: :weary: fixed

mschmidt87 (approved):
LGTM

Issue #244, opened 2020-10-30
Add objective wrapper


HenrikMettler:
In #238 it was discussed to have an objective wrapper. This has been removed from #238 since implementing the counter as an attribute of the wrapper did not work with multiple processes in parallel. However an objective wrapper might still be nice to have, since we could remove the `if individual.fitness is not None` from the implementations of objective functions.

jakobj:
hmm, i'm not sure about this any more. if it's just about catching the `fitness is not None` individuals, i'd prefer to just leave it up to the user to add it, or not (e.g., for stochastic objectives it should *not* be added), seems like less cognitive overhead when first learning to use the library.
what do you think @mschmidt87 ?

jakobj:
ping @mschmidt87 

also @HenrikMettler do you have strong opinions on this one?

HenrikMettler:
Not really. I agree with your point on stochastic objectives, I think could become a common use case. We could close this for now and for the case that the wrapper becomes more useful, I keep the local development branch?

jakobj:
> Not really. I agree with your point on stochastic objectives, I think could become a common use case. We could close this for now and for the case that the wrapper becomes more useful, I keep the local development branch?

sounds good to me :+1:

mschmidt87:
Sorry, I had overlooked this discussion completely. 
As far as I can see, this decorator is useful for deterministic objective functions, but not useful for stochastic objective functions. So, what about just renaming the decorator in #245  to something like `deterministic_objective` or more descriptive `single_probe_objective`?

jakobj:
hmm, i'm not sure. what are the benefits for the disadvantages for the user of providing such a decorator? benefit is saving two lines of boilerplate code for deterministic objectives, disadvantage is increased complexity/cognitive load. anything else?

jakobj:
ping :bell: @mschmidt87

mschmidt87:
Sorry for the delay. I am fine with closing this without merging. I see your points, but the additional cost of maintaining and explaining this code is probably not worth merging it.

PR #245, opened 2020-10-30
Objective wrapper

HenrikMettler:
Objective wrapper that checks `if individual.fitness is None` and returns `objective(individual, *args)` or the `individual` otherwise.


cgp/utils.py
```diff
@@ -201,3 +201,25 @@ def primitives_from_class_names(primitives_str: Tuple[str, ...]) -> Tuple[Type[N
         primitives.append(primitives_dict[s])
 
     return tuple(primitives)
+
+
+def objective(
+    f: Callable[[IndividualBase, Any, Any], IndividualBase]
+) -> Callable[[IndividualBase, Any, Any], IndividualBase]:
+    """ Decorator for the objective function.
+    Skips evaluations of individuals with fitness != None
```
mschmidt87, line 210: perhaps add one more half-sentence like I suggested.

mschmidt87, line 210:
```suggestion
    Skips evaluations of individuals with fitness != None, i.e., for individuals where the fitness has been evaluated already.
```

cgp/utils.py
```diff
@@ -201,3 +201,25 @@ def primitives_from_class_names(primitives_str: Tuple[str, ...]) -> Tuple[Type[N
         primitives.append(primitives_dict[s])
 
     return tuple(primitives)
+
+
```
mschmidt87, line 205: The `mypy` issue can be solved with these suggestions:

cgp/utils.py
```diff
@@ -201,3 +201,25 @@ def primitives_from_class_names(primitives_str: Tuple[str, ...]) -> Tuple[Type[N
...
+def objective(
+    f: Callable[[IndividualBase, Any, Any], IndividualBase]
+) -> Callable[[IndividualBase, Any, Any], IndividualBase]:
```
mschmidt87, line 208:
```suggestion
F = TypeVar('F', bound=Callable[..., IndividualBase])

def objective(f: F) -> F:
```

cgp/utils.py
mschmidt87: you'll need to import `TypeVar, cast` from `typing`

cgp/utils.py
```diff
@@ -201,3 +201,25 @@ def primitives_from_class_names(primitives_str: Tuple[str, ...]) -> Tuple[Type[N
...
+    """
+
+    @functools.wraps(f)
+    def wrapped(individual: IndividualBase, *args: Any, **kwargs) -> IndividualBase:
+        if individual.fitness is None:
+            return f(individual, *args, **kwargs)
+        else:
+            return individual
+
+    return wrapped
```
mschmidt87, line 225:
```suggestion
    return cast(F, wrapped)
```

mschmidt87 (changes requested):
This is very neat. 

I made one minor suggestion for the docstring and a suggestion to solve the mypy issue.

mschmidt87 (approved):
LGTM

jakobj:
@mschmidt87 what about the discussion in #244?

mschmidt87:
See the discussion in #244.

PR #248, opened 2020-11-09
Use release version of sphinx-gallery

mschmidt87:
This PR replaces the explicit commit of sphinx-gallery with its release version. This is safer and also required by pypi.org to release the package.


jakobj (approved):
excellent, thanks! :+1:

PR #253, opened 2020-11-26
Introduce callback to initialize individuals of the initial parent population

jakobj:
In some cases, one may want to provide prior information about suitable functions, e.g., to search for functions which achieve higher fitness than this minimal solution.

This PR introduces an extension of the `Population` API which introduces a callback that is called for each member of the initial parent population. This allows setting the dna of these individuals, thereby providing an initial solution.

Please do not only review with respect to correctness, but also in terms of usability @mschmidt87 @HenrikMettler

### DIFF
--- a/cgp/population.py
+++ b/cgp/population.py
@@ -1,4 +1,4 @@
-from typing import List, Union
+from typing import Callable, List, Union
 
 import numpy as np
 
@@ -17,6 +17,7 @@ class Population:
         mutation_rate: float,
         seed: int,
         genome_params: Union[dict, List[dict]],
+        individual_init: Union[Callable[[IndividualBase], IndividualBase], None] = None,
     ) -> None:
         """Init function.
 
@@ -30,6 +31,9 @@ class Population:
             Seed for internal random number generator.
         genome_params : dict
             Parameters for the genomes of the population's individuals.
+        individual_init: callable, optional
+            If not None, called for each individual of the initial
+            parent population, for example, to set the dna of parents.
         """
         self.n_parents = n_parents  # number of individuals in parent population
 
@@ -49,7 +53,10 @@ class Population:
         self.generation = 0
         self._max_idx = 0  # keeps track of maximal idx in population used to label individuals
 
-        self._generate_random_parent_population()
+        if individual_init is not None and not callable(individual_init):
+            raise TypeError("individual_init must be a callable")
+
+        self._generate_random_parent_population(individual_init)
 
     @property
     def champion(self) -> IndividualBase:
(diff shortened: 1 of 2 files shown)


test/test_population.py
```diff
@@ -76,3 +77,48 @@ def test_parent_individuals_are_assigned_correct_indices(population_params, geno
...
+        "n_columns": 3,
+        "n_rows": 1,
+        "levels_back": None,
+        "primitives": (cgp.Add, cgp.Sub, cgp.ConstantFloat),
+    }
+
+    # without passing dna_init comparison fails
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    for ind in pop.parents:
+        with pytest.raises(AssertionError):
```
HenrikMettler, line 118: Couldn't this fail by chance?

cgp/population.py
```diff
@@ -17,6 +17,7 @@ def __init__(
         mutation_rate: float,
         seed: int,
         genome_params: Union[dict, List[dict]],
+        dna_init: Union[Callable[[IndividualBase], IndividualBase], None] = None,
```
HenrikMettler, line 20: If the purpose is mainly setting the dna, why do you pass it as a Callable and not as a List?

cgp/population.py
```diff
@@ -49,7 +53,7 @@ def __init__(
         self.generation = 0
         self._max_idx = 0  # keeps track of maximal idx in population used to label individuals
 
-        self._generate_random_parent_population()
+        self._generate_random_parent_population(dna_init)
```
HenrikMettler, line 56: Maybe add
```python
if not callable(dna_init):
    raise TypeError('dna_init must be a callable')

File: cgp/population.py (hunk 2 above)
Comment (HenrikMettler) [reply cgp/population.py:20]: And I guess it would be nice to be able to set multiple initial solution for different parents. 
(or am I missing something and this is possible thanks to 'dna_init' being a callable?)

Comment (jakobj) [reply cgp/population.py:20]: a callable just allows for more flexibility. say you would like all outputs to initially read from the input while having an otherwise random dna. with a callable you could directly manipulate particular genes of the dna instead of having to construct it by yourself. and if you want to overwrite the whole dna you can just as easily to it.

File: test/test_population.py (hunk 1 above)
Comment (jakobj) [reply test/test_population.py:118]: it could indeed, it's just _very_ unlikely. i'm too lazy to do the correct computation but just assume three options for each gene, 15 genes, the probability of generating exactly this dna randomly would be something like ~10^-7, so not very likely ;)

**Comment by jakobj** [comment]
ping @mschmidt87

File: cgp/population.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (jakobj) [reply cgp/population.py]: good point, fixed!

**Comment by jakobj** [comment]
before merging i will squash the commits

File: cgp/population.py [hunk 4]
```diff
@@ -49,7 +53,10 @@ def __init__(
         self.generation = 0
         self._max_idx = 0  # keeps track of maximal idx in population used to label individuals
 
-        self._generate_random_parent_population()
+        if individual_init is not None and not callable(individual_init):
+            raise TypeError("individual_init must be a callable")
```
HenrikMettler, line 57: Sorry, I  have an additional thing to nitpick around: Can this be extended to raise an error (or warning) if `individual_init`is not of type `Callable(IndividualBase) -> IndividualBase`

HenrikMettler (changes requested):

cgp/population.py
jakobj: i don't think this is easily possible. the only way to check signatures (that i know of) is via mypy. however, the user would need to execute mypy on their code themselves.

HenrikMettler (approved):

jakobj:
thanks! squashed. now waiting for travis.

PR #254, opened 2020-12-17
Support custom atomic operators

### LINKED ISSUE #235: Allow users to define custom atomic operators
~As discussed in #234, currently the compilation of computational graphs, for example to a numpy-compatible function, relies on the presence of the necessary imports (i.e., numpy) in `cartesian_graph.py`. We should try to avoid such dependency.~
Users should be able to define custom atomic operators.

One possibility is to allow the user to set up a "compilation context" in which `eval` is called for the respective compilation targets. This could maybe also enable users to make use of custom functions in `OperatorNode`, e.g.,:
```python
def myifelse(x_0, x_1, x_2):
    if x_0 < 0:
        return x_1
    else:
        return x_2

class MyIfElseNode(cgp.OperatorNode):
    _arity = 3
    _def_output = "myifelse(x_0, x_1, x_2)"
```

jakobj:
This PR introduces a decorator which users can apply to custom functions which are then available as atomic operators in the strings defining the operation of custom nodes. Hard to explain, easy to show:
```python
    @cgp.atomic_operator
    def f_scale(x):
        return 2.0 * x

    class MyScaledAdd(cgp.node.OperatorNode):
        _arity = 2
        _def_output = "f_scale((x_0 + x_1))"
```
These few lines introduced a new primitive, i.e., a new node function, to CGP. This approach also works for custom atomic operators which depend on external libraries.

closes #235

### DIFF
--- a/cgp/__init__.py
+++ b/cgp/__init__.py
@@ -7,7 +7,7 @@ __url__ = "https://happy-algorithms-league.github.io/hal-cgp/"
 __doc__ = f"{__description__} <{__url__}>"
 
 from . import ea, local_search, utils
-from .cartesian_graph import CartesianGraph
+from .cartesian_graph import CartesianGraph, atomic_operator
 from .genome import Genome
 from .hl_api import evolve
 from .individual import IndividualMultiGenome, IndividualSingleGenome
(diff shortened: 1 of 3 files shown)


HenrikMettler (approved):
This is so nice :heart_eyes: Maybe I like it too much to be critical, but everything looks good to me

mschmidt87 (approved):
Beautiful.

PR #259, opened 2020-12-21
Provide support for hurdles

### LINKED ISSUE #255: Allow users to specify "hurdles"
[Real at al. (2020)](https://arxiv.org/abs/2003.03384) suggest that hurdles can increase the speed of an evolutionary search . This idea "[...] uses statistics of the population to early-stop the training of low performing models[...] The early-stopping criterion is the failure to reach a minimum accuracy - the _hurdle_.". Similar to their use cases a single fitness evaluation for our target applications is quite expensive, so it would make sense to adopt this strategy.

Below I suggest an implementation and, in particular, a simple API for this concept by relying on generators. This allows the user to define a fitness function as previously, but instead of `return` they will use `yield`, with each `yield` defining a "checkpoint" for early stopping. So far, this implementation only works for serial execution, a parallel version is in the making. Please let me know what you think of this in terms of usability @mschmidt87 @HenrikMettler.

```python
import numpy as np

def objective(x):
    """The object/fitness function. """
    print('first', x, '->', x)
    yield x  # first hurdle
    print('second', x, '->', x)
    yield x  # second hurdle
    print('third', x, '->', x)
    yield x + 99  # third hurdle

hurdle_rel_fitness = 0.5  # only individuals within the top 50% percentile will be allowed through each hurdle

# ==== everything below will be hidden from the user ====

 # inputs to the fitness function, i.e., individuals
combined = [0, 1, 2, 3, 4, 5, 6, 7]

# initialize generators for each individual
generators = []
for x in combined:
    generators.append(objective(x))

# keeps track of which individual still needs to be evaluated
evaluating = [True for _ in range(len(combined))]

# initial fitness evaluation until first yield, i.e., fitness after first hurdle
fitness_eval = [next(g) for g in generators]
print(fitness_eval)  # -> [0, 1, 2, 3, 4, 5, 6, 7]

while any(evaluating):

    fitness_hurdle = np.percentile([f for f in fitness_eval if f > -np.inf], hurdle_rel_fitness * 100)

    # conditional fitness evaluation; only those with fitness in upper
    # /hurdle_rel_fitness * 100/ percentile are evaluated [Real et al., 2020]
    for idx, (f, g) in enumerate(zip(fitness_eval, generators)):
        if evaluating[idx]:
            if f >= fitness_hurdle:
                try:
                    fitness_eval[idx] = next(g)
                except StopIteration:
                    evaluating[idx] = False
            else:
                fitness_eval[idx] = -np.inf
                evaluating[idx] = False

    if any(evaluating):  # only set fitness in those iterations which
                         # do not exhaust the generators; values
                         # determined in the last iteration do not
                         # represent actual fitness values
        fitness = list(fitness_eval)

print(fitness)  # -> [-inf, -inf, -inf, -inf, -inf, -inf, 105, 106]
```

jakobj:
This PR suggests an implementation that allows users to specify hurdles. Hurdles effectively implement early stopping for poorly performing individuals. The implementation allows users to pass lists of objectives to the EA/hl API and to define the "hurdle percentile" as a parameter of the EA for each of the objectives. Only individuals achieving fitness in the upper "hurdle percentile" of the _unique_ fitness values of the _whole_ population (parents + offspring) for the first objective are evaluated on the second objective, and so on.

An example is provided as extension of the minimal example (docstrings are yet to be adapted), where the performance difference is quite visible when considering a large number of function evaluations (28s w/o hurdles, 11s with hurdles on an i5-8350U).

Please review in particular with respect to usability and transparency of the implementation.

closes #255

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -26,6 +26,7 @@ class MuPlusLambda:
         local_search: Callable[[IndividualBase], None] = lambda combined: None,
         k_local_search: Union[int, None] = None,
         reorder_genome: bool = False,
+        hurdle_percentile: List = [0.0]
     ):
         """Init function
 
@@ -45,13 +46,18 @@ class MuPlusLambda:
         k_local_search : int
             Number of individuals in the whole population (parents +
             offsprings) to apply local search to.
-       reorder_genome: bool, optional
+       reorder_genome : bool, optional
             Whether genome reordering should be applied.
             Reorder shuffles the genotype of an individual without changing its phenotype,
             thereby contributing to neutral drift through the genotypic search space.
             If True, reorder is applied to each parents genome at every generation
             before creating offsprings.
             Defaults to True.
+        hurdle_percentile : List[float], optional
+            Specifies which percentile of individuals passes the
+            respective hurdle, i.e., is evaluated on the next
+            objective when providing a list of objectives to be
+            evaluated sequentially.
         """
         self.n_offsprings = n_offsprings
 
@@ -60,6 +66,7 @@ class MuPlusLambda:
         self.local_search = local_search
         self.k_local_search = k_local_search
         self.reorder_genome = reorder_genome
+        self.hurdle_percentile = hurdle_percentile
 
         self.process_pool: Union[None, "mp.pool.Pool"]
         if self.n_processes > 1:
@@ -88,7 +95,7 @@ class MuPlusLambda:
         """
         # TODO can we avoid this function? how should a population be
         # initialized?
-        pop._parents = self._compute_fitness(pop.parents, objective)
+        pop._parents = self._compute_fitness(pop.parents, objective, use_hurdles=False)
 
     def step(
         self, pop: Population, objective: Callable[[IndividualBase], IndividualBase],
@@ -155,7 +162,7 @@ class MuPlusLambda:
         offsprings: List[IndividualBase] = []
         while len(offsprings) < self.n_offsprings:
             tournament_pool = pop.rng.permutation(pop.parents)[: self.tournament_size]
-            best_in_tournament = sorted(tournament_pool, key=lambda x: -x.fitness)[0]
+            best_in_tournament = sorted(tournament_pool, reverse=True)[0]
             offsprings.append(best_in_tournament.clone())
 
         # mutate individuals to create offsprings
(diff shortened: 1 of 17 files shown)


jakobj:
rebased on master and fixed conflicts

cgp/ea/mu_plus_lambda.py
```diff
@@ -167,41 +169,57 @@ def _create_new_offspring_generation(self, pop: Population) -> List[IndividualBa
...
-                )
+                assert isinstance(self.process_pool, mp.pool.Pool)
+                ind_evaluating = self.process_pool.map(obj, ind_evaluating)
 
-        return sorted(combined, key=sort_func, reverse=True)
+            fitness_hurdle = compute_fitness_hurdle(ind_evaluating)
+
+            ind_evaluating_new = []
+            for ind in ind_evaluating:
+                assert isinstance(ind.fitness_current_objective, float)
```
HenrikMettler, line 211: If I understand correctly this will raise an error if an `ind.fitness_current_objective = np.nan` - do we really want that? This can happen easily if one uses eg. division operator genes - should the run always stop then?

cgp/ea/mu_plus_lambda.py
```diff
@@ -60,6 +61,7 @@ def __init__(
         self.local_search = local_search
         self.k_local_search = k_local_search
         self.reorder_genome = reorder_genome
+        self.hurdle_percentile = hurdle_percentile
```
HenrikMettler, line 64: missing docstring for `hurdle_percentile`

cgp/individual.py
```diff
@@ -28,21 +28,45 @@ class IndividualBase:
...
-    def __init__(self, fitness: Union[float, None]) -> None:
+    def __init__(self) -> None:
         """Init function.
-
-        fitness : float
-            Fitness of the individual.
         """
-        self.fitness: Union[float, None] = fitness
+        self._fitness: List[Union[None, float]] = [None]
+        self._objective_idx: int = 0
```
HenrikMettler, line 35: This is a nice way of implementing it imo

cgp/population.py
```diff
@@ -64,7 +64,7 @@ def champion(self) -> IndividualBase:
         """
 
         def key(ind: IndividualBase) -> float:
-            assert isinstance(ind.fitness, float)
+            # assert isinstance(ind.fitness, float)
```
HenrikMettler, line 67: remove

HenrikMettler (changes requested):

cgp/ea/mu_plus_lambda.py
```diff
@@ -26,6 +26,7 @@ def __init__(
         local_search: Callable[[IndividualBase], None] = lambda combined: None,
         k_local_search: Union[int, None] = None,
         reorder_genome: bool = False,
+        hurdle_percentile: List = [0.0]
```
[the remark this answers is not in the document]
jakobj: the issue with an additional argument would be redundancy: what if `use_hurdles` is `True`, but `hurdle_percentile` is `[0.0]` or `[0.0, 0.0]`? should this be an error? a warning? the way it is handled now is easier for the user i think

cgp/ea/mu_plus_lambda.py
jakobj: nope, this wouldn't raise an error:
```python
In [3]: isinstance(np.nan, float)
Out[3]: True
```

jakobj:
thanks for the review @HenrikMettler, i've implemented your suggestions now

examples/example_hurdles.py
```diff
@@ -0,0 +1,178 @@
+"""
+Minimal example for evolutionary regression
```
HenrikMettler, line 2: please adapt the docstrings before merging

examples/example_minimal.py
```diff
@@ -42,10 +42,11 @@ def f_target(x):
 
 
 def objective(individual):
-    if individual.fitness is not None:
+
```
HenrikMettler, line 45: Revert these changes (?)

HenrikMettler (approved):
Please just adapt the new example before merging

jakobj:
thanks @HenrikMettler ! implemented your suggestions, now waiting for travis

PR #260, opened 2020-12-21
Use a persistent process pool instead of a context manager

jakobj:
A persistent worker pool avoids the overhead associated with creating and destroying processes. This results in a significant speedup when function evaluations are not too expensive. For example, the minimal example runs approximately twice as fast (~11s w/ context manager, ~5s w/ persistent pool). For a larger number of processes this difference becomes larger.


mschmidt87 (approved):
lgtm

PR #262, opened 2020-12-29
Implement local search that does not require differential objectives -- evolution strategies

### LINKED ISSUE #146: Implement local search strategy that does not require differentiable objectives
The currently available local search strategy can only be applied to objectives that are differentiable. We should provide an alternative method that does not require this. For simplicity (and legacy ;) ) reasons I would suggest to implement [natural evolution strategies](https://en.wikipedia.org/wiki/Natural_evolution_strategy).

### LINKED ISSUE #215: Only accept outcome of local search if fitness has increased
Currently whatever changes the local search generates, the updated parameters are written into the individual and are hence going to affect the individual's fitness. We should implement a mechanism that makes sure the changes suggested by the local search are only incorporated into the genome if they do in fact increase the individual's fitness. Other strategies do not seem sensible to me.

What do you think @mschmidt87 @HenrikMettler ?

jakobj:
This PR implements separable natural evolution strategies (https://jmlr.org/papers/volume15/wierstra14a/wierstra14a.pdf) as a local search algorithm. Tests and an example are provided.

To get everything working smoothly, I needed to make sure outcomes of the local search are only accepted if they indeed increase the fitness of the respective individual (https://github.com/Happy-Algorithms-League/hal-cgp/compare/master...jakobj:enh/local-search-es?expand=1#diff-e76e9c33f502310da4d066b5ee5bd4e34e475e453c754ed908f19f34e4afef19R140 ff). Thereby closes #215.

Please review in particular with respect to ease of use and maintainability.

closes #146

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -24,7 +24,7 @@ class MuPlusLambda:
         *,
         tournament_size: Union[None, int] = None,
         n_processes: int = 1,
-        local_search: Callable[[IndividualBase], None] = lambda combined: None,
+        local_search: Union[None, Callable[[IndividualBase], None]] = None,
         k_local_search: Union[int, None] = None,
         reorder_genome: bool = False,
         hurdle_percentile: List = [0.0],
@@ -152,17 +152,53 @@ class MuPlusLambda:
         combined = self._compute_fitness(combined, objective)
         combined = self._sort(combined)
 
-        n_total = self.n_offsprings + pop.n_parents
-        k_local_search = n_total if self.k_local_search is None else self.k_local_search
-        for idx in range(k_local_search):
-            self.local_search(combined[idx])
+        if self.local_search is not None:
+            assert isinstance(pop.champion.fitness, float)
+            prev_avg_fitness: float = np.mean([ind.fitness for ind in combined])
 
-        combined = self._compute_fitness(combined, objective)
-        combined = self._sort(combined)
+            combined_copy = [ind.copy() for ind in combined]
+
+            k_local_search = (
+                len(combined_copy) if self.k_local_search is None else self.k_local_search
+            )
+            for idx in range(k_local_search):
+                self.local_search(combined_copy[idx])
+
+            combined_copy = self._compute_fitness(combined_copy, objective)
+
+            new_combined = self._create_new_combined_population_after_local_search(
+                combined, combined_copy
+            )
+
+            combined = self._sort(new_combined)
+
+            avg_fitness: float = np.mean([ind.fitness for ind in combined])
+            if prev_avg_fitness > avg_fitness:
+                raise RuntimeError(
+                    "The average fitness decreased after executing the local search. This"
+                    "indicates that something went wrong during the"
+                    "optimization. Aborting."
+                )
 
         pop.parents = self._create_new_parent_population(pop.n_parents, combined)
 
         return pop
+
+    @staticmethod
+    def _create_new_combined_population_after_local_search(
+        combined: List["IndividualBase"], combined_copy: List["IndividualBase"]
+    ) -> List["IndividualBase"]:
+        new_combined: List["IndividualBase"] = []
+        for ind in combined:
+            for ind_copy in combined_copy:
+                if ind.idx == ind_copy.idx:
+                    assert ind.fitness is not None
+                    assert ind_copy.fitness is not None
+                    if ind.fitness < ind_copy.fitness:
+                        new_combined.append(ind_copy)
+                    else:
+                        new_combined.append(ind)
+        return new_combined
 
     def _create_new_offspring_generation(self, pop: Population) -> List[IndividualBase]:
         # use tournament selection to randomly select individuals from
--- a/cgp/local_search/evolution_strategies.py
+++ b/cgp/local_search/evolution_strategies.py
@@ -1,0 +1,207 @@
+import collections
+import functools
+import math
+from typing import TYPE_CHECKING, Callable, Dict, List, Tuple, Union
+
+import numpy as np
+
+if TYPE_CHECKING:
+    from ..individual import IndividualBase
+
+
+class EvolutionStrategies:
+    def __init__(
+        self,
+        objective: Callable[["IndividualBase"], float],
+        seed: int,
+        *,
+        learning_rate_mu: float = 1.0,
+        learning_rate_sigma: Union[None, float] = None,
+        population_size: Union[None, int] = None,
+        max_steps: int = 10,
+        min_sigma: float = 1e-9,
+        fitness_shaping: bool = True,
+        mirrored_sampling: bool = True,
+    ) -> None:
+        """Evolution strategies using the natural gradient of multinormal
+        search distributions in natural coordinates.
+
+        Standard deviation of the search distribution for each
+        parameter is stored per individual. Offspring may use the
+        values of their parents if the number of used parameters has
+        not changed. Does not consider covariances between parameters.
+
+        Implementation and default values following Wierstra et
+        al. (2014). Natural evolution strategies. Journal of Machine
+        Learning Research, 15(1), 949-980.
+
+        Parameters
+        ----------
+        individual : Individual
+            Individual for which to perform local search.
+        objective : Callable[[Callable[[np.ndarray[float]], np.ndarray[float]]], float]
+            Objective function used for the local search. Needs to accept
+            an numpy-compatible function as first argument (e.g., from
+            IndividualBase.to_numpy()) and return a float representing the
+            fitness of the individual.
+        seed : int
+            Seed for internal random number generator.
+        learning_rate_mu : float, optional
+            Learning rate for mean of search distribution.
+            Defaults to 1.0.
+        learning_rate_sigma : float, optional
+            Learning rate for standard deviation of search
+            distribution. Defaults to (3 + log(dim)) / (5.0 *
+            sqrt(dim)), where dim is the dimension of the search space.
+        population_size : int, optional
+            Number of samples evaluated per step. Defaults to 4
+            + floor(3 * log(dim))), where dim is the dimension of the
+            search space..
+        max_steps : int, optional
+            Maximal number of steps. Defaults to 10.
+        min_sigma : float, optional
+            Minimal value for standard deviation of search
+            distribution. Defaults to 1e-9.
+        fitness_shaping : bool, optional
+            Whether to use fitness shaping. Defaults to True.
+        mirrored_sampling : bool, optional
+            Whether to use mirrored sampling. WARNING: Doubles the number
+            of samples per step. Defaults to True.
+
+        """
+        self.objective = objective
+        self.seed = seed
+        self.learning_rate_mu = learning_rate_mu
+        self.learning_rate_sigma = learning_rate_sigma
+        self.population_size = population_size
+        self.max_steps = max_steps
+        self.min_sigma = min_sigma
+        self.fitness_shaping = fitness_shaping
+        self.mirrored_sampling = mirrored_sampling
+
+        self.sigma: Dict[int, Dict[str, np.ndarray[float]]] = collections.defaultdict(dict)
+
+    def __call__(self, ind: "IndividualBase") -> None:
+
+        rng = np.random.RandomState(self.seed)
+
+        mu: np.ndarray[float]
+        params_names: List[str]
+        mu, params_names = ind.parameters_to_numpy_array(only_active_nodes=True)
+
+        sigma: np.ndarray[float] = self._load_sigma(ind, mu, params_names)
+
+        if len(mu) > 0:
+            if self.learning_rate_sigma is None:
+                learning_rate_sigma = (3 + math.log(len(mu))) / (5.0 * math.sqrt(len(mu)))
+            else:
+                learning_rate_sigma = self.learning_rate_sigma
+
+            if self.population_size is None:
+                population_size = 4 + int(math.floor(3 * math.log(len(mu))))
+            else:
+                population_size = self.population_size
+
+            obj: Callable[[np.ndarray[float]], float] = functools.partial(
+                self._objective_wrapper, ind=ind, params_names=params_names
+            )
+
+            for step in range(self.max_steps):
+
+                s: np.ndarray[float]
+                z: np.ndarray[float]
+                s, z = self._sample_s_and_z(mu, sigma, population_size, rng)
+
+                fitness: np.ndarray[float] = np.fromiter(map(obj, z), np.float)
+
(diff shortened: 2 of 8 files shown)


cgp/local_search/evolution_strategies.py
```diff
@@ -0,0 +1,218 @@
...
+            fitness of the individual.
+        seed : int
+            Seed for internal random number generator.
+        learning_rate_mu : float, optional
+            Learning rate for mean of search distribution.
+            Defaults to 1.0.
+        learning_rate_sigma : float, optional
+            Learning rate for standard deviation of search
+            distribution. Defaults to (3 + log(dim)) / (5.0 *
+            sqrt(dim)), where dim is the dimension of the search space.
```
HenrikMettler, line 56: The default in the docstring is different to the one above?

cgp/local_search/evolution_strategies.py
```diff
@@ -0,0 +1,218 @@
...
+        population_size : int, optional
+            Number of samples evaluated per generation. Defaults to 4
+            + floor(3 * log(dim))), where dim is the dimension of the
+            search space..
```
HenrikMettler, line 60: same

cgp/local_search/evolution_strategies.py
```diff
@@ -0,0 +1,218 @@
...
+        mu: np.ndarray[float]
+        params_names: List[str]
+        mu, params_names = ind.parameters_to_numpy_array(only_active_nodes=True)
+
+        sigma: np.ndarray[float]
+        if ind.idx in self.sigma:  # try loading sigma
+            sigma = self.sigma[ind.idx]
+            assert len(sigma) == len(mu)
+        elif ind.parent_idx in self.sigma and len(mu) == len(
+            self.sigma[ind.parent_idx]
```
HenrikMettler, line 106: Probably I am getting something wrong, but what if lengths are equal to the parent, but its different active genes?

test/test_ls_evolution_strategies.py
```diff
@@ -0,0 +1,146 @@
...
+    ind.genome._parameter_names_to_values["<p1>"] = 1.1
+    ind.genome._parameter_names_to_values["<p2>"] = 1.2
+    es = cgp.local_search.EvolutionStrategies(objective, rng_seed, max_generations=1)
+    es(ind)
+    assert ind.genome._parameter_names_to_values["<p1>"] < 1.1
+    assert ind.genome._parameter_names_to_values["<p2>"] < 1.2
+
+
+def _objective_convergence_to_maximum(f):
+    x_dummy = np.zeros((1, 1))  # input, not used
```
HenrikMettler, line 50: why this name?

HenrikMettler (commented):
Just some comments, mainly about my understanding. It looks very good to me, but I don't want to approve, since I know too little about ES

cgp/local_search/evolution_strategies.py
jakobj: yes, the default depends on the dimensionality of the search space, so we initially set it to `None`, then check for `None` below and compute the default if necessary (e.g., l115ff)

cgp/local_search/evolution_strategies.py
jakobj: that would indeed be a problem that's not handled well here. any idea how to avoid this?

test/test_ls_evolution_strategies.py
jakobj: no point, inputs are just irrelevant to the graph, so i thought to make this clear in the variable name. if this is confusing, i can change it?

cgp/ea/mu_plus_lambda.py
```diff
@@ -137,13 +137,34 @@ def step(
...
+                        assert ind.fitness is not None
+                        assert ind_copy.fitness is not None
+                        if ind.fitness < ind_copy.fitness:
+                            new_combined.append(ind_copy)
+                        else:
+                            new_combined.append(ind)
+
+            combined = self._sort(new_combined)
+
+            assert pop.champion.fitness >= prev_max_fitness
```
mschmidt87, line 167: why is this necessary?

cgp/ea/mu_plus_lambda.py
```diff
@@ -137,13 +137,34 @@ def step(
...
-        k_local_search = n_total if self.k_local_search is None else self.k_local_search
-        for idx in range(k_local_search):
-            self.local_search(combined[idx])
+        if self.local_search is not None:
+            assert isinstance(pop.champion.fitness, float)
+            prev_max_fitness: float = pop.champion.fitness
 
-        combined = self._compute_fitness(combined, objective)
-        combined = self._sort(combined)
+            combined_copy = [ind.copy() for ind in combined]
```
mschmidt87, line 144: why do we need the combined copy here and below?

cgp/local_search/evolution_strategies.py
```diff
@@ -0,0 +1,218 @@
...
+        self.population_size = population_size
+        self.max_generations = max_generations
+        self.min_sigma = min_sigma
+        self.fitness_shaping = fitness_shaping
+        self.mirrored_sampling = mirrored_sampling
+        self.n_processes = n_processes
+
+        self.rng = np.random.RandomState(seed)
+
+        self.sigma: Dict[int, np.ndarray[float]] = {}
```
mschmidt87, line 87: Is this dictionary cleared at some point? Or do we just keep adding sigmas for the parameters of individuals even if these individuals are not longer evolved?

examples/example_local_search_evolution_strategies.py
```diff
@@ -0,0 +1,174 @@
...
+# and evolutionary algorithm.
+
+pop = cgp.Population(**population_params, genome_params=genome_params)
+
+# define the function for local search; an instance of the
+# EvolutionStrategies can be called with an individual as an argument
+# for which the local search is performed
+local_search = cgp.local_search.EvolutionStrategies(
+    objective=functools.partial(inner_objective, seed=population_params["seed"]),
+    seed=population_params["seed"],
```
mschmidt87, line 117: isn't it a bit strange to use the same seed here as for the inner objective?

mschmidt87 (commented):
Looks good in general, I added some questions to understand your implementation better.

cgp/ea/mu_plus_lambda.py
jakobj: strictly speaking it's not necessary, it's just a consistency check that the local search did not decrease the maximal fitness. we could also check whether it didn't decrease the average fitness of the population, maybe that's better?

cgp/ea/mu_plus_lambda.py
jakobj: we perform the local search on a copy of the population since we might want to revert back to the original parameters if the fitness has decreased due to local search. to revert we need to keep both the original and a copy. l154ff then chooses which one to keep.

cgp/local_search/evolution_strategies.py
jakobj: it's not cleared. i'm generally not too happy with this handling of sigma. as @HenrikMettler pointed out above sigmas might be reused although the apply to different parameters if the offspring and parents have different but an identical number of active parameters. any suggestion for improvement is appreciated!

examples/example_local_search_evolution_strategies.py
jakobj: maybe, not sure. what do you mean? ;)

jakobj:
resolved conflicts.

cgp/ea/mu_plus_lambda.py
mschmidt87: My problem with this is "what if this happens?". The search is aborted but what should the user do then? Change some parameters? Abandon this method altogether?

Has this happened in your tests? If yes, do you have an idea why this happened? At the very least, we should add a meaningful error message, ideally with some sort of hint of what the user can do.

cgp/ea/mu_plus_lambda.py
mschmidt87: I see. But if we do that, can the assert statement above ever be triggered? It seems that this manoeuvre should prevent the assert statement discussed above from being triggered altogether, shouldn't it?

examples/example_local_search_evolution_strategies.py
mschmidt87: I mean, these are different random processes, so we should in principle use different seeds? Or one common RNG.
I don't think this has practical implications, but it seems a bit dirty to me to use the same seed in two places.

cgp/ea/mu_plus_lambda.py
jakobj: this should never happen unless there is an programming error somewhere in the double loop right above. error messages hence do not make sense, this is rather a protection against mistakes there, as well as documentation ("the maximal fitness should be larger or equal than before"). this has happened a lot before, but those problems have been fixed ;)

cgp/ea/mu_plus_lambda.py
jakobj: it should never be triggered, yes :+1:

examples/example_local_search_evolution_strategies.py
jakobj: yes, these are different random processes with different calls to the respective rngs. sure we can use different seeds if that is less confusing

cgp/ea/mu_plus_lambda.py
mschmidt87: So, if this has been fixed, we can remove the assert statement, right? Is this like a parachute because you're not 100% sure that the code is correct?

cgp/ea/mu_plus_lambda.py
mschmidt87: So, we can remove both the assert statement and the copy ?

cgp/ea/mu_plus_lambda.py
jakobj: i'm sure the code is correct, for now ;) it's a parachute for future generations if you will. the script should exit if this condition is not fulfilled for whatever as something terrible went wrong.

cgp/ea/mu_plus_lambda.py
jakobj: nope, copy is necessary. the copy is the thing that makes sure only fitness improvements are accepted. regarding the assert, see above.

jakobj:
this PR should not be merged until #279 has been merged and this has been rebased. this way the evostrat class can use the same parallel context as mu+lambda

cgp/ea/mu_plus_lambda.py
mschmidt87: Okay. I would either remove it or add an informative error message like
```The fitness decreased after executing the local search. This indicates that something went wrong during the optimization. Therefore aborting.```

I would prefer to just remove it.

File: cgp/ea/mu_plus_lambda.py (hunk 6 above)
Comment (mschmidt87) [reply cgp/ea/mu_plus_lambda.py:159]: Okay, understood.

File: examples/example_local_search_evolution_strategies.py (hunk 8 above)
Comment (mschmidt87) [reply examples/example_local_search_evolution_strategies.py:117]: yes, please use a different seed. Just adding +1 would be sufficient, I guess.

File: cgp/ea/mu_plus_lambda.py (hunk 5 above)
Comment (HenrikMettler) [reply cgp/ea/mu_plus_lambda.py:182]: Imo, the message would be fine too

Comment (jakobj) [reply cgp/ea/mu_plus_lambda.py:182]: ok, replaced with proper error message

File: examples/example_local_search_evolution_strategies.py (hunk 8 above)
Comment (jakobj) [reply examples/example_local_search_evolution_strategies.py:117]: ok

**Comment by jakobj** [comment]
ok, to simplify this, i've dropped support for multiple processes, otherwise we need to figure out how to properly nest multiprocessing pools. this should be added in a later PR. the last thing to figure out is how to handle `sigma` properly across multiple generations. any suggestions?

**Comment by HenrikMettler** [comment]
I am currently looking into whether we can reuse something from the `parameter_names_to_values` that was recently introduced in `genome.reorder`. 
If I have something working, should I simply create a PR on your fork?

File: cgp/local_search/evolution_strategies.py (hunk 3 above)
Comment (jakobj) [reply cgp/local_search/evolution_strategies.py]: #907e2b6 fixes this issue

**Comment by jakobj** [comment]
i've fixed the handling of sigma in 907e2b6, please have another look

File: cgp/local_search/evolution_strategies.py [hunk 9]
```diff
@@ -0,0 +1,209 @@
...
+                fitness: np.ndarray[float] = np.fromiter(map(obj, z), np.float)
+
+                s, z, utility = self._determine_utility(s, z, fitness)
+
+                mu, sigma = self._update_parameters_of_search_distribution(
+                    s, utility, learning_rate_sigma, mu, sigma
+                )
+
+                if np.all(sigma < self.min_sigma):
+                    break
```
HenrikMettler, line 124: Is this to stop the local search, if it has "converged" to a narrow interval?

examples/example_local_search_evolution_strategies.py
```diff
@@ -0,0 +1,180 @@
...
+    "k_local_search": 2,
+}
+
+evolve_params = {"max_generations": int(args["--max-generations"]), "min_fitness": 0.0}
+
+# restrict the number of steps in the local search; since parameter
+# values are propagated from parents to offsprings, parameter values
+# may be iteratively improved across generations despite the small
+# number of steps per generation
+local_search_params = {"max_generations": 5}
```
HenrikMettler, line 110: Just thinking, whether we maybe shouldn't name an iteration in the local search `generation` This might lead to confusion, especially since (correct me if i am wrong) 5 (or n) generations of local search are taken care of within one generation of cgp evolution.

HenrikMettler (approved):
Very nice! Just added a small question and a discussion point about a variable name. Other than that ready for merge I think :) :rocket:

examples/example_local_search_evolution_strategies.py
jakobj: good point, maybe `step` is better than `generation`?

cgp/local_search/evolution_strategies.py
jakobj: yes, typically convergence of sigma is a sufficient criterion for convergence of the search. but i think this should be removed to avoid suprises. in that case always `max_generations` steps will be performed, easy to understand.

examples/example_local_search_evolution_strategies.py
jakobj: done.

jakobj:
thanks for your reviews @HenrikMettler @mschmidt87 

i've squashed commits now and waiting for travis before merging

examples/example_local_search_evolution_strategies.py
HenrikMettler: sorry I forgot to reply

PR #263, opened 2020-12-30
Improved caching decorator

jakobj:
This PR improves the disk cache decorator:
- the cache file consists of a single large dict containing (key: fitness) pairs; no need for a linear search through keys to find a match
- function arguments are now only stored for a single function call to support consistency checks; this avoid unnecessarily large cache files by storing arguments for each(!) result
- for using multiple processes, a lock is introduced to make sure they are reading from/writing to the cache file without interference

### DIFF
--- a/cgp/utils.py
+++ b/cgp/utils.py
@@ -2,12 +2,15 @@ import functools
 import hashlib
 import os
 import pickle
-from typing import Any, Callable, Dict, List, Tuple, Type, Union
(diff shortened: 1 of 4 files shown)


HenrikMettler (approved):
Looks good to me :)

jakobj:
thanks @HenrikMettler! squashed commits, now waiting for travis

PR #264, opened 2021-01-04
Reset parameters when the corresponding node is created by mutation of a genome

jakobj:
For suboptimal expressions, for example at the beginning of the evolution, the local search can lead to abnormally large parameter values. If these value persist across all generations, the corresponding parameterized nodes may never be used in the final expression: as soon as they are included, the (initially set) large parameter values may lead to poor fitness.

This PR adapts the mutation step such that when a node with parameters is created, the corresponding values are reset to the initial values specified by the user in the definition of the node. This should avoid the issue described above.


HenrikMettler (approved):
Nice, looks all good to me :)

PR #267, opened 2021-01-05
Rename input -> address for address genes in hidden/outp…

### LINKED ISSUE #242: Use "address gene" for genes encoding indices of input nodes
Currently we use "input gene" for these which gets very confusing because "inputs" are also the nodes that hold variable leaf values. As discussed with @HenrikMettler we should rather use "address gene" for these, as this also agrees with the nomenclature in the literature.

HenrikMettler:
- renamed input_gene to address gene to describe the nodes from which hidden/output nodes read. Consistent with general literature. Closes #242.

### DIFF
--- a/cgp/cartesian_graph.py
+++ b/cgp/cartesian_graph.py
@@ -181,7 +181,7 @@ class CartesianGraph:
                 continue
             active_nodes_by_hidden_column_idx[self._hidden_column_idx(node.idx)].add(node)
 
-            for i in node.input_nodes:
+            for i in node.addresses:
                 nodes_to_process.append(self._nodes[i])
 
         return active_nodes_by_hidden_column_idx
--- a/cgp/node.py
+++ b/cgp/node.py
@@ -33,25 +33,25 @@ class Node:
 
     _arity: int
     _active: bool = False
-    _input_nodes: List[int]
+    _addresses: List[int]
     _output: float
     _output_str: str
     _idx: int
 
-    def __init__(self, idx: int, inputs: List[int]) -> None:
+    def __init__(self, idx: int, addresses: List[int]) -> None:
         """Init function.
 
         Parameters
         ----------
         idx : int
             id of the node.
-        inputs : List[int]
-            List of integers specifying the input nodes to this node.
+        addresses : List[int]
+            List of integers specifying the address of input nodes to this node.
         """
         self._idx = idx
-        self._input_nodes = inputs[: self._arity]
+        self._addresses = addresses[: self._arity]
 
-        assert idx not in inputs
+        assert idx not in addresses
 
     def __init_subclass__(cls: Type["Node"]) -> None:
         super().__init_subclass__()
@@ -63,11 +63,11 @@ class Node:
 
     @property
     def max_arity(self) -> int:
-        return len(self._input_nodes)
+        return len(self._addresses)
 
     @property
-    def input_nodes(self) -> List[int]:
-        return self._input_nodes
+    def addresses(self) -> List[int]:
+        return self._addresses
 
     @property
     def idx(self) -> int:
@@ -76,15 +76,15 @@ class Node:
     def __repr__(self) -> str:
         return (
             f"{self.__class__.__name__}(idx: {self.idx}, active: {self._active}, "
-            f"arity: {self._arity}, input_nodes {self._input_nodes}"
+            f"arity: {self._arity}, addresses of inputs {self._addresses}"
         )
 
     def pretty_str(self, n: int) -> str:
         used_characters = 0
         used_characters += 3  # for two digit idx and following whitespace
         used_characters += 3  # for "active" marker
-        # for brackets around inputs, two digits inputs separated by
-        # comma and last input without comma
+        # for brackets around addresses, two digits addresses separated by
+        # comma and last address without comma
         used_characters += 2 + self.max_arity * 3 - 1
 
         assert n > used_characters
@@ -99,7 +99,7 @@ class Node:
             if self._arity > 0:
                 s += "("
                 for i in range(self._arity):
-                    s += f"{self._input_nodes[i]:02d},"
+                    s += f"{self._addresses[i]:02d},"
                 s = s[:-1]
                 s += ")"
                 for i in range(self.max_arity - self._arity):
@@ -159,7 +159,7 @@ class OperatorNode(Node):
     Subclasses provide the atomic operations of the computational graph.
     """
 
-    _input_names: Tuple[str, ...]
+    _address_names: Tuple[str, ...]
     _parameter_names: Tuple[str, ...]
     _initial_values: Dict[str, Callable[[], float]]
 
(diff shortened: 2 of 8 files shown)


cgp/node_impl.py
```diff
@@ -11,35 +11,35 @@ class ConstantFloat(OperatorNode):
 
 
 class Add(OperatorNode):
-    """A node that adds its two inputs."""
+    """A node that adds its two addresses."""
```
jakobj, line 14: mhm. nope, i don't think we should rename this here ;) i mean we don't need to distort language, the important thing is to use good variables names in the code.

examples/example_parametrized_nodes.py
```diff
@@ -5,7 +5,7 @@
 Example demonstrating the use of Cartesian genetic programming for a
 regression task that requires fine tuning of constants in parametrized
 nodes. This is achieved by introducing a new node, "ParametrizedAdd"
-which produces a scaled and shifted version of the sum of its inputs.
+which produces a scaled and shifted version of the sum of its addresses.
```
jakobj, line 8: (see above, i would keep input here, as this is easily understandable, but what is "the sum of its addresses"?)

test/test_cartesian_graph.py
```diff
@@ -20,7 +20,7 @@ def test_direct_input_output():
     )
     genome.randomize(np.random)
 
-    genome[-2:] = [0, ID_NON_CODING_GENE]  # set inputs for output node to input node
+    genome[-2:] = [0, ID_NON_CODING_GENE]  # set address node for output node to input node
```
jakobj, line 23:
```suggestion
    genome[-2:] = [0, ID_NON_CODING_GENE]  # set address for output node to input node
```

cgp/cartesian_graph.py
```diff
@@ -181,7 +181,7 @@ def _determine_active_nodes(self) -> Dict[int, Set[Node]]:
                 continue
             active_nodes_by_hidden_column_idx[self._hidden_column_idx(node.idx)].add(node)
 
-            for i in node.input_nodes:
+            for i in node.address_nodes:
```
jakobj, line 184: (see below, maybe better `addresses`?)

examples/example_parametrized_nodes.py
jakobj: (also in code below)

test/test_genome.py
```diff
@@ -359,15 +359,15 @@ def test_mutation_rate(rng_seed, mutation_rate):
     genome.randomize(rng)
 
     def count_n_immutable_genes(n_inputs, n_outputs, n_rows):
-        length_per_region = genome.primitives.max_arity + 1  # function gene + input gene addresses
+        length_per_region = genome.primitives.max_arity + 1  # function gene + gene addresses
```
jakobj, line 362:
```suggestion
        length_per_region = genome.primitives.max_arity + 1  # function gene + address gene
```

test/test_genome.py
```diff
@@ -447,7 +447,7 @@ def test_only_silent_mutations(genome_params, mutation_rate, rng_seed):
     genome.dna = dna_fixed
     graph = CartesianGraph(genome)
     active_regions = graph.determine_active_regions()
-    length_per_region = genome.primitives.max_arity + 1  # function gene + input gene addresses
+    length_per_region = genome.primitives.max_arity + 1  # function gene + gene addresses
```
jakobj, line 450:
```suggestion
    length_per_region = genome.primitives.max_arity + 1  # function gene + address gene
```

test/test_genome.py
```diff
@@ -488,10 +488,10 @@ def test_permissible_values(genome_params):
...
 
     permissible_function_gene_values = np.arange(len(genome._primitives._primitives))
-    permissible_inputs_first_internal_column = np.array([0, 1])
-    permissible_inputs_second_internal_column = np.array([0, 1, 2, 3, 4])
-    permissible_inputs_third_internal_column = np.array([0, 1, 2, 3, 4, 5, 6, 7])
-    permissible_inputs_output_gene = np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+    permissible_addresses_first_internal_column = np.array([0, 1])
+    permissible_addresses_second_internal_column = np.array([0, 1, 2, 3, 4])
+    permissible_addresses_third_internal_column = np.array([0, 1, 2, 3, 4, 5, 6, 7])
+    permissible_addresses_output_gene = np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
```
jakobj, line 494:
```suggestion
    permissible_addresses_output = np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
```

test/test_node.py
```diff
@@ -5,23 +5,23 @@
...
-    provided, Node.inputs only returns the initial <arity> inputs,
+def test_addresses_are_cut_to_match_arity():
+    """Test that even if a list of addresses longer than the node arity is
+    provided, Node.addresses only returns the initial <arity> addresses,
     ignoring the inactive genes.
 
     """
     idx = 0
-    input_nodes = [1, 2, 3, 4]
+    address_nodes = [1, 2, 3, 4]
```
jakobj, line 15:
```suggestion
    addresses = [1, 2, 3, 4]
```

jakobj (changes requested):
great, thanks for working on this @HenrikMettler, a tedious task.

i've added some inline comments that should be addresses to improve the usage of "address" a bit.

cgp/node.py
```diff
@@ -33,25 +33,25 @@ class Node:
...
         """Init function.
 
         Parameters
         ----------
         idx : int
             id of the node.
-        inputs : List[int]
-            List of integers specifying the input nodes to this node.
+        addresses : List[int]
+            List of integers specifying the address nodes to this node.
```
jakobj, line 49:
```suggestion
            List of integers specifying the address of input nodes to this node.
```

cgp/node.py
```diff
@@ -175,7 +175,7 @@ class OperatorNode(Node):
 
     def __init_subclass__(cls: Type["OperatorNode"]) -> None:
         super().__init_subclass__()
-        OperatorNode._extract_input_names_from_def_output(cls)
+        OperatorNode._extract_address_names_from_def_output(cls)
```
jakobj, line 178: i think this should still be called "input names", "address names" is not very intuitive is it?

test/test_node.py
```diff
@@ -536,22 +536,22 @@ class CustomAdd(cgp.OperatorNode):
 
 def test_repr():
     idx = 0
-    input_nodes = [1, 2, 3, 4]
+    address_nodes = [1, 2, 3, 4]
```
jakobj, line 539:
```suggestion
    addresses = [1, 2, 3, 4]
```

cgp/node.py
HenrikMettler: agree

cgp/node.py
```diff
@@ -76,15 +76,15 @@ def idx(self) -> int:
     def __repr__(self) -> str:
         return (
             f"{self.__class__.__name__}(idx: {self.idx}, active: {self._active}, "
-            f"arity: {self._arity}, input_nodes {self._input_nodes}"
+            f"arity: {self._arity}, address_nodes {self._addresses}"
```
jakobj, line 79:
```suggestion
            f"arity: {self._arity}, addresses {self._addresses}"
```

jakobj: or "addresses of inputs", this should easily understandable as its printed to the screen

jakobj (approved):
great stuff! :+1: 

one more small comment, then please squash the commits

jakobj:
excellent :+1: and merging

PR #268, opened 2021-01-05
Check tournament_size <= n_parents in hl_api

### LINKED ISSUE #246: Make sure `tournament_size` is always smaller or equal to number of parents
Currently, there does not seem to be a restriction on the tournament size. However, for one parent, a tournament of two does not make any sense as one would choose an individual from a set of two identical individuals. Such a setting should hence raise an exception.

HenrikMettler:
Makes the sanity check that the tournament size is smaller or equal to the number of parents in a population. Closes #246 
Note that this PR is not ready for merging - due to the sanity checks various examples now raise errors. (Will adapt if the concept is approved)

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -159,6 +159,10 @@ class MuPlusLambda:
     def _create_new_offspring_generation(self, pop: Population) -> List[IndividualBase]:
         # use tournament selection to randomly select individuals from
         # parent population
+
+        if self.tournament_size > pop.n_parents:
+            raise ValueError("tournament_size must be less or equal n_parents")
+
         offsprings: List[IndividualBase] = []
         while len(offsprings) < self.n_offsprings:
             tournament_pool = pop.rng.permutation(pop.parents)[: self.tournament_size]
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -20,8 +20,8 @@ class MuPlusLambda:
     def __init__(
         self,
         n_offsprings: int,
-        tournament_size: int,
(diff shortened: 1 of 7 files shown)


jakobj (changes requested):
thanks for working on this!

however, i would strongly suggest to have such checks not only in the high level API, but deeper in the library. i think it makes most sense in `mu_plus_lambda._create_offspring_population` before doing anything else. there you have all information necessary to perform the check.

HenrikMettler:
Note to myself: Address #271 within this PR

HenrikMettler:
> however, i would strongly suggest to have such checks not only in the high level API, but deeper in the library. i think it makes most sense in `mu_plus_lambda._create_offspring_population` before doing anything else. there you have all information necessary to perform the check.

I can move it there, but the downside is, that `mu_plus_lambda._create_offspring_population` is called within `mu_plus_lambda.step` -> the sanity check will be done at every time step which seems unreasonable to me.

jakobj:
> > however, i would strongly suggest to have such checks not only in the high level API, but deeper in the library. i think it makes most sense in `mu_plus_lambda._create_offspring_population` before doing anything else. there you have all information necessary to perform the check.
> 
> I can move it there, but the downside is, that `mu_plus_lambda._create_offspring_population` is called within `mu_plus_lambda.step` -> the sanity check will be done at every time step which seems unreasonable to me.

indeed, the lower you go in the code the more often you will evaluate such checks. however, in terms of runtime we will not see _any_ effect (+1 `if` per generation) while making the code more solid. so while in general i agree with your concern, in this case i think it's the best option.

jakobj (approved):
awesome, thanks! :+1:

let's wait for travis to be happy before merging.

PR #270, opened 2021-01-05
Remove support for calling graph directly

jakobj:
There does not seem to be any need for calling the graph object directly. As soon as more than a single call is required, the `to_...` versions are much faster as they only require traversing the cartesian graph once.

This PR hence removes support for this functionality to i) prevent users from accidentally making their code slow and ii) relieve us of the burden of maintaining it.


mschmidt87 (approved):
Makes sense to me. Removing code is always appreciated.

PR #272, opened 2021-01-06
Also use all args and kwargs when computing fec cache key

jakobj:
This PR fixes a bug in the fec cache where only the evaluation of the function encoded in an individual would determine the cache key, while all other (keyword)arguments would be ignored. This means that calls like this for a decorated inner objective:
```python
    y0 = inner_objective(ind, 5)
    y1 = inner_objective(ind, 10)
```
would yield the same values as the second one is assigned the same key and hence the cached value is returned.

### DIFF
--- a/cgp/utils.py
+++ b/cgp/utils.py
@@ -96,7 +96,7 @@ def __compute_key_from_evaluation_and_args(
     else:
         assert False  # should never be reached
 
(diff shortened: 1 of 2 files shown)


HenrikMettler (approved):
LGTM!

PR #273, opened 2021-01-06
Fix error message wrong size passed to to_torch/to_numpy graphs

jakobj:
The current error message can be a bit confusing as we are trying to guess the batch dimension. If someone provides a transposed array with shape, e.g., (2, 200) to a graph with two inputs, the message claims the input should have shape (2, 2), while the correct size would be (200, 2).

This PR changes the error message to avoid such confusing output.


HenrikMettler (approved):
LGTM

PR #275, opened 2021-01-07
Remove outdated n_processes argument in hl_api

jakobj:


HenrikMettler (changes requested):
Could you remove it from the examples too?

jakobj:
> Could you remove it from the examples too?

hmm, where exactly? it should not be used in the examples, otherwise our tests should catch it

HenrikMettler (approved):
my bad, those were actually in the `ea_params` not `evolve_params`

PR #279, opened 2021-01-07
Create process pools via "spawn" from a custom context

### LINKED ISSUE #274: Use `spawn` instead of `fork` for process pool in `MuPlusLambda` to avoid issues with autograd
Autograd is not compatible with `fork`-based parallelism (https://github.com/pytorch/pytorch/wiki/Autograd-and-Fork). We should hence use the `spawn` variant. This will require some thinking as the `Lock` for the cache decorator will need to be from the same context (https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods).

jakobj:
Otherwise using multiple processes can lead to issues with torch's autograd (https://github.com/pytorch/pytorch/wiki/Autograd-and-Fork).

closes #274

### DIFF
--- a/cgp/__init__.py
+++ b/cgp/__init__.py
@@ -11,6 +11,7 @@ from .cartesian_graph import CartesianGraph, atomic_operator
 from .genome import Genome
 from .hl_api import evolve
 from .individual import IndividualMultiGenome, IndividualSingleGenome
+from .mp_context import mp_context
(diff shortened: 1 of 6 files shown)


jakobj:
this might be superfluous as one determine the start method at runtime by calling `mp.set_start_method("spawn")` at the beginning of the main method. will investigate further.

Issue #276, opened 2021-01-07
Provide default for `levels_back`


jakobj:
Typically we choose `levels_back` = `n_columns`, we should hence just set this as a default.

HenrikMettler:
Isn't this already the case? [level_back default](https://github.com/Happy-Algorithms-League/hal-cgp/blob/05fed9df8a2b1230673bbeff23c98db33e0f4b44/cgp/genome.py#L71)

jakobj:
> Isn't this already the case? [level_back default](https://github.com/Happy-Algorithms-League/hal-cgp/blob/05fed9df8a2b1230673bbeff23c98db33e0f4b44/cgp/genome.py#L71)

yes, you are right, but this still requires the user to pass `None`. to make the user's life easier we should provide `None` here: https://github.com/Happy-Algorithms-League/hal-cgp/blob/05fed9df8a2b1230673bbeff23c98db33e0f4b44/cgp/genome.py#L33

HenrikMettler:
Ah right, I will quickly push a PR

PR #281, opened 2021-01-08
Make levels_back parameter optional

HenrikMettler:
Addresses #276. There is one minor thing: The current checks allow `n_columns=0` and set `levels_back=n_columns` (if None provided), but they don't allow `levels_back=0` 
This could lead to confusing error messages

### DIFF
--- a/cgp/node_validation.py
+++ b/cgp/node_validation.py
@@ -28,7 +28,7 @@ def _create_genome(cls: Type["OperatorNode"]) -> "Genome":
     from .genome import ID_INPUT_NODE, ID_NON_CODING_GENE, ID_OUTPUT_NODE, Genome
 
     primitives = (cls,)
-    genome = Genome(1, 1, 1, 1, 1, primitives)
+    genome = Genome(1, 1, 1, 1, primitives)
     dna = [ID_INPUT_NODE]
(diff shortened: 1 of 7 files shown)


cgp/node_validation.py
```diff
@@ -28,7 +28,7 @@ def _create_genome(cls: Type["OperatorNode"]) -> "Genome":
     from .genome import ID_INPUT_NODE, ID_NON_CODING_GENE, ID_OUTPUT_NODE, Genome
 
     primitives = (cls,)
-    genome = Genome(1, 1, 1, 1, 1, primitives)
+    genome = Genome(1, 1, 1, 1, primitives, 1)
```
jakobj, line 31: you can drop it here and in all following cases where `levels_back == n_columns`

jakobj (changes requested):
thanks for taking care of this!

i think there are more places in which you can just drop the `levels_back` values as it is the same as `n_columns`. these occurrences should be fixed.

regarding your question: i think the best option is to treat the `n_columns == 0` case separately and allow `n_levels_back = 0` for this.

jakobj (approved):
looks good :+1:

let's wait for travis to be happy before merging

PR #285, opened 2021-01-20
Use mp fork context by default, but allow switching to spawn context

### LINKED ISSUE #274: Use `spawn` instead of `fork` for process pool in `MuPlusLambda` to avoid issues with autograd

jakobj:
This PR introduces custom fork and spawn context for our library. Be default the default is used ("fork"), but the user may switch to the "spawn" context by calling `ea.use_spawn_context()` after creating an instance of `MuPlusLambda`. The implementation tries to have minimal impact on the user experience while fixing fixing issues with autograd in connection with forked processes.

fixes #274

### DIFF
--- a/cgp/__init__.py
+++ b/cgp/__init__.py
@@ -11,6 +11,7 @@ from .cartesian_graph import CartesianGraph, atomic_operator
 from .genome import Genome
 from .hl_api import evolve
 from .individual import IndividualMultiGenome, IndividualSingleGenome
+from .mp_context import fork_context, spawn_context
 from .node import OperatorNode
 from .node_impl import Add, ConstantFloat, Div, IfElse, Mul, Parameter, Pow, Sub
 from .population import Population
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -4,6 +4,7 @@ from typing import Callable, List, Union
 import numpy as np
 
 from ..individual import IndividualBase
+from ..mp_context import fork_context, spawn_context
 from ..population import Population
 
 
@@ -78,7 +79,7 @@ class MuPlusLambda:
 
         self.process_pool: Union[None, "mp.pool.Pool"]
         if self.n_processes > 1:
-            self.process_pool = mp.Pool(processes=self.n_processes)
+            self.process_pool = fork_context.Pool(processes=self.n_processes)
         else:
             self.process_pool = None
         self.n_objective_calls: int = 0
@@ -283,3 +284,8 @@ class MuPlusLambda:
         for off in offsprings:
             off.mutate(self._mutation_rate, rng)
         return offsprings
+
+    def use_spawn_context(self):
+        assert self.n_processes > 1
+        self.process_pool.close()
+        self.process_pool = spawn_context.Pool(processes=self.n_processes)
--- a/cgp/mp_context.py
+++ b/cgp/mp_context.py
@@ -1,0 +1,4 @@
+import multiprocessing as mp
+
+fork_context: mp.context.ForkContext = mp.get_context("fork")
+spawn_context: mp.context.SpawnContext = mp.get_context("spawn")
--- a/cgp/utils.py
+++ b/cgp/utils.py
@@ -220,7 +220,7 @@ def disk_cache(
         __check_cache_consistency(fn, func)
 
         @functools.wraps(func)
-        def wrapper(*args: Any, **kwargs: Any) -> Union[float, None]:
+        def wrapper(*args: Any, **kwargs: Any) -> float:
 
             key: str
             if use_fec:


mschmidt87:
I might have missed that from looking at the code and the issue, but: if I understand correctly, we need to use the `spawn` context if autograd gets involved, i.e., when we use local differentiable optimization. Correct? Shouldn't we build in an automatic switch for that then, i.e. shouldn't we switch to the spawn_context if `MuPlusLambda(..., local_search = cgp.local_search.gradient_based, ...)` ?

jakobj:
we can not switch depending on the type of local search method, as even without local search one might want to use autograd in objective functions. we could implement an automatic check by catching exactly this autograd error, but i would much rather rely on the user to switch this if they observe an issue.

mschmidt87:
> > I might have missed that from looking at the code and the issue, but: if I understand correctly, we need to use the `spawn` context if autograd gets involved, i.e., when we use local differentiable optimization. Correct? Shouldn't we build in an automatic switch for that then, i.e. shouldn't we switch to the spawn_context if `MuPlusLambda(..., local_search = cgp.local_search.gradient_based, ...)` ?
> 
> we can not switch depending on the type of local search method, as even without local search one might want to use autograd in objective functions. we could implement an automatic check by catching exactly this autograd error, but i would much rather rely on the user to switch this if they observe an issue.

Okay, makes sense. However, what about catching the autograd error and then submit our own error message that tells the user what (probably) the right thing to do is, namely to switch to the spawn context?

jakobj:
> Okay, makes sense. However, what about catching the autograd error and then submit our own error message that tells the user what (probably) the right thing to do is, namely to switch to the spawn context?

I'm not really comfortable with catching error message which we are not controlling and assuming that our code always handles this correctly. As you also imply ("probably") this does not seem bullet proof. I'd much rather just let torch trigger the exception, and add an FAQ to the readme where it is explained how to fix it.

mschmidt87:
> > Okay, makes sense. However, what about catching the autograd error and then submit our own error message that tells the user what (probably) the right thing to do is, namely to switch to the spawn context?
> 
> I'm not really comfortable with catching error message which we are not controlling and assuming that our code always handles this correctly. As you also imply ("probably") this does not seem bullet proof. I'd much rather just let torch trigger the exception, and add an FAQ to the readme where it is explained how to fix it.

That would also be okay. However, what can go wrong in submitting our own error message? To be more clear, I am thinking of something like this:

```python
except TorchError as e:
        raise TorchError(f"{str(e)} \n You need to use spawn.")
```
where `TorchError` needs to be replaced by the correct error type (not sure which one is actually raised by torch).

jakobj:
> > I'm not really comfortable with catching error message which we are not controlling and assuming that our code always handles this correctly. As you also imply ("probably") this does not seem bullet proof. I'd much rather just let torch trigger the exception, and add an FAQ to the readme where it is explained how to fix it.
> 
> That would also be okay. However, what can go wrong in submitting our own error message? To be more clear, I am thinking of something like this:
> 
> ```python
> except TorchError as e:
>         raise TorchError(f"{str(e)} \n You need to use spawn.")
> ```
> 
> where `TorchError` needs to be replaced by the correct error type (not sure which one is actually raised by torch).

well, a lot can go wrong by catching an error message, assuming we know what it says and raising an error ourselves -- this might result in mighty confusion for the user ;) but yeah, maybe we could check the whole error text to match, and only catch it then? i'll give that a try.

mschmidt87:
> > > I'm not really comfortable with catching error message which we are not controlling and assuming that our code always handles this correctly. As you also imply ("probably") this does not seem bullet proof. I'd much rather just let torch trigger the exception, and add an FAQ to the readme where it is explained how to fix it.
> > 
> > 
> > That would also be okay. However, what can go wrong in submitting our own error message? To be more clear, I am thinking of something like this:
> > ```python
> > except TorchError as e:
> >         raise TorchError(f"{str(e)} \n You need to use spawn.")
> > ```
> > 
> > 
> > where `TorchError` needs to be replaced by the correct error type (not sure which one is actually raised by torch).
> 
> well, a lot can go wrong by catching an error message, assuming we know what it says and raising an error ourselves -- this might result in mighty confusion for the user ;) but yeah, maybe we could check the whole error text to match, and only catch it then? i'll give that a try.

Yes, I was thinking of checking the error message. Sure, it's a bit hacky, I admit, but perhaps better than nothing. What about writing in our suffix to the message: "You need to use spawn, please check the FAQ for more details." and then we add a more detailed explanation to the FAQ.

If we catch the error --> more information for the user, if we don't catch it, the user can still go to the FAQ. So, it's definitely better than not saying anything and always expecting the user to figure out they need to go check the FAQ.

mschmidt87:
What's the status here? I made a suggestion above regarding an error message.
Why is this PR now marked as draft?

jakobj:
> What's the status here? I made a suggestion above regarding an error message.
> Why is this PR now marked as draft?

yes, your suggestion was good. but i was trying to create a test for this and failed. i didn't want to merge this before i'm not 100% certain under which conditions the error is generated, but didn't find the time to do that yet. i'd suggest this to remain in draft state until i figure it out.

mschmidt87:
I think this branch needs to be rebased on master to pull in the changes in the continuous integration and make the checks run.

jakobj:
closing for now. will reopen when i'm actively working on this.

PR #294, opened 2021-03-20
Redesign signatures of compilation targets

### LINKED ISSUE #287: Redesign call signatures of compilation targets
Currently, function obtained from `to_func` and `to_sympy` both expect a list of length `n_inputs`  and return a list of length `n_outputs`. Furthermore, the `to_numpy` and `to_torch` results expect inputs of shape `(<d>, n_inputs)` where `d` can have arbitrary size, and the function is applied to each element from `d`, resulting in an output of shape `(<d>, n_outputs)`. This is rather unintuitive and can be cumbersome to use, e.g.,
```python
y = f([x])[0]
```
or
```python
v = self.f(np.vstack([m, s]).T)
```

I suggest to change the call signatures of all these functions to accept `n_inputs` arguments and return a tuple of size `n_outputs`. For the numpy and torch targets it shouldn't matter any more what the size of the input is.

What do you think @mschmidt87 @HenrikMettler ?

closely related to #278

jakobj:
This PR aims at making the usage of the results from various `to_...` functions more intuitive. It also improves their documentation. In the following the new usage is described in more detail.

`to_func` -> f:
- f expects the same number of arguments as the graph has input
- each argument must be a single value
- f returns a list of values, one for each output
- if only one output is present it'll return only it's output value; _not_ a list with one entry
example usage:
```python
f = ind.to_func()
y = f(1.0, 2.0)
```

`to_numpy` -> f:
- f expects the same number of arguments as the graph has input
- each argument must be a single numpy array of identical lengths
- f returns a list of arrays, one for each output; each array will have the same length as all input arrays
- if only one output is present it'll return only it's output array; _not_ a list with one array
example usage:
```python
f = ind.to_numpy()
y = f(np.array([1.0, 1.1]), np.array([2.0, 2.1]))
```

`to_sympy` -> List[sympy expr] or single sympy expr
- depending on the number of outputs, either a list of sympy expressions is returned or, for a single output, a single sympy expression

closes #287

### DIFF
--- a/cgp/cartesian_graph.py
+++ b/cgp/cartesian_graph.py
@@ -2,7 +2,7 @@ import collections
 import copy
 import math  # noqa: F401
 import re
-from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Set
+from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Set, Union
 
 import numpy as np  # noqa: F401
 
@@ -230,24 +230,35 @@ class CartesianGraph:
                 )
         return func_str
 
-    def to_func(self) -> Callable[[List[float]], List[float]]:
-        """Compile the function(s) represented by the graph.
+    def to_func(self) -> Callable[..., List[float]]:
+        """Create a Python callable implementing the function described by
+        this graph.
 
-        Generates a definition of the function in Python code and
-        executes the function definition to create a Callable.
+        The returned callable expects as many arguments as the number
+        of inputs defined in the genome. The function returns a tuple
+        with length equal to the number of outputs defined in the
+        genome. For convenience, if only a single output is defined
+        the function will *not* return a tuple but only its first
+        element.
 
         Returns
         -------
         Callable
-            Callable executing the function(s) represented by the graph.
+
         """
         self._format_output_str_of_all_nodes()
         s = ", ".join(node.output_str for node in self.output_nodes)
         func_str = f"""\
-def _f(x):
+def _f(*x):
     if len(x) != {self._n_inputs}:
         raise ValueError(f'input has length {{len(x)}}, expected {self._n_inputs}')
-    return [{s}]
+
+    res = [{s}]
+
+    if len(res) == 1:
+        return res[0]
+    else:
+        return res
 """
         func_str = self._fill_parameter_values(func_str)
         exec(func_str, {**globals(), **CUSTOM_ATOMIC_OPERATORS}, locals())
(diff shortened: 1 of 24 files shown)


mschmidt87:
I think this branch needs to be rebased on master to pull in the changes in the continuous integration and make the checks run.

jakobj:
> I think this branch needs to be rebased on master to pull in the changes in the continuous integration and make the checks run.

i rebased, thanks! ready for review :)

jakobj:
ping @mschmidt87 @HenrikMettler

cgp/cartesian_graph.py
```diff
@@ -263,46 +274,56 @@ def _format_output_str_numpy_of_all_nodes(self):
...
-        """Compile the function(s) represented by the graph to NumPy
-        expression(s).
+    def to_numpy(self) -> Callable[..., List[np.ndarray]]:
+        """Create a NumPy-array-compatible Python callable implementing the
+        function described by this graph.
 
-        Generates a definition of the function in Python code and
-        executes the function definition to create a Callable
-        accepting NumPy arrays.
+        The function expects as many arguments as the number of inputs
```
HenrikMettler, line 281: `The returned function` or `The returned callable` to make it clear you are talking about the output and not this function?

HenrikMettler: (actually similar as you did below for the torch instance)

cgp/cartesian_graph.py
```diff
@@ -263,46 +274,56 @@ def _format_output_str_numpy_of_all_nodes(self):
...
+        return res
 """
         func_str = self._fill_parameter_values(func_str)
         exec(func_str, {**globals(), **CUSTOM_ATOMIC_OPERATORS}, locals())
 
         return locals()["_f"]
 
     def to_torch(self) -> "torch.nn.Module":
-        """Compile the function(s) represented by the graph to a Torch class.
+        """Create a Torch class instance implementing the function defined by
```
HenrikMettler, line 316: `Create a torch nn module` ?

cgp/cartesian_graph.py
```diff
@@ -372,15 +398,25 @@ def to_sympy(self, simplify: Optional[bool] = True) -> List["sympy_expr.Expr"]:
...
         Returns
         ----------
-        List[sympy.core.expr.Expr]
-            List of SymPy expressions.
+        List[sympy.core.expr.Expr] or sympy.core.expr.Expr
+            List of SymPy expressions or single expression.
+
         """
+
+        def possibly_unpack(
```
HenrikMettler, line 406: for my taste the code is more complicated with this helper function, than with two `if len(sympy_exprs) == 1:` below, but I guess it is a matter of taste... :) 
(also out of topic here, but do you think the option `simplify = False` is actually ever used in practise?)

HenrikMettler (approved):
Just two minor formulations which might be improved

cgp/cartesian_graph.py
```diff
@@ -230,24 +230,35 @@ def _fill_parameter_values(self, func_str: str) -> str:
...
 
-    def to_func(self) -> Callable[[List[float]], List[float]]:
-        """Compile the function(s) represented by the graph.
+    def to_func(self) -> Callable[..., List[float]]:
+        """Create a Python callable implementing the function described by
+        this graph.
 
-        Generates a definition of the function in Python code and
-        executes the function definition to create a Callable.
+        The function expects as many arguments as the number of inputs
```
mschmidt87, line 237: Same as below indicated by @HenrikMettler : To be clear and consistent with the wording, I would use "The returned Callable" here and in the rest of the docstring.

cgp/cartesian_graph.py
mschmidt87: I agree with @HenrikMettler

cgp/cartesian_graph.py
```diff
@@ -397,11 +433,11 @@ def to_sympy(self, simplify: Optional[bool] = True) -> List["sympy_expr.Expr"]:
...
-            return sympy_exprs
+            return possibly_unpack(sympy_exprs)
         else:  # simplify expression if desired and possible
             for i, expr in enumerate(sympy_exprs):
                 try:
                     sympy_exprs[i] = expr.simplify()
                 except TypeError:
                     RuntimeWarning(f"SymPy could not simplify expression: {expr}")
-            return sympy_exprs
+            return possibly_unpack(sympy_exprs)
```
mschmidt87, line 443: The return statements can be outside of the if/else-branch and then you only need what is currently the else branch,

mschmidt87: like so:
```python
if simplify:
    for i, expr in enumerate(sympy_exprs):
                try:
                    sympy_exprs[i] = expr.simplify()
                except TypeError:
                    RuntimeWarning(f"SymPy could not simplify expression: {expr}")
return possibly_unpack(sympy_exprs)
```

mschmidt87 (changes requested):
Looks good in principle, thanks for doing this, looks like a tedious job. Added some suggestions.

cgp/cartesian_graph.py
jakobj: i've packed this into a function to avoid a comment, hoping the function name would show intention. ideas for other names?

`simplify=False` can be quite helpful, but it's mainly for debugging, not production i think.

jakobj:
thanks @mschmidt87 @HenrikMettler for your comments, i've included most of them. please have another look.

cgp/cartesian_graph.py
mschmidt87: I would find a comment without a function easier to understand, to be honest.

jakobj:
ok, simplified the code in question @mschmidt87

also squashed all commits, let's see what CI thinks

mschmidt87 (approved):

PR #295, opened 2021-03-23
Replace min_fitness by termination_fitness

### LINKED ISSUE #289: Rename "min_fitness" in evolve_params
Given the user experience in #288 I think the variable name `min_fitness` in the `evolve_params` is confusing. I think we should rename it to sthg like `target_fitness`.

mschmidt87:
I discovered that `min_fitness` is currently a positional argument. Shouldn't it be optional? Why is the user forced to always set it?
I made it optional in this PR, but if there is a good reason to have it as a positional arg, I am happy to revert this change.

I didn't completely remove `min_fitness` to not break the API. The idea is to remove it in the next major release (1.0 ?).

Add DeprecationWarning for min_fitness
Add assertion that min_fitness and termination_fitness are not both defined
Make both args optional
Add test

Closes #289

### DIFF
--- a/cgp/hl_api.py
+++ b/cgp/hl_api.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Callable, Optional
 
 import numpy as np
@@ -11,7 +12,8 @@ def evolve(
     pop: Population,
     objective: Callable[[IndividualBase], IndividualBase],
     ea: MuPlusLambda,
-    min_fitness: float,
+    min_fitness: float = np.inf,
+    termination_fitness: float = np.inf,
     max_generations: int = np.inf,
     max_objective_calls: int = np.inf,
     print_progress: Optional[bool] = False,
@@ -33,6 +35,10 @@ def evolve(
         `initialize_fitness_parents` and `step` method.
     min_fitness : float
         Minimum fitness at which the evolution is stopped.
+        Warning: This argument is deprecated and will be removed in the 0.4
+        release. Please use `termination_fitness` instead.
+    termination_fitness : float
+        Minimum fitness at which the evolution is terminated
     max_generations : int
         Maximum number of generations.
         Defaults to positive infinity.
@@ -50,6 +56,23 @@ def evolve(
     -------
     None
     """
+    if np.isfinite(min_fitness) and np.isfinite(termination_fitness):
+        raise RuntimeError(
+            "Both `min_fitness` and `termination_fitness` have been set. The "
+            "`min_fitness` argument is deprecated and will be removed in the 0.4 "
+            "release. Please use `termination_fitness` instead."
+        )
+
+    if np.isfinite(min_fitness):
+        warnings.warn(
+            DeprecationWarning(
+                "The `min_fitness` argument is deprecated and "
+                "will be removed in the 0.4 release. Please use "
+                "`termination_fitness` instead."
+            )
+        )
+        termination_fitness = min_fitness
+
     if np.isinf(max_generations) and np.isinf(max_objective_calls):
         raise ValueError("Either max_generations or max_objective_calls must be finite.")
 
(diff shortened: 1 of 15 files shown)


cgp/hl_api.py
```diff
@@ -50,6 +56,23 @@ def evolve(
...
+                "The `min_fitness` argument is deprecated and "
+                "will be removed in the next major release. Please use "
+                "`termination_fitness` instead."
+            )
+        )
+    assert (min_fitness == np.inf) or (termination_fitness == np.inf), (
+        "Both `min_fitness` and `termination_fitness` have been set. The `min_fitness` argument "
+        "is deprecated and will be removed in the next major release. Please use "
+        "`termination_fitness` instead."
+    )
```
jakobj, line 71:
```suggestion
    if np.isfinite(min_fitness)) and np.isfinite(termination_fitness):
        throw RuntimeError(
                "Both `min_fitness` and `termination_fitness` have been set. The `min_fitness` argument "
                "is deprecated and will be removed in the next major release. Please use "
                "`termination_fitness` instead."
         )
```

examples/example_evo_regression.py
```diff
@@ -134,9 +134,9 @@ def evolution(f_target):
         "primitives": (cgp.Add, cgp.Sub, cgp.Mul, cgp.Div, cgp.ConstantFloat),
     }
 
-    ea_params = {"n_offsprings": 10, "tournament_size": 2, "mutation_rate": 0.03, "n_processes": 2}
+    ea_params = {"n_offsprings": 10, "tournament_size": 2, "mutation_rate": 0.03, "n_processes": 1}
```
jakobj, line 137: why change the number of processes here?

jakobj (changes requested):
nice work, thanks!

i agree that `termination_fitness` can be an optional argument. however, we should have a check that at least one of the stopping criteria is set, right? this would avoid potential confusion.

i've also been thinking of making all keywords except `population`, `objective` and `ea` keyword only, but maybe that's for a seperate PR. what do you think?

cgp/hl_api.py
```diff
@@ -33,6 +35,10 @@ def evolve(
         `initialize_fitness_parents` and `step` method.
     min_fitness : float
         Minimum fitness at which the evolution is stopped.
+        Warning: This argument is deprecated and will be removed in the next
+        major release. Please use `termination_fitness` instead.
```
jakobj, line 39: i would not say major release. i think before 1.0 we can have this deprecation warning in place for one minor release, meaning we would remove this argument in 0.4.

mschmidt87: Good point, I changed it.

examples/example_evo_regression.py
mschmidt87: By mistake. I reverted that change.

cgp/hl_api.py
mschmidt87: Applied your suggestion.

mschmidt87:
> i agree that `termination_fitness` can be an optional argument. however, we should have a check that at least one of the stopping criteria is set, right? this would avoid potential confusion.
> 
> also, if the argument is optional and we always provide `max_generations` in the examples, could we not just remove it from those?
> 
> i've also been thinking of making all keywords except `population`, `objective` and `ea` keyword only, but maybe that's for a seperate PR. what do you think?

With "one of the stopping criteria" you mean `termination_fitness`, `max_generations` or `max_objective_calls`?
I think the two latter ones are different from `termination_fitness` because they guarantee that we stop at some point, but `termination_fitness` doesn't. That is why already do that check, but only for them.

So, I think that `termination_fitness` really is optional.

jakobj (approved):
great stuff, thanks! :rocket: 

now we just need travis to be happy.

PR #299, opened 2021-04-13
Testing `n_columns=0`

### LINKED ISSUE #283: Make sure `n_columns = 0` works properly
This came up during review of #281. We should add a test if none exists yet.

HenrikMettler:
So, weirdly `n_columns = 0` works in the minimal example and in a small test in population it gives the expected sympy expression. However a lower level test fails when calling directly `CartesianGraph(genome).to_sympy()`. Any ideas why? 
Also imo we should rather deprecate the `n_columns=0` or `n_rows=0` functionality, since they don't make much sense anyway. 
Closes #283

### DIFF
--- a/test/test_genome.py
+++ b/test/test_genome.py
@@ -750,3 +750,19 @@ def test_parameters_numpy_array_consistency():
 
     assert genome._parameter_names_to_values["<p1>"] == pytest.approx(1.1)
     assert genome._parameter_names_to_values["<p2>"] == pytest.approx(1.2)
+
+
+def test_ncolumns_zero(rng):
+
+    genome_params = {
+        "n_inputs": 1,
+        "n_outputs": 1,
+        "n_columns": 0,
+        "n_rows": 1,
+        "primitives": (cgp.Mul, cgp.Sub, cgp.Add, cgp.ConstantFloat),
+    }
+    genome = cgp.Genome(**genome_params)
+    genome.randomize(rng)
+
+    CartesianGraph(genome).to_func()
+    CartesianGraph(genome).to_numpy()
--- a/test/test_population.py
+++ b/test/test_population.py
@@ -105,3 +105,18 @@ def test_individual_init_expects_callable(population_params, genome_params):
     # passing a list as individual_init fails
     with pytest.raises(TypeError):
         cgp.Population(**population_params, genome_params=genome_params, individual_init=[])
+
+
+def test_ncolumns_zero(population_params):
+    sympy = pytest.importorskip("sympy")
+    genome_params = {
+        "n_inputs": 1,
+        "n_outputs": 1,
+        "n_columns": 0,
+        "n_rows": 1,
+        "primitives": (cgp.Mul, cgp.Sub, cgp.Add, cgp.ConstantFloat),
+    }
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    for ind in pop:
+        sympy_expr = ind.to_sympy()
+        assert sympy_expr == sympy.sympify("x_0")


test/test_genome.py
```diff
@@ -750,3 +750,17 @@ def test_parameters_numpy_array_consistency():
 
     assert genome._parameter_names_to_values["<p1>"] == pytest.approx(1.1)
     assert genome._parameter_names_to_values["<p2>"] == pytest.approx(1.2)
+
+
+def test_ncolumns_zero():
```
jakobj, line 755:
```suggestion
def test_ncolumns_zero(rng):
```

test/test_genome.py
```diff
@@ -750,3 +750,17 @@ def test_parameters_numpy_array_consistency():
...
+    genome_params = {
+        "n_inputs": 1,
+        "n_outputs": 1,
+        "n_columns": 0,
+        "n_rows": 1,
+        "primitives": (cgp.Mul, cgp.Sub, cgp.Add, cgp.ConstantFloat),
+    }
+    genome = cgp.Genome(**genome_params)
+    sympy_expression = CartesianGraph(genome).to_sympy(simplify=False)
+    print(sympy_expression)
```
jakobj, line 766:
```suggestion
    genome = cgp.Genome(**genome_params)
    genome.randomize(rng)
    
    CartesianGraph(genome).to_func()
    CartesianGraph(genome).to_numpy()
    CartesianGraph(genome).to_torch()
    CartesianGraph(genome).to_sympy()
```

test/test_population.py
```diff
@@ -105,3 +106,18 @@ def test_individual_init_expects_callable(population_params, genome_params):
...
+        "n_inputs": 1,
+        "n_outputs": 1,
+        "n_columns": 0,
+        "n_rows": 1,
+        "primitives": (cgp.Mul, cgp.Sub, cgp.Add, cgp.ConstantFloat),
+    }
+    pop = cgp.Population(**population_params, genome_params=genome_params)
+    sympy_expr = pop.champion.to_sympy()
+    expected_sympy_expr = symbols('x_0')
+    assert sympy_expr[0] == expected_sympy_expr
```
jakobj, line 123:
```suggestion
    for ind in pop:
        sympy_expr = ind.to_sympy()
        assert sympy_expr[0] == sympy.sympify("x_0")
```

test/test_population.py
```diff
@@ -3,6 +3,7 @@
 
 import cgp
 from cgp.genome import ID_INPUT_NODE, ID_NON_CODING_GENE, ID_OUTPUT_NODE
+from sympy import symbols
```
jakobj, line 6:
```suggestion
```

test/test_population.py
```diff
@@ -105,3 +106,18 @@ def test_individual_init_expects_callable(population_params, genome_params):
     # passing a list as individual_init fails
     with pytest.raises(TypeError):
         cgp.Population(**population_params, genome_params=genome_params, individual_init=[])
+
+
+def test_ncolumns_zero(population_params):
+
```
jakobj, line 112:
```suggestion
    sympy = pytest.importorskip("sympy")
```

jakobj (changes requested):
great, thanks for working on this! :rocket: 

i've added a few inline comments that should be addressed.

regarding your question: the low-level error was generated because the genome was not initialized. in also, i think we should not explicitly forbid `n_columns=0` if our library does the right thing, why not?

HenrikMettler:
Ok, i am a bit lost. Somehow the test doesn't pass on the python 3.7 testsuite, however I can't reproduce the error locally. Any ideas why?

mschmidt87:
I also tried to reproduce the error locally without success. I also tried to google the error and found this link which might (or might not) be helpful: https://stackoverflow.com/questions/66640404/basic-loop-with-sympy-gives-me-this-error-message-symbol-object-is-not-subsc
Another, a bit brute-force option, is to do:
- remove python 3.7 temporarily to see whether 3.6 and 3.8 throw the same error in the CI
- add some debug output to the test and rerun the CI.
Also, I did create a new conda env with python 3.7 and installed the library, so I basically followed the steps in the CI, but I didn't confirm that the configuration is really exactly identical with the CI environment. So, you could/should also do that with your local environment.

jakobj:
have you tried rebasing on the most recent master? one of the recent PRs has removed the necessity to do `sympy_expr[0]` if only one expression is encoded in the genome, as it is automatically unpacked. i think this might be where the error comes from.

HenrikMettler:
ah yes, that was it! my bad, thanks for suggesting

mschmidt87 (approved):

jakobj (approved):
looking good! :+1:

and merging

Issue #251, opened 2020-11-11
Multi-version documentation


mschmidt87:
We want to publish the documentation both for the release version and the latest version from master.

mschmidt87:
After several attempts to get this working, I am giving up on trying to make this run with publishing the docs on github pages. In principle, `sphinx-multiversion` is the way to go. I can also make it run locally (kind of, keep running into issues there as well), however, making this in the CI (travis and github actions) has proven impossible (for me, at least). See #298 .

Another issue is that `sphinx-multiversion` does not seem well maintained, since the maintainer declared it "compete" (see https://github.com/sphinx-contrib/sphinxcontrib-versioning/issues/77#issuecomment-737481668) and seems to ignore Pull-requests by other developers.

Therefore, I will migrate our documentation to readthedocs which supports multiple versions.

mschmidt87:
Reopening here, because I will redirect the link in the README to the new docs and do some maintenance behind the scenes.

mschmidt87:
Currently, the build for 0.2.0 is failing, (I believe) because the `.readthedocs.yaml` config file is missing. I guess the only way to fix this would be to push that file to the release branch/tag. This is a bit dirty because it will alter the release retroactively, but as far as I understand, inevitable. @jakobj

jakobj:
well, we certainly don't want to rewrite the history of the whole repo ;) one way would be to create a new, permanent 0.2.1 release branch which contains the relevant files. alternatively we could start providing multiversion doc from 0.3.0 onwards. i have a feeling 0.4.0 might not be too far in the future ;)

mschmidt87:
No, I also thought that creating a release branch would be the best option. I would do that rather than waiting for the 0.3.0 release.

PR #300, opened 2021-04-14
Add readthedocs config file and requirements for docs

mschmidt87:
Activate building the documentation to migrate from github pages to readthedocs. This supports multi-version building. After merging this PR, I will create another PR to replace the link in the README and then close the associated ticket #251 after that.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,15 @@
+version: 2
+
+# Build from the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+
+# Explicitly set the version of Python and its requirements
+python:
+  version: 3.8
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - extra
+        - doc
--- a/extra-requirements.txt
+++ b/extra-requirements.txt
@@ -3,6 +3,7 @@ matplotlib ~= 3.2.1
 scipy ~= 1.4.1
 sympy ~= 1.6.1
 torch ~= 1.5.0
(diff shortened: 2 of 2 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,15 @@
+version: 2
+
+# Build from the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+
+# Explicitly set the version of Python and its requirements
+python:
+  version: 3.8
```
HenrikMettler, line 9: Shouldn't python version 3.6 and 3.7 also be supported?

extra-requirements.txt
```diff
@@ -3,6 +3,7 @@ matplotlib ~= 3.2.1
 scipy ~= 1.4.1
 sympy ~= 1.6.1
 torch ~= 1.5.0
+gym
```
HenrikMettler, line 6: 0.17.0? (at least that's what I am using ^^)

HenrikMettler (approved):

.readthedocs.yaml
mschmidt87: This only specifies the Python version used to build the docs, so I went for the most recent supported version.

extra-requirements.txt
mschmidt87: I went into some weird dependency problems, related to the pillow package. Removing the specifier made it work.

HenrikMettler:
Looks good, thanks a lot for taking care of this!

PR #311, opened 2021-04-22
Support Python 3.9, drop Python 3.6 support, update requirements

### LINKED ISSUE #303: Support Python 3.9 and drop support of python 3.6
Python 3.9 was released more than 6 months ago, so we should make hal-cgp compatible with it. This probably doesn't require much, simply adding it to the CI might be enough.

mschmidt87:
This PR makes hal-cgp compatible with Python 3.9 and removes support for python 3.6. 

To that end, several dependencies are updated, most importantly numpy  to version 1.20 which supports python 3.9. That led to a cascade of errors signaled by mypy which are fixed in this PR:
- many are mismatches between return types and variable types
- numpy now supports type hints internally. This leads to a problem with our type hints which specify a dtype for arrays. As far as I understand, before, mypy just did not check those at all. Now, mypy checks them, but numpy does not support hints for dtypes yet, so this threw an error. I fixed it by removing all dtypes from the array type ints.
- mypy now correctly complained about the type hint in the high-level api that defines e.g. `max_generations` as an integer but then sets its default value to `np.inf` which is a float. I solved this by setting its default value to the highest representable integer value. This works, but is a bit ugly. If you have a better idea, please comment.

Note: Github currently reports the checks to be in process because it's still waiting for the python 3.6 check which is no longer done in this PR. I will change this after getting approval from the reviewers to be able to merge the PR then.
Closes #303  and #314

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -154,7 +154,7 @@ class MuPlusLambda:
 
         if self.local_search is not None:
             assert isinstance(pop.champion.fitness, float)
-            prev_avg_fitness: float = np.mean([ind.fitness for ind in combined])
+            prev_avg_fitness = np.mean([ind.fitness for ind in combined])
 
             combined_copy = [ind.copy() for ind in combined]
 
@@ -172,7 +172,7 @@ class MuPlusLambda:
 
             combined = self._sort(new_combined)
 
-            avg_fitness: float = np.mean([ind.fitness for ind in combined])
+            avg_fitness = np.mean([ind.fitness for ind in combined])
             if prev_avg_fitness > avg_fitness:
                 raise RuntimeError(
                     "The average fitness decreased after executing the local search. This"
--- a/cgp/genome.py
+++ b/cgp/genome.py
@@ -587,7 +587,7 @@ class Genome:
 
     def _select_gene_indices_for_mutation(
         self, mutation_rate: float, rng: np.random.RandomState
-    ) -> List[int]:
+    ) -> np.ndarray:
         """Selects the gene indices for mutations
 
         Parameters
@@ -599,7 +599,7 @@ class Genome:
 
         Returns
         ----------
-        selected_gene_indices: np.ndarray
+        np.ndarray
             indices of the genes selected for mutation.
         """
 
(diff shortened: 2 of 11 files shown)


cgp/hl_api.py
```diff
@@ -73,15 +74,18 @@ def evolve(
...
+            " value smaller than the largest representable integer."
+        )
 
     ea.initialize_fitness_parents(pop, objective)
     if callback is not None:
         callback(pop)
 
     # perform evolution
-    max_fitness = np.finfo(np.float).min
+    max_fitness = np.finfo(np.float_).min
```
HenrikMettler, line 88: Probably just a minor detail, but the DepreceationWarning suggested to use `np.float64` or `float` here, any specific reason why `np.float_`

cgp/ea/mu_plus_lambda.py
```diff
@@ -154,7 +154,7 @@ def step(
 
         if self.local_search is not None:
             assert isinstance(pop.champion.fitness, float)
-            prev_avg_fitness: float = np.mean([ind.fitness for ind in combined])
+            prev_avg_fitness = np.mean([ind.fitness for ind in combined])
```
HenrikMettler, line 157: Why did you have to drop the type hints here and below?

cgp/genome.py
```diff
@@ -599,7 +599,7 @@ def _select_gene_indices_for_mutation(
 
         Returns
         ----------
-        selected_gene_indices: np.ndarray
+        np.ndarray
```
HenrikMettler, line 602: Thx for fixing, that was my mistake i think

HenrikMettler (commented):
Thanks a lot for doing this :) Looks godd, just a few comments before merging

HenrikMettler:
Does this also address #314 ?

cgp/hl_api.py
mschmidt87: Ah I see, I didn't see that DeprecationWarning. Okay, I will use `float` instead. In fact, `np.float_` is just an alias for `np.float64`. Thanks.

mschmidt87:
> Does this also address #314 ?

Actually yes, I believe so.

jakobj (approved):
great thanks for taking care of this! :+1:

some changes seem a bit unrelated to the PR title, maybe this can be easily solved by adapting the title? also please rewrite history to make different changes explicit if not already done.

PR #316, opened 2021-04-29
Use Optional[<type>] instead of Union[<type>, None]

### LINKED ISSUE #297: Use `Optional[<type>]` type hint instead of `Union[<type>, None]`
Currently we are using `Union[<type>, None]` for many cases in which we actually should use `Optional[<type>]` as more explicitly communicates intent, e.g., https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/cgp/individual.py#L37. This should be fixed.

jakobj:
As the title promises :)

closes #297


HenrikMettler (approved):
Nice, thanks! Like a good pokemon trainer, you caugth them all! (i think)

PR #321, opened 2021-05-14
Fix warnings in testsuite

### LINKED ISSUE #304: Fix warnings in testsuite
Currently, there are two sources of warnings when running the testsuite: 1) using the deprecated `min_fitness` argument 2) the named tuple `TestParams` in `test_individual.py`.

Both should be fixed.

jakobj:
- avoid using `min_fitness`
- avoid global variables which start with `Test`

closes #304


mschmidt87 (approved):
LGTM

PR #322, opened 2021-05-14
Avoid global rngs in testsuite

### LINKED ISSUE #305: Avoid dependence on global NumPy rng in testsuite
The testsuite provides its own rng (https://github.com/Happy-Algorithms-League/hal-cgp/blob/master/test/conftest.py#L13). To avoid difficult to reproduce testsuite errors, we should always prefer this rng over the global one (`np.random`).

jakobj:
- replace numpy global rng
- replace torch global rng

closes #305

### DIFF
--- a/test/test_hl_api.py
+++ b/test/test_hl_api.py
@@ -12,8 +12,6 @@ def _objective_test_population(individual, rng_seed):
     if not individual.fitness_is_None():
         return individual
 
-    np.random.seed(rng_seed)
-
     n_function_evaluations = 100
 
     f_graph = individual.to_func()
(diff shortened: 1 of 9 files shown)


test/test_hl_api.py
```diff
@@ -12,16 +12,15 @@ def _objective_test_population(individual, rng_seed):
...
-
     n_function_evaluations = 100
 
     f_graph = individual.to_func()
 
     def f_target(x):  # target function
         return x[:, 0] - x[:, 1]
 
-    x = np.random.normal(size=(n_function_evaluations, 2))
+    rng = np.random.RandomState(seed=rng_seed)
```
mschmidt87, line 22: Why don't you use the fixture here?

test/test_hl_api.py
```diff
@@ -81,6 +78,8 @@ def _objective(individual):
         if not individual.fitness_is_None():
             return individual
 
+        rng = np.random.RandomState(rng_seed)
```
mschmidt87, line 81: same here: why not the fixture?

mschmidt87 (changes requested):
Good work. But why don't you consistently use the `rng` fixture everywhere?

test/test_hl_api.py
jakobj: I think the intention here is to use the same random number stream in every call to the objective such that the fitness does *not* depend on the random number stream. I think if i'd use the fixture, then rngs are shared across objective calls, right? same reason below

jakobj:
ping @mschmidt87

mschmidt87 (approved):
LGTM. I think zou need to push once again to trigger the new testsuite including Python 3.9

PR #330, opened 2021-06-24
Raise RuntimeError when trying to access uninitialized dna

### LINKED ISSUE #306: Using uninitialized genomes should throw an exception
Apparently, it's possible to call, for example, `to_func()` on an uninitialized genome:

```python
    CartesianGraph(genome).to_func()
```

which gives rise to difficult to understand errors. We should just raise an error in such a case.

jakobj:
like the title suggests.

closes #306


HenrikMettler (approved):
LGTM. Thanks for taking care of this :) :rocket:

PR #332, opened 2021-06-24
Provide somewhat sensible defaults for all hyperparameters

### LINKED ISSUE #307: Provide sensible defaults for all(most?) hyperparameters
To streamline the (common) user experience, I think we should provide defaults for most hyperparameters of the library. The defaults should be taken from the CGP literature, possibly with some exceptions, e.g., ~10 columns, rather than thousands.

I believe this would significantly improve the user experience by not having to worry about many hyperparameter settings, but rather starting from a reasonable default, figuring out which should be changed.

Thoughts @mschmidt87 @HenrikMettler ?

jakobj:
as the title suggests

we could also simplify the minimal example a lot with these new defaults. what do you think?

closes #307

### DIFF
--- a/cgp/genome.py
+++ b/cgp/genome.py
@@ -26,30 +26,30 @@ class Genome:
 
     def __init__(
         self,
-        n_inputs: int,
-        n_outputs: int,
-        n_columns: int,
-        n_rows: int,
-        primitives: Tuple[Type[Node], ...],
+        n_inputs: int = 1,
+        n_outputs: int = 1,
+        n_columns: int = 128,
+        n_rows: int = 1,
+        primitives: Optional[Tuple[Type[Node], ...]] = None,
         levels_back: Optional[int] = None,
     ) -> None:
         """Init function.
 
         Parameters
         ----------
-        n_inputs : int
-            Number of inputs of the function represented by the genome.
-        n_outputs : int
-            Number of outputs of the function represented by the genome.
-        n_columns : int
-            Number of columns in the representation of the genome.
-        n_rows : int
-            Number of rows in the representation of the genome.
-        primitives : Tuple[Type[Node], ...]
-           Tuple of primitives that the genome can refer to.
-        levels_back : Optional[int]
+        n_inputs : int, optional
+            Number of inputs of the function represented by the genome. Defaults to 1.
+        n_outputs : int, optional
+            Number of outputs of the function represented by the genome. Defaults to 1.
+        n_columns : int, optional
+            Number of columns in the representation of the genome. Defaults to 12.
+        n_rows : int, optional
+            Number of rows in the representation of the genome. Defaults to 1.
+        primitives : Tuple[Type[Node], ...], optional
+           Tuple of primitives that the genome can refer to. Defaults to (+, -, *, 1.0).
+        levels_back : Optional[int], optional
             Maximal column distance of inputs to an internal node. If
-            set to `None`, no restrictions are used. Defaults to None
+            set to `None`, no restrictions are used. Defaults to None.
 
         """
         if n_inputs <= 0:
@@ -76,7 +76,19 @@ class Genome:
             raise ValueError("levels_back can not be larger than n_columns")
         self._levels_back = levels_back
 
-        self._primitives = Primitives(primitives)
+        if primitives is None:
+            # we need to delay this import to avoid circular imports: node_impl
+            # -> node -> node_validation -> genome
+            from .node_impl import (  # delayed to avoid circular imports
+                Add,
+                ConstantFloat,
+                Mul,
+                Sub,
+            )
+
+            self._primitives = Primitives((Add, Sub, Mul, ConstantFloat))
+        else:
+            self._primitives = Primitives(primitives)
         self._length_per_region = (
             1 + self._primitives.max_arity
         )  # one function gene + multiple address genes
--- a/cgp/hl_api.py
+++ b/cgp/hl_api.py
@@ -8,39 +8,37 @@ from .population import Population
 
 
 def evolve(
-    pop: Population,
     objective: Callable[[IndividualBase], IndividualBase],
-    ea: MuPlusLambda,
+    pop: Population = Population(),
+    ea: MuPlusLambda = MuPlusLambda(),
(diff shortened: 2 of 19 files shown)


cgp/genome.py
```diff
@@ -26,30 +26,30 @@ class Genome:
...
-        n_inputs: int,
-        n_outputs: int,
-        n_columns: int,
-        n_rows: int,
-        primitives: Tuple[Type[Node], ...],
+        n_inputs: int = 1,
+        n_outputs: int = 1,
+        n_columns: int = 12,
+        n_rows: int = 1,
+        primitives: Optional[Tuple[Type[Node], ...]] = None,
```
HenrikMettler, line 33:
```python
primitives: Optional[Tuple[Type[Node], ...]] = (Add, Sub, Mul, ConstantFloat),
```
to be consistent with the doc string

cgp/genome.py
```diff
@@ -76,7 +76,17 @@ def __init__(
             raise ValueError("levels_back can not be larger than n_columns")
         self._levels_back = levels_back
 
-        self._primitives = Primitives(primitives)
+        if primitives is None:
+            from .node_impl import (  # delayed to avoid circular imports
```
HenrikMettler, line 80: Why would you have circular imports?

cgp/genome.py
```diff
@@ -26,30 +26,30 @@ class Genome:
 
     def __init__(
         self,
-        n_inputs: int,
-        n_outputs: int,
-        n_columns: int,
-        n_rows: int,
-        primitives: Tuple[Type[Node], ...],
+        n_inputs: int = 1,
```
HenrikMettler, line 29: Line 9 (sorry cant write there somehow): 
from .node_impl import Add, ConstantFloat, Mult, Sub

cgp/hl_api.py
```diff
@@ -9,44 +9,43 @@
...
 def evolve(
-    pop: Population,
     objective: Callable[[IndividualBase], IndividualBase],
-    ea: MuPlusLambda,
+    pop: Population = Population(),
+    ea: MuPlusLambda = MuPlusLambda(),
     min_fitness: float = np.inf,
     termination_fitness: float = np.inf,
-    max_generations: int = np.iinfo(np.int64).max,
+    max_generations: int = 1000,
```
HenrikMettler, line 17: Not sure this change is a good idea. If the user wants to use `max_objective_calls` this default will still be used (unless overwritten)

cgp/hl_api.py
```diff
@@ -9,44 +9,43 @@
...
-        Minimum fitness at which the evolution is terminated
-    max_generations : int
+    termination_fitness : float, optional
+        Minimum fitness at which the evolution is terminated. Defaults to positive infinity.
+    max_generations : int, optional
         Maximum number of generations.
-        Defaults to largest representable integer.
-        Either this or `max_objective_calls` needs to be set to a value less than that.
-    max_objective_calls: int
+        Defaults to 1000.
```
HenrikMettler, line 43: see above

cgp/hl_api.py
```diff
@@ -55,7 +54,8 @@ def evolve(
 
     Returns
     -------
-    None
+    Population
```
HenrikMettler, line 57: I think this is a good idea, but should we maybe put it in a separate PR?

cgp/population.py
```diff
@@ -13,24 +13,24 @@ class Population:
...
-            Number of parent individuals.
-        seed : int
-            Seed for internal random number generator.
+        n_parents : int, optional
+            Number of parent individuals. Defaults to 4.
+        seed : int, optional
+            Seed for internal random number generator. Defaults to 1234.
         genome_params : dict
-            Parameters for the genomes of the population's individuals.
+            Parameters for the genomes of the population's individuals. Defaults to None.
```
HenrikMettler, line 30: Hmm might it be more user-friendly to have the defaults here instead of nested in `Genome()`?

HenrikMettler: At least the defaults should be documented here as well

examples/example_caching.py
```diff
@@ -98,7 +98,7 @@ def objective(individual):
 # Finally, we call the `evolve` method to perform the evolutionary search.
 
 
-cgp.evolve(pop, objective, ea, **params["evolve_params"], print_progress=True)
+cgp.evolve(objective, pop, ea, **params["evolve_params"], print_progress=True)
```
HenrikMettler, line 101: I think we should be consistent in calling it `objective` or `obj` in all examples (but maybe this can be a separate maintenance PR)

test/test_utils.py
```diff
@@ -66,7 +66,7 @@ def objective(ind):
         def recording_callback(pop):
             history["fitness_champion"].append(pop.champion.fitness)
 
-        cgp.evolve(pop, objective, ea, **evolve_params, callback=recording_callback)
+        cgp.evolve(objective, pop, ea, **evolve_params, callback=recording_callback)
```
HenrikMettler, line 69: Same here: Name them all `obj` or `objective`

HenrikMettler (changes requested):
Thanks for working on this :) Some points on parameter-values and user-friendliness.

cgp/genome.py
HenrikMettler: Understood, but please describe circle

examples/example_caching.py
HenrikMettler: see separate issue #336

HenrikMettler:
What a nice opportunity to update the readme ;)

cgp/population.py
jakobj: i avoided documenting the results here but put a reference to where they are defined. if we duplicate this information one part will become outdated quickly.

jakobj:
thanks @HenrikMettler for the review. i've addressed all your comments, rebased on master and squashed the commits. please have another look.

jakobj:
ping @HenrikMettler

HenrikMettler (approved):
LGTM, thanks for taking care of it :) Please squish the `fix formatting` commit

HenrikMettler:
ping @jakobj

PR #334, opened 2021-06-24
Check max_generations/max_objective_calls against maximal integer

### LINKED ISSUE #327: Wrong check for `np.isfinite` in printing progress in hl_api
`max_generations`/`max_objective_calls` are now initialized to `np.iinfo(np.int64).max` and not `np.inf`. The checks for `np.isfinite` for printing progress either on number of generation of number of objective calls are hence wrong. They should instead check for equality with `np.iinfo(np.int64).max`.

Thanks @HenrikMettler for spotting this.

jakobj:
as the title suggests.

closes #327

### DIFF
--- a/cgp/hl_api.py
+++ b/cgp/hl_api.py
@@ -99,13 +99,13 @@ def evolve(
             max_fitness = pop.champion.fitness
 
         if print_progress:
-            if np.isfinite(max_generations):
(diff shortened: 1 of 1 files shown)


HenrikMettler (approved):
LGTM. Thanks for taking care of this :) :rocket:

PR #333, opened 2021-06-24
Raise exception if fitness is None after calls to objectives

jakobj:
as the title suggests. this should help to pinpoint bugs in objective functions.

closes #309

PR #339, opened 2021-11-03
Bugfix argument order in new test

HenrikMettler:
Fixes the bug introduced by merging #333 (with wrong argument order for call to cgp.evolve)


jakobj (approved):
excellent, thanks for the quick fix! :rocket:

Issue #312, opened 2021-04-23
Create an "advanced features" list


jakobj:
It seems to me we have reached the point where some of the awesome advanced features of hal-cgp may be too hidden from the user. We should enrich the README with an overview such that users at least knows that these features exist. Maybe each of these items could then link to a documentation page or, maybe even better, an example. Let's collect all advanced features in this issue first.

Advanced features
- parallel fitness evaluation
- translation of grah into numpy/torch/sympy compatible functions
- individuals with multiple genomes/encoded functions
- hurdles
- custom operators
- local search of parameters (gradient based and evolutionary strategies)
- caching (with functional equivalence checking)
- genome reordering

Please feel free to extend this list @mschmidt87 @HenrikMettler.

PR #341, opened 2021-11-15
Add simple example using NumPy to vectorized function evaluation

jakobj:
It seems we were missing an example of how one would use NumPy to vectorize function evaluations. This PR contains such an example as an extension of our standard example.

### DIFF
--- a/examples/example_evo_regression_numpy.py
+++ b/examples/example_evo_regression_numpy.py
@@ -1,0 +1,197 @@
+"""
+Example for evolutionary regression using NumPy
+===============================================
+
+Example demonstrating the use of Cartesian genetic programming for
+two regression tasks using NumPy to vectorized function evaluations.
+"""
+
+# The docopt str is added explicitly to ensure compatibility with
+# sphinx-gallery.
+docopt_str = """
+   Usage:
+     example_evo_regression_numpy.py [--max-generations=<N>]
+
+   Options:
+     -h --help
+     --max-generations=<N>  Maximum number of generations [default: 1000]
+"""
+
+import functools
+import warnings
+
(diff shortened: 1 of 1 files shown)


jakobj:
ping @HenrikMettler

examples/example_evo_regression_numpy.py
```diff
@@ -0,0 +1,197 @@
...
+        Target function
+
+    Returns
+    -------
+    dict
+        Dictionary containing the history of the evolution
+    Individual
+        Individual with the highest fitness in the last generation
+    """
+    population_params = {"n_parents": 10, "seed": 818821}
```
HenrikMettler, line 120: Is there a reason to not use the default parameterization? (Except for `genome_params['n_inputs] = 2` ?

HenrikMettler (changes requested):
Left a question about parameterization. 
Also did we ever find a solution to the problem of ever increasing number of examples? (I guess implementing #312 would partially answer that ^^)

examples/example_evo_regression_numpy.py
jakobj: no reason except staying close to the "evo_regression" example. i'd like to keep it this way to really make minimal changes to the existing one, but i created a new issue that we should update the examples to use default parameters where possible.

jakobj:
Thanks for the review!

So far we didn't find a good solution. But in general, I think the increase in increased accessibility provided by many examples may outweigh the cost of maintaining them. So far I think it's still manageable.

HenrikMettler (approved):
Good solution with the seperate Issue and I agree with respect to the No. of examples.

PR #344, opened 2021-12-01
Add test for sorting of individuals, fix bug in sorting

### LINKED ISSUE #342: Test that sorting a list of individuals yields the expected results
We currently do not have test for this. We should create one that tests whether the sorting step of the ea (`sorted(<list of individuals>, reverse=True)`) yields the expected results both for single as well as multiple objectives separated by hurdles. In particular, we should also check the case where no individual made it through the first hurdle.

jakobj:
As the title says. Introducing this test unearthed a bug in the sorting of individuals in the case of multiple objectives (used with hurdles) which this PR also fixes: individuals should be sorted according to their fitness values, starting from the _last_ objective, whereas previously they were sorted starting from the _first_ objective. Behind this underlies the assumption that later objectives (in terms of hurdles) are more informative about the quality of an individual.

closes #342

### DIFF
--- a/cgp/individual.py
+++ b/cgp/individual.py
@@ -171,11 +171,11 @@ class IndividualBase:
         return genome.update_parameters_from_numpy_array(params, params_names)
 
     def __lt__(self, other: "IndividualBase") -> bool:
-        for i in range(len(self._fitness)):
+        for i in range(len(self._fitness) - 1, -1, -1):
             this_fitness = self._fitness[i]
             other_fitness = other._fitness[i]
             if this_fitness is None and other_fitness is None:
-                return False
+                continue
             elif this_fitness is not None and other_fitness is None:
                 return False
             elif this_fitness is None and other_fitness is not None:
@@ -185,6 +185,8 @@ class IndividualBase:
             assert other_fitness is not None
             if this_fitness < other_fitness:
                 return True
+            else:
+                return False
 
         return False
 
--- a/cgp/individual.py
+++ b/cgp/individual.py
@@ -171,9 +171,10 @@ class IndividualBase:
         return genome.update_parameters_from_numpy_array(params, params_names)
 
     def __lt__(self, other: "IndividualBase") -> bool:
-        for i in range(len(self._fitness) - 1, -1, -1):
-            this_fitness = self._fitness[i]
-            other_fitness = other._fitness[i]
+        # determine order of self and other by iterating through fitness values
+        # in reversed order; for multiple objectives seperated by hurdles, this
+        # will try to sort by fitness on later objectives first
+        for this_fitness, other_fitness in zip(reversed(self._fitness), reversed(other._fitness)):
             if this_fitness is None and other_fitness is None:
                 continue
             elif this_fitness is not None and other_fitness is None:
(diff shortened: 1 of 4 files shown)


cgp/individual.py
```diff
@@ -171,11 +171,11 @@ def _update_parameters_from_numpy_array(
         return genome.update_parameters_from_numpy_array(params, params_names)
 
     def __lt__(self, other: "IndividualBase") -> bool:
-        for i in range(len(self._fitness)):
+        for i in range(len(self._fitness) - 1, -1, -1):
```
HenrikMettler, line 174: If I understand this correctly this makes it loop backwards (which is correct I think), but why start at `len()-1` and end at `-1` instead of `len` and end at `0`?

HenrikMettler: Ah I guess it is to access the index below. However maybe this should be commented or documented somehow?(personally I had too look up the python documentation to make sure I understand the logic correctly :see_no_evil:

cgp/individual.py
```diff
@@ -185,6 +185,8 @@ def __lt__(self, other: "IndividualBase") -> bool:
             assert other_fitness is not None
             if this_fitness < other_fitness:
                 return True
```
HenrikMettler, line 187: I am a bit torn on this logic now - it makes sense to sort according to the last hurdle, however it would be kinda cool also to have this customable (eg the sum). At least I think we need to properly document this (in addition to the nice comment of #345). I am making #312 high priority.

test/test_ea_mu_plus_lambda.py
```diff
@@ -351,3 +352,36 @@ def objective(ind):
...
+    ind1._fitness = [1, None]
+    ind2 = IndividualSingleGenome([])
+    ind2._fitness = [2, 4]
+    ind3 = IndividualSingleGenome([])
+    ind3._fitness = [3, None]
+    individuals = [ind0, ind1, ind2, ind3]
+    ea = cgp.ea.MuPlusLambda()
+    sorted_individuals = ea._sort(individuals)
+    expected = [[0, 5], [2, 4], [3, None], [1, None]]
+    assert [ind._fitness for ind in sorted_individuals] == expected
```
HenrikMettler, line 387: What happens if the fitness array is implausible? Sthg like: `[[0, 1,5 ], [2, 4,1], [3, None, 2], [1, None, None]]` I guess this should raise an error? but not sure what kind and not sure this is a task for the sorting operator

HenrikMettler (changes requested):
Looks good! Thanks. Left two comments

cgp/individual.py
jakobj: yes, this was a bit too complicated :D simplified it now

cgp/individual.py
jakobj: hmm, not sure what you mean by the sum? added additional documentation to the docstring of `MuPlusLambda`

test/test_ea_mu_plus_lambda.py
jakobj: good question. i think we can catch this somewhere in the `fitness.setter` of `Individual`. will do and add a test

jakobj:
thanks for the comments! addressed them, pls have another look

cgp/individual.py
HenrikMettler: Sorry realised my point made little sense, as this can just be done by the user (had one particular use-case of my self in my mind)

HenrikMettler (changes requested):
Lgtm! Thanks of taking care of this :) 🚀

HenrikMettler (approved):

Issue #350, opened 2022-01-05
Make sure caching decorator's key from numpy evaluation contains not only nan


jakobj:
Currently, two functions which return all `nan` values from the numpy evaluation are assigned identical keys without emitting any warning.

This should be fixed, e.g., by raising an exception if this is detected.

PR #351, opened 2022-01-06
Make sure that at least one value is finite for computing key in caching decorator

jakobj:
As the title suggests: If no finite value is returned from the numpy evaluation, two functions which are indeed different could result in the same key. This PR introduces a check which raises an exception if such a situation is encountered.

fixex #350

### DIFF
--- a/cgp/utils.py
+++ b/cgp/utils.py
@@ -2,6 +2,7 @@ import functools
 import hashlib
 import os
 import pickle
+import warnings
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Type
 
 import numpy as np
@@ -110,8 +111,14 @@ def compute_key_from_numpy_evaluation_and_args(
             rng.uniform(_min_value, _max_value, size=_batch_size)
             for _ in range(ind.genome._n_inputs)
         ]
-        y = f_single(*x)
-        s = np.array_str(np.array(y), precision=15)
+        y = np.array(f_single(*x))
+        if not np.any(np.isfinite(y)):
+            warnings.warn(
+                "no finite output for the chosen input range for the caching decorator."
+                " please choose an appropriate range.",
+                RuntimeWarning,
+            )
+        s = np.array_str(y, precision=15)
     elif isinstance(ind, IndividualMultiGenome):
         f_multi = ind.to_numpy()
         s = ""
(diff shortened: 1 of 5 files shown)


cgp/utils.py
```diff
@@ -108,8 +108,13 @@ def compute_key_from_numpy_evaluation_and_args(
             rng.uniform(_min_value, _max_value, size=_batch_size)
             for _ in range(ind.genome._n_inputs)
         ]
-        y = f_single(*x)
-        s = np.array_str(np.array(y), precision=15)
+        y = np.array(f_single(*x))
+        if not np.any(np.isfinite(y)):
+            raise RuntimeError(
```
HenrikMettler, line 113: I wonder if raising an error is a good solution here. For a user this could cause problems, if during a run suddenly for one individual no finite value can be computed the whole run is terminated. 
Could we just skip calculating a key for such an individual? So that if this happens the individual will just be "ignored" for the decorator and evaluated?

HenrikMettler (commented):
Thank for implementing this. Left a comment on the logic, hope my point is understandable.

cgp/utils.py
jakobj: hmm, i'm not sure we should solve this for the user by default. wouldn't it be better to encourage the user to completely avoid such situations and rather adapt, for example the `min_value, max_value` ranges?

if they explicitly want to protect themselves against such crashes (sounds like a bad idea, but nevertheless) they can use a try except, wrapping the `compute_key...` function:
```python
def my_protected_compute_key(*args, **kwargs):
    try:
        return compute_key_from_numpy_evaluation_and_args(*args, **kwargs)
    except RuntimeError:
        <return a unique identifier which makes sure the cache is not used>
```

this way, the "good practice" is easy (adapt the ranges) while the "bad practice" (continue even though some keys are meaningless) is slightly more difficult.

HenrikMettler: Maybe I am misunderstanding this, but this could happen in (almost) any run, I think? The user has no way of a priori knowing what `f_single` is, so it is hard (impossible?) to define a good range for `x`. 
Let me try to make an example: Take a `genome` with default params (` n_inputs=1, n_outputs=1, n_columns=128, n_rows: int = 1, primitives=Primitives((Add, Sub, Mul, ConstantFloat))`. Now, `f_single(x)=2^128` is a possible function encoded by this genome. If my thinking is correct, this would lead to a RuntimeError (numeric overflow), which is undesirable from a user perspective imo

jakobj: i've thought more about it and i now think you're right. :+1: let's go with a warning here. it would be super annoying if you're 23h42min run is canceled bc of one key that wasn't computed well.

jakobj:
ok, i've changed the exception to a warning as you requested. pls have another look :)

jakobj:
ping @HenrikMettler

test/test_utils.py
```diff
@@ -500,3 +501,62 @@ def test_compute_key_from_object():
...
     assert key == key_expected
+
+
+def test_warn_if_decorator_numpy_evalutation_all_nan():
+    @cgp.utils.disk_cache(tempfile.mkstemp()[1])
+    def inner_objective(ind):
+        return 0.0
+
+    genome_params = {
+        "n_inputs": 1,
```
HenrikMettler, line 512: Please remove the defaults `n_inputs`, `n_outputs`, `n_rows`, `levels_back`

HenrikMettler (approved):
Just one minor change in the test left. Thanks :)

PR #352, opened 2022-01-06
Restrict argument types to decorated functions

### LINKED ISSUE #325: Restrict valid (keyword) arguments for decorated functions to objects that can be reasonably converted to strings
Besides the individual (first argument) the remaining arguments of decorated functions are transformed into a string (`str(args[1:]) + str(kwargs)`) which jointly with the individual determine the key used to store return values. This can easily lead to bugs when two arguments which result in the same string have indeed different internal state. This is true, for example, for most custom classes which do not take extra care in the implementation of `__repr__`.

I suggest to *forbid* passing all types to decorated functions except (collections of)
- boolean
- strings
- ints
- floats
- arrays
- ~tensors~
- tuples

Whenever necessary, we can expand this list. I believe this will significantly improve user experience by avoiding difficult to reproduce bugs and meaningless error messages generated by the cache decorator, while remaining flexible enough to support most use cases.

jakobj:
As the title suggests: to avoid difficult-to-debug bugs this PR restricts the types which can be used as arguments to functions decorated with the caching decorator. At the moment these are restricted to: `str, int, float, tuple, list, np.ndarray, dict`. This list can be extended as necessary.

fixes #325

### DIFF
--- a/cgp/utils.py
+++ b/cgp/utils.py
@@ -64,7 +64,8 @@ def compute_key_from_sympy_expr_and_args(*args: Any, **kwargs: Any) -> str:
     ):
         raise ValueError("first argument of decorated function must be an Individual instance")
 
-    s: str = str(args[0].to_sympy()) + str(args[1:]) + str(kwargs)
+    s: str = str(args[0].to_sympy())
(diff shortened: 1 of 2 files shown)


HenrikMettler (approved):
Looks good, thanks for doing this!

PR #354, opened 2022-01-06
Use default parameters where possible in examples

### LINKED ISSUE #346: Use default parameters in examples
Many examples still define all parameters since they were created before #332 was merged. We should update the examples to only use custom settings where necessary.

jakobj:
As the title suggests. This should make all examples easier to understand.

fixes #346

### DIFF
--- a/examples/example_hurdles.py
+++ b/examples/example_hurdles.py
@@ -62,21 +62,16 @@ def objective_one(individual):
     if not individual.fitness_is_None():
         return individual
 
-    n_function_evaluations = 100000
+    n_function_evaluations = 1000000
 
     np.random.seed(1234)
-    values = np.random.uniform(-4, 4, n_function_evaluations)
+    x = np.random.uniform(-4, 4, n_function_evaluations)[:100]
 
-    f = individual.to_func()
-    loss = 0
-    for x in values[:100]:
-        # the callable returned from `to_func` accepts and returns
-        # lists; accordingly we need to pack the argument and unpack
-        # the return value
-        y = f(x)
-        loss += (f_target(x) - y) ** 2
+    f = individual.to_numpy()
+    y = f(x)
+    loss = np.mean((f_target(x) - y) ** 2)
 
-    individual.fitness = -loss / n_function_evaluations
+    individual.fitness = -loss
 
     return individual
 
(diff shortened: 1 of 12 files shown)


examples/example_hurdles.py
```diff
@@ -86,62 +81,34 @@ def objective_two(individual):
...
-ea_params = {
-    "n_offsprings": 4,
-    "mutation_rate": 0.03,
-    "n_processes": 1,
-    "hurdle_percentile": [0.5, 0.0],
-}
+# Next, we set up the evolutionary search. We first define the parameters of the
+# evolutionary algorithm. We define the upper percentile of individuals which
+# are evaluated on the (n+1)th objective by a list of numbers between 0 and 1
+ea_params = {"hurdle_percentile": [0.75, 0.0]}
```
HenrikMettler, line 102: Just out of curiosity: Why the percentile change?

examples/example_hurdles.py
```diff
@@ -193,4 +161,4 @@ def recording_callback(pop):
 ax_function.set_ylabel(r"$f(x)$")
 ax_function.set_xlabel(r"$x$")
 
-fig.savefig("example_minimal.pdf", dpi=300)
+fig.savefig("example_hurdles.pdf", dpi=300)
```
HenrikMettler, line 164: nice catch :)

examples/example_parametrized_nodes.py
```diff
@@ -96,23 +96,18 @@ def objective(individual, seed):
 
 
 # %%
-# Next, we define the parameters for the population, the genome of
+# Next, we define the parameters for the genome of
 # individuals, the evolutionary algorithm, and the local search. Note
 # that we add the custom node defined above as a primitive.
 
-population_params = {"n_parents": 1, "seed": 818821}
+seed = 818821
```
HenrikMettler, line 103: `seed = 1234`

examples/example_piecewise_target_function.py
```diff
@@ -72,55 +72,35 @@ def objective(individual, rng):
...
 # Next, we set up the evolutionary search. First, we define the parameters for
-# the population, the genomes of individuals, and the evolutionary
-# algorithm.
-population_params = {
-    "n_parents": 1,
-    "seed": 8188211,
-}
+# the genomes of individuals, and the evolutionary algorithm.
+
+seed = 818821
```
HenrikMettler, line 77: `seed = 1234`

examples/example_reorder.py
```diff
@@ -50,59 +50,37 @@ def objective(individual):
...
-    "n_offsprings": 4,
-    "mutation_rate": 0.03,
-    "n_processes": 2,
-    "reorder_genome": True,
-}
+# Next, we set up the evolutionary search. We first define the parameters for
+# the genome of individuals, and two evolutionary algorithms without (default)
+# and with genome reordering.
+
+seed = 818821
```
HenrikMettler, line 71: `seed = 1234`

HenrikMettler (changes requested):
Thanks a lot for taking care of this tedious task! I made some small observations. Also I didn't run the examples, I assume you did, to make sure they still work?

examples/example_hurdles.py
jakobj: yeah, we do like copy and paste :D

examples/example_hurdles.py
jakobj: just to make the runtime difference a bit more obvious. i thought i'd sneak it in with this PR ;)

examples/example_parametrized_nodes.py
jakobj: you mean bc it looks too hand picked? ;) should we change all seeds in the examples to 1234 then?

in general i think we can not avoid some handpicking just bc we want the examples to finish in reasonable time. a bad seed can prevent the search from finding a good solution. :(

HenrikMettler: I just copied it from above, where you changed It in an example, however if possible we can just use the same everywhere?

jakobj: yes, it seems like we can use the same everywhere. just pushed a new commit where the seeds are all identical (except a few cases where we want to use multiple different seeds in the same script).

jakobj:
> Thanks a lot for taking care of this tedious task! I made some small observations. Also I didn't run the examples, I assume you did, to make sure they still work?

thanks for the review! addressed all your comments. i think i tried all examples, but in any case, CI should catch those that don't work. :crossed_fingers:

HenrikMettler (approved):
LG! Thanks for taking care of this! 🚀

HenrikMettler:
ah, before I merge, could you squash the two commits? thx

jakobj:
thx for the review! squashed and pushed, let's see whether CI is happy.

Issue #349, opened 2022-01-05
Slow process startup time in OSX


jakobj:
The standard `ProcessPoolExecutor` seems to cause slow process startup on OSX systems. Besides runtime implications for relatively fast objectives, this causes the `test_cache_decorator` test to fail.

We should look into alternative methods to start the child processes (https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) to fix this.

PR #357, opened 2022-01-12
Use fork as explicit context for ProcessPoolExecutor in test

HenrikMettler:
The reason for the the slow process startup time in OS11 (#349) for the ProcessPoolExecutor is the use default use of spawn as a start method on macOS (see: https://docs.python.org/3/library/multiprocessing.html). Explicitly setting the context as fork removes the issue.


jakobj (approved):
looks good, thanks for taking care of this! :+1:

PR #359, opened 2022-01-13
Compare set expression for output to target expression

HenrikMettler:
When using `genome.set_expression_for_output` one usually has a target expression that the output node should compile to. Since mistakes in setting the dna can happen easily (eg. by changing the order of primitives) - a built-in check can be useful. This PR implements the comparison using `sympy.parse_expr` and a compilation to sympy of the genome.

### DIFF
--- a/cgp/genome.py
+++ b/cgp/genome.py
@@ -2,11 +2,17 @@ import re
 from typing import Dict, Generator, List, Optional, Set, Tuple, Type
 
 import numpy as np
-import sympy
 
 from .cartesian_graph import CartesianGraph
 from .node import Node, OperatorNode
 from .primitives import Primitives
+
+try:
+    import sympy
+
+    sympy_available = False
+except ModuleNotFoundError:
+    sympy_available = True
 
 try:
     import torch  # noqa: F401
--- a/cgp/genome.py
+++ b/cgp/genome.py
@@ -376,7 +376,6 @@ class Genome:
             dna segment to be inserted at the first hidden nodes.
         target_expression: str
              Expression the output node should compile to. Numbers must be written as float.
-             Defaults to None.
         hidden_start_node: int
             Index of the hidden node, where the insert starts.
             Relative to the first hidden node.
@@ -403,15 +402,14 @@ class Genome:
                 "Can not check output expression. No module named 'sympy' (extra requirement)"
             )
 
-        if target_expression is not None:
-            if self._n_outputs > 1:
-                output_as_sympy = CartesianGraph(self).to_sympy()[output_node_idx]
-            else:
-                output_as_sympy = CartesianGraph(self).to_sympy()
+        if self._n_outputs > 1:
+            output_as_sympy = CartesianGraph(self).to_sympy()[output_node_idx]
+        else:
+            output_as_sympy = CartesianGraph(self).to_sympy()
 
-            target_expression_as_sympy = sympy.parse_expr(target_expression)
-            if not output_as_sympy == target_expression_as_sympy:
-                raise ValueError("expression of output and target expression do not match")
+        target_expression_as_sympy = sympy.parse_expr(target_expression)
+        if not output_as_sympy == target_expression_as_sympy:
+            raise ValueError("expression of output and target expression do not match")
 
     def reorder(self, rng: np.random.RandomState) -> None:
         """Reorder the genome
--- a/cgp/genome.py
+++ b/cgp/genome.py
@@ -398,18 +398,18 @@ class Genome:
         try:
             import sympy
 
-            if target_expression is not None:
-                if self._n_outputs > 1:
-                    output_as_sympy = CartesianGraph(self).to_sympy()[output_node_idx]
-                else:
-                    output_as_sympy = CartesianGraph(self).to_sympy()
-
-                target_expression_as_sympy = sympy.parse_expr(target_expression)
-                if not output_as_sympy == target_expression_as_sympy:
-                    raise ValueError("expression of output and target expression do not match")
-
         except ModuleNotFoundError:
-            raise Warning("Sympy not available, can not compare written output to target")
+            raise ModuleNotFoundError("Can not check output expression. No module named 'sympy' (extra requirement)")
+
+        if target_expression is not None:
+            if self._n_outputs > 1:
+                output_as_sympy = CartesianGraph(self).to_sympy()[output_node_idx]
+            else:
+                output_as_sympy = CartesianGraph(self).to_sympy()
+
+            target_expression_as_sympy = sympy.parse_expr(target_expression)
+            if not output_as_sympy == target_expression_as_sympy:
+                raise ValueError("expression of output and target expression do not match")
 
     def reorder(self, rng: np.random.RandomState) -> None:
         """Reorder the genome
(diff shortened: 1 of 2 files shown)


cgp/genome.py
```diff
@@ -7,6 +7,13 @@
 from .node import Node, OperatorNode
 from .primitives import Primitives
 
+try:
```
jakobj, line 10: if we only need this in one function, you could delay the import such that users without sympy only get an error when they call the function that needs it.

also, i don't see where the `sympy_available` variables are used, also, i think they are set the wrong way around? `ModuleNotFoundError` means sympy is *not* available, right?

in case we don't want to rely on sympy, maybe we can misuse the `compute_key_from_numpy...` function from `utils.py` to check that the output of the graph and the desired target function are *functionally* identical?

cgp/genome.py
```diff
@@ -359,7 +366,11 @@ def change_address_gene_of_output_node(self, new_address: int, output_node_idx:
         self.dna = dna
 
     def set_expression_for_output(
-        self, dna_insert: List[int], hidden_start_node: int = 0, output_node_idx: int = 0
+        self,
+        dna_insert: List[int],
+        hidden_start_node: int = 0,
+        output_node_idx: int = 0,
+        target_expression: Optional[str] = None,
```
jakobj, line 373: i would force users to always provide a target expression for their own benefit by making this *not* optional. we all need help not shooting ourselves in the foot sometimes ;)

cgp/genome.py
```diff
@@ -388,6 +402,16 @@ def set_expression_for_output(
...
         )
+        if target_expression is not None:
+            if self._n_outputs > 1:
+                output_as_sympy = CartesianGraph(self).to_sympy()[output_node_idx]
+            else:
+                output_as_sympy = CartesianGraph(self).to_sympy()
+
+            target_expression_as_sympy = sympy.parse_expr(target_expression)
+            if not output_as_sympy == target_expression_as_sympy:
+                print(target_expression_as_sympy, output_as_sympy)
```
jakobj, line 413: i think this is a leftover from debugging, isn't it? ;)

cgp/genome.py
```diff
@@ -388,6 +402,16 @@ def set_expression_for_output(
...
+                raise ValueError("Target expression and set output expression do not match")
```
jakobj, line 414:
```suggestion
                raise ValueError("expression of output and target expression do not match")
```

test/test_genome.py
```diff
@@ -843,9 +843,25 @@ def test_set_expression_for_output(genome_params, rng):
     assert CartesianGraph(genome).to_sympy() == x_0 + x_1
 
     new_dna = [1, 0, 1]
-    genome.set_expression_for_output(new_dna)
+    genome.set_expression_for_output(dna_insert=new_dna, target_expression=" x_0 - x_1")
```
jakobj, line 846:
```suggestion
    genome.set_expression_for_output(dna_insert=new_dna, target_expression="x_0 - x_1")
```

test/test_genome.py
```diff
@@ -843,9 +843,25 @@ def test_set_expression_for_output(genome_params, rng):
...
     assert CartesianGraph(genome).to_sympy() == x_0 - x_1
 
     new_dna = [0, 0, 1, 2, 0, 0, 1, 0, 0, 0, 2, 3]  # x_0+x_1; 1.0; 0; x_0+x_1 + 1.0
-    genome.set_expression_for_output(new_dna)
+    genome.set_expression_for_output(dna_insert=new_dna, target_expression=" x_0 + x_1 + 1.0")
```
jakobj, line 850:
```suggestion
    genome.set_expression_for_output(dna_insert=new_dna, target_expression="x_0 + x_1 + 1.0")
```

test/test_genome.py
```diff
@@ -843,9 +843,25 @@ def test_set_expression_for_output(genome_params, rng):
...
     assert CartesianGraph(genome).to_sympy() == x_0 + x_1 + 1.0
+
+    with pytest.raises(ValueError):
+        genome.set_expression_for_output(dna_insert=new_dna, target_expression=" x_0 + x_1 + 1")
```
jakobj, line 854: hmm this test needs a comment, otherwise it's pretty surprising that it's expected to fail.

in addition i would suggest to add an expected-failure test where the expression are indeed different, for example by replacing `+` with `8`.

test/test_genome.py
```diff
@@ -843,9 +843,25 @@ def test_set_expression_for_output(genome_params, rng):
...
+    genome2_params = {
+        "n_inputs": 2,
+        "n_outputs": 2,
+        "primitives": (cgp.Add, cgp.Sub, cgp.ConstantFloat),
+    }
+    genome2 = cgp.Genome(**genome2_params)
+    genome2.randomize(rng)
+
+    genome2.set_expression_for_output(
+        new_dna, output_node_idx=1, target_expression=" x_0 + x_1 + 1.0"
```
jakobj, line 865:
```suggestion
        new_dna, output_node_idx=1, target_expression="x_0 + x_1 + 1.0"
```

jakobj (changes requested):
great, thanks for working on this! this will prevent much pain. :eyes: 

i've left a few comments that should be addressed before merging.

cgp/genome.py
```diff
@@ -388,6 +395,21 @@ def set_expression_for_output(
...
+                    output_as_sympy = CartesianGraph(self).to_sympy()[output_node_idx]
+                else:
+                    output_as_sympy = CartesianGraph(self).to_sympy()
+
+                target_expression_as_sympy = sympy.parse_expr(target_expression)
+                if not output_as_sympy == target_expression_as_sympy:
+                    raise ValueError("expression of output and target expression do not match")
+
+        except ModuleNotFoundError:
+            raise Warning("Sympy not available, can not compare written output to target")
```
jakobj, line 412:
```suggestion
            raise ModuleNotFoundError("No module named 'sympy' (extra requirement)")
```
as mentioned above, i would raise an exception with our standard error message. you can possibly extend this with more detailed info, e.g., "can not check output expression. No module..."

jakobj (changes requested):
almost done :rocket: left two more comments

cgp/genome.py
```diff
@@ -370,6 +374,9 @@ def set_expression_for_output(
         ----------
         dna_insert: List[int]
             dna segment to be inserted at the first hidden nodes.
+        target_expression: str
+             Expression the output node should compile to. Numbers must be written as float.
+             Defaults to None.
```
jakobj, line 379:
```suggestion
```

cgp/genome.py
```diff
@@ -388,6 +395,23 @@ def set_expression_for_output(
...
         )
+        try:
+            import sympy
+
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError(
+                "Can not check output expression. No module named 'sympy' (extra requirement)"
+            )
+
+        if target_expression is not None:
```
jakobj, line 406:
```suggestion
```

jakobj (changes requested):
great, just some leftover to cleanup from when `target_expression` was optional. :+1:

afterwards, please squash commits.

jakobj:
ping @HenrikMettler

mschmidt87:
ping @HenrikMettler

jakobj:
i think i don't have any more pressing comments, besides squashing commits please and fixing tests :)

mschmidt87:
@HenrikMettler could you fix the tests, rebase on master and squash the commits?

PR #360, opened 2022-01-13
Set reorder=True per default. 

### LINKED ISSUE #347: Always use "reorder=True" and remove possiblity to configure this?
Is there a downside to reordering the genome? If not I think we should remove the possibility of choosing this setting and just always doing it. We need to keep an eye on the number of configurable parameters since with every parameter the learning curve becomes steeper and the probability of choosing "wrong" settings increases.

What do you think @HenrikMettler?

HenrikMettler:
Moves the boolean `reorder_genome` into population to check whether genome conditions (n_rows=1, levels_back=n_columns) for `reorder=True` are met. 
I am not removing the possibility to configure this, since `reorder` is limited to certain genome parameter configurations. 
Closes #347.

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -26,7 +26,6 @@ class MuPlusLambda:
         n_processes: int = 1,
         local_search: Optional[Callable[[IndividualBase], None]] = None,
         k_local_search: Optional[int] = None,
-        reorder_genome: bool = False,
         hurdle_percentile: List = [0.0],
     ):
         """Init function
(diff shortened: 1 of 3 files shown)


jakobj:
good catch that we can not just remove the configuration option. how do you feel about removing the possibility of setting `n_rows` and always use `n_rows=1`? my impression is that modern general purpose CGP anyways uses exclusively only one row but may columns, while multiple rows were historically used for evolving circuits.

HenrikMettler:
yeah, i think I agree. Should we also remove `levels_back`? Atm I can't see any use case where anything but `levels_back=n_columns` is useful.

jakobj:
yes, i think you're right, `levels_back` can also be removed. :+1:

could you please create a separate PR, removing `n_rows` and `levels_back`? after that has been merged, we can come back to this one.

Issue #364, opened 2022-05-25
CI should make sure all examples are working properly


jakobj:
Currently, we merely make sure that the examples don't crash, not that they lead to the correct results. We should add strategic `assert`s, warning us if an example does not find good solutions any more.

PR #366, opened 2022-05-25
Evaluate objective in C

HenrikMettler:
For some applications, one might want to evaluate fitness in C. Here `CartesianGraph` is extended with a function `to_c` which writes the computational graph into a C-module. An example is provided which performs L2 regression on a simple target function. 
_Note_: Due to a late api change (writing files instead of returning strings) the test is currently not working - I left it as commented out code since large junks of it should be reusable with reading from generated files

### DIFF
--- a/cgp/cartesian_graph.py
+++ b/cgp/cartesian_graph.py
@@ -1,6 +1,7 @@
 import collections
 import copy
 import math  # noqa: F401
+import os
 import re
 from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Set
 
@@ -12,6 +13,7 @@ from .node_input_output import InputNode, OutputNode
 try:
     import sympy
     from sympy.core import expr as sympy_expr  # noqa: F401
+    from sympy.utilities.codegen import codegen
 
     sympy_available = True
 except ModuleNotFoundError:
@@ -435,3 +437,35 @@ class _C(torch.nn.Module):
             return sympy_exprs[0]
         else:
             return sympy_exprs
+
+    def to_cpp(self, function_name, filename, path):
+        """Create a C++ module described by the graph.
+
+        Returns a C++ module. Currently only available for a single output node.
+        See: https://docs.sympy.org/latest/modules/utilities/codegen.html
+
+        Returns
+        ----------
+        C++ module
+
+        """
+
+        if not sympy_available:
+            raise ModuleNotFoundError("No sympy module available. Required for exporting C++ module")
+
+        if not self._n_outputs == 1:
+            raise ValueError("C++ module export only available for single output node.")
+
+        sympy_expr = self.to_sympy()
+
+        # from sympy.abc import x, y, z
+        # [(c_name, c_code), (h_name, c_header)] = codegen(("f", x + y * z), "C89", "test", header=False, empty=False)
+        [(filename_cpp, code_cpp), (filename_header, code_header)] = codegen((function_name, sympy_expr),
+                                                         "C99", filename, header=False, empty=False)
+
+        if not os.path.exists(path):
+            os.makedirs(path)
+        with open("%s/%s"%(path, filename_cpp), 'w') as f:
+            f.write(f'{code_cpp}')
+        with open("%s/%s"%(path, filename_header), 'w') as f:
+            f.write(f'{code_header}')
--- a/cgp/cartesian_graph.py
+++ b/cgp/cartesian_graph.py
@@ -438,36 +438,56 @@ class _C(torch.nn.Module):
         else:
             return sympy_exprs
 
-    def to_cpp(self, function_name, filename, path):
-        """Create a C++ module described by the graph.
+    def to_c(self, function_name, filename, path):
+        """Create a C module described by the graph.
 
-        Returns a C++ module. Currently only available for a single output node.
-        See: https://docs.sympy.org/latest/modules/utilities/codegen.html
+        Writes code and header into files in the given path.
+        Important: function_name and filename have to be different, due to
+        Currently only available for a single output node.
 
         Returns
         ----------
-        C++ module
-
+        None
         """
 
         if not sympy_available:
-            raise ModuleNotFoundError("No sympy module available. Required for exporting C++ module")
+            raise ModuleNotFoundError("No sympy module available. Required for exporting C module")
 
         if not self._n_outputs == 1:
-            raise ValueError("C++ module export only available for single output node.")
+            raise ValueError("C module export only available for single output node.")
+
+        if function_name in filename:
+            raise ValueError("function_name can not be substring of filename, due to function declaration"
+                             "consistency checks")
 
         sympy_expr = self.to_sympy()
 
-        # from sympy.abc import x, y, z
-        # [(c_name, c_code), (h_name, c_header)] = codegen(("f", x + y * z), "C89", "test", header=False, empty=False)
-        [(filename_cpp, code_cpp), (filename_header, code_header)] = codegen((function_name, sympy_expr),
+        [(filename_c, code_c), (filename_header, code_header)] = codegen((function_name, sympy_expr),
                                                          "C99", filename, header=False, empty=False)
+
+        def replace_func_declaration_in_code_and_header_with_full_variable_set(code_c, code_header, function_name):
+
+            arg_string_list = [f'double x_{idx}' for idx in range(self._n_inputs)]
+            permanent_header = f'{function_name}(' + ", ".join(arg_string_list) + ')'
+
+            c_replace_start_idx = code_c.find(function_name)
+            c_replace_end_idx = code_c.find(')', c_replace_start_idx) + 1  # +1 offset for
+            code_c = code_c.replace(code_c[c_replace_start_idx:c_replace_end_idx], permanent_header)
+
+            h_replace_start_idx = code_header.find(function_name)
+            h_replace_end_idx = code_header.find(')', h_replace_start_idx) + 1
+            code_header = code_header.replace(code_header[h_replace_start_idx:h_replace_end_idx], permanent_header)
+
+            return code_c, code_header
+
+        # assert function declaration consistency - replace declaration in header and code
+        code_c, code_header = replace_func_declaration_in_code_and_header_with_full_variable_set(code_c, code_header,
+                                                                                                 function_name)
 
         if not os.path.exists(path):
             os.makedirs(path)
-        with open("%s/%s"%(path, filename_cpp), 'w') as f:
-            f.write(f'{code_cpp}')
+        with open("%s/%s"%(path, filename_c), 'w') as f:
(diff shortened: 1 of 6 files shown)


examples/c_code/main.c
```diff
@@ -0,0 +1,50 @@
+#include "individual.h"
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+
+double target(double x_0, double x_1) {
+    double target;
+    target = x_0 * x_1 + 1.0;
+    return target;
```
jakobj, line 10:
```suggestion
    return x_0 * x_1 + 1.0;
```
i guess the compiler does this anyway, but it's a bit faster to read as well ;)

examples/c_code/main.c
```diff
@@ -0,0 +1,50 @@
...
+/* generate a random floating point number from min to max */
+double rand_from(double min, double max) 
+{
+    double range = (max - min); 
+    double div = RAND_MAX / range;
+    return min + (rand() / div);
+}
+
+
+double l2_norm_rule_target() {
```
jakobj, line 22:
```suggestion
double loss {
```

examples/c_code/main.c
```diff
@@ -0,0 +1,50 @@
...
+
+
+double target(double x_0, double x_1) {
+    double target;
+    target = x_0 * x_1 + 1.0;
+    return target;
+}
+
+/* generate a random floating point number from min to max */
+double rand_from(double min, double max)
```
jakobj, line 14:
```suggestion
double rand_from_to(double min, double max) 
```

examples/c_code/main.c
```diff
@@ -0,0 +1,50 @@
...
+double rand_from(double min, double max) 
+{
+    double range = (max - min); 
+    double div = RAND_MAX / range;
+    return min + (rand() / div);
+}
+
+
+double l2_norm_rule_target() {
+    int sz = 100;
```
jakobj, line 23: maybe a bit more obvious variable name, what about "n_samples"?

examples/c_code/main.c
```diff
@@ -0,0 +1,50 @@
+#include "individual.h"
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+
+double target(double x_0, double x_1) {
```
jakobj, line 7:
```suggestion
double target(const double x_0, const double x_1) {
```
i think it's good practice to try to make things reasonably `const` ;)

examples/c_code/main.c
```diff
@@ -0,0 +1,50 @@
...
+    double target_value;
+    double rule_output;
+    double sum_l2_difference = 0.0;
+
+    for(int i=0;i<sz;i++){
+        /* generate two random values for x_0, x_1 */
+        double min = -1.0;
+        double max = 1.0;
+        x_0_rand=rand_from(min, max);
+        x_1_rand=rand_from(min, max);
```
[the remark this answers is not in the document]
jakobj: you can just declare them here

examples/c_code/main.h
```diff
@@ -0,0 +1,3 @@
+double target(double x_0, double x_1);
```
jakobj, line 1: i think you don't need a header for "standalone" files like main

examples/example_evaluate_in_c.py
```diff
@@ -0,0 +1,107 @@
+"""
+Example for evolutionary regression, with evaluation in c
```
jakobj, line 2:
```suggestion
Example for evolutionary regression, with evaluation in C
```

jakobj: etc

examples/example_evaluate_in_c.py
```diff
@@ -0,0 +1,107 @@
...
+    path_script_h = pathlib.Path(f"{path}/{scriptname}.h")
+    assert (
+        path_file_c.is_file()
+        & path_file_h.is_file()
+        & path_script_c.is_file()
+        & path_script_h.is_file()
+    )
+
+    # compile file with rule
+    subprocess.run(["gcc", "-c", "-fPIC", f"{path}/{filename}.c", "-o", f"{path}/{filename}.o"])
```
jakobj, line 44:
```suggestion
    subprocess.check_call(["gcc", "-c", "-fPIC", f"{path}/{filename}.c", "-o", f"{path}/{filename}.o"])
```
this makes noise when the compilation fails. same for the other calls

examples/example_evaluate_in_c.py
```diff
@@ -0,0 +1,107 @@
...
+    # compile script
```
jakobj, line 45: i think you only need to compile main once not everytime after changing the individual bc main remains the same, i.e., the object file doesn't change. but it can't hurt if you double check this advice ;)

examples/example_evaluate_in_c.py
```diff
@@ -0,0 +1,107 @@
...
+    subprocess.run(
+        ["gcc", f"{path}/{scriptname}.o", f"{path}/{filename}.o", "-o", f"{path}/{filename}"]
+    )
+
+
+# %%
+# We define the objective function for the evolution. It creates a
+# c module and header from the computational graph. File with rule
+# and script for evaluation are compiled using the above helper function.
+# It assigns fitness to the negative float of the print of the script execution.
```
jakobj, line 59:
```suggestion
# We define the objective function for the evolution. It creates a
# C module from the computational graph. The module
# and the main source file for evaluation are compiled using the above helper function.
# Here the objective obtains the fitness by reading the screen output of the C program.
```

examples/example_evaluate_in_c.py
```diff
@@ -0,0 +1,107 @@
...
+
+def objective(individual):
+
+    if not individual.fitness_is_None():
+        return individual
+
+    graph = cgp.CartesianGraph(individual.genome)
+    function_name = "rule"
+    filename = "individual"
+    scriptname = "main"
```
jakobj, line 70: does this need to be adjustable by the user? can't we force them to stick to "main"? at least until someone complains about it? ;)

examples/example_evaluate_in_c.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # compile file with rule
+    subprocess.run(["gcc", "-c", "-fPIC", f"{path}/{filename}.c", "-o", f"{path}/{filename}.o"])
+    # compile script
+    subprocess.run(
+        ["gcc", "-c", "-fPIC", f"{path}/{scriptname}.c", "-o", f"{path}/{scriptname}.o"]
+    )
+    # create executable
+    subprocess.run(
+        ["gcc", f"{path}/{scriptname}.o", f"{path}/{filename}.o", "-o", f"{path}/{filename}"]
```
jakobj, line 51:
```suggestion
        ["gcc", f"{path}/{scriptname}.o", f"{path}/{filename}.o", "-o", f"{path}/{scriptname}"]
```
i would find this naming more intuitive

examples/example_evaluate_in_c.py
```diff
@@ -0,0 +1,107 @@
...
+    graph.to_c(function_name=function_name, filename=filename, path=path)
+
+    # compile_c_code()
+    compile_c_code(filename=filename, scriptname=scriptname, path=path)
+
+    # assert that the executable returns something
+    assert subprocess.check_output(pathlib.Path().absolute() / f"{path}/{filename}")
+    # run simulation and assign fitness
+    individual.fitness = -1.0 * float(
+        subprocess.check_output(pathlib.Path().absolute() / f"{path}/{filename}")
```
jakobj, line 82:
```suggestion
    result = subprocess.check_output(pathlib.Path().absolute() / f"{path}/{filename}")
    assert result
    individual.fitness = -1.0 * float(result)
```
no need to execute the program twice ;)

examples/example_evaluate_in_c.py
```diff
@@ -0,0 +1,107 @@
...
+# %%
+# and finally perform the evolution relying on the libraries default
+# hyperparameters except that we terminate the evolution as soon as one
+# individual has reached fitness zero.
+
+pop = cgp.evolve(objective=objective, pop=pop, termination_fitness=0.0, print_progress=True)
+
+# %%
+# After finishing the evolution, we print the final evolved expression.
+print(pop.champion.to_sympy())
```
jakobj, line 107: we could try working towards our new goal (https://github.com/Happy-Algorithms-League/hal-cgp/issues/364) by adding `assert(pop.champion.to_sympy() == "x_0*x_1 + 1.0")` :)

jakobj (changes requested):
awesome, great work @HenrikMettler!!1 :rocket: 

i left a couple of comments that should be addressed before merging.

cgp/cartesian_graph.py
```diff
@@ -435,3 +437,66 @@ def to_sympy(self, simplify: Optional[bool] = True):
             return sympy_exprs[0]
         else:
             return sympy_exprs
+
+    def to_c(self, function_name, filename, path):
```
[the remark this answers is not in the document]
HenrikMettler: Not really, should I just set it to `rule`?

cgp/cartesian_graph.py
```diff
@@ -435,3 +437,66 @@ def to_sympy(self, simplify: Optional[bool] = True):
...
+        if not sympy_available:
+            raise ModuleNotFoundError("No sympy module available. Required for exporting C module")
+
+        if not self._n_outputs == 1:
+            raise ValueError("C module export only available for single output node.")
+
+        if function_name in filename:
+            raise ValueError(
+                "function_name can not be substring of filename, due to function declaration"
+                "consistency checks"
```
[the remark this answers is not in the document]
HenrikMettler: Looking at this again, maybe it is easiest to just remove the ability to set `function_name` and `filename`? Then we don't have to worry about this check?

cgp/cartesian_graph.py
```diff
@@ -435,3 +437,66 @@ def to_sympy(self, simplify: Optional[bool] = True):
...
+
+            h_replace_start_idx = code_header.find(function_name)
+            h_replace_end_idx = code_header.find(")", h_replace_start_idx) + 1
+            code_header = code_header.replace(
+                code_header[h_replace_start_idx:h_replace_end_idx], permanent_header
+            )
+
+            return code_c, code_header
+
+        # assert function declaration consistency - replace declaration in header and code
```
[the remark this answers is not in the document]
HenrikMettler: I tried to add a description to make things more clear in the docstring of `replace_func_signature_in_source_and_header_with_full_variable_set`

Maybe that helps a bit - the whole function is a bit messy, but I wasn't sure how to make it more readable

cgp/cartesian_graph.py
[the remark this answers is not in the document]
jakobj: it's not always a learning rule, is it? so how about some generic function name, like `f`? i know this one letter function names should be avoided, but here we may have an exception ;)

PR #368, opened 2022-06-22
Refine github workflow and freeze some packages

### LINKED ISSUE #363: Run code checks in workflow separately
Currently, it's not easy to see which code check actually failed:
2022-03-09-111549_1441x434_scrot
(here, flake8 failed, not black!)

We should separate code checks into separate steps to make it obvious, for example, whether black or flake8 failed.

jakobj:
As the title suggests.

closes #363

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -20,34 +20,41 @@ jobs:
           - python-version: 3.8
             dep: "[dev]"
     steps:
-      - name: checkout
+      - name: Checkout repository
         uses: actions/checkout@v3
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
+      - name: Install pip
+        run: |
+          python -m pip install --upgrade pip
+      - name: Install library and dependencies
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -20,36 +20,53 @@ jobs:
...
           isort --check-only cgp examples test
-      - name: run tests
+      - name: Run tests
         run: |
+          pytest --version
           pytest --cov=cgp
-      - name: Build Documentation
+      - name: Build documentation
         run: |
-          pip install gym || exit 1;
```
mschmidt87, line 47: Is this line not necessary anymore? It seems that you deleted it altogether?

mschmidt87: Ah nevermind, `gym` is already installed via the extra requirements.

mschmidt87 (commented):
Is there a particular reason to freeze those packages? Or just a general policy that it's better to freeze instead of being surprise by new releases?

jakobj:
> Is there a particular reason to freeze those packages? Or just a general policy that it's better to freeze instead of being surprise by new releases?

yep, it's necessary to make everything play well together. it's a consequence of having chosen to freeze some packages :( but see #369 for the future of this!

mschmidt87 (approved):

PR #370, opened 2022-06-23
Don't create objects via default function parameters in hl api

### LINKED ISSUE #353: Default `Population` and `MuPlusLambda` objects in `evolve` are persistent
The current `evolve` signature
```python
def evolve(
    ...
    pop: Population = Population(),
    ea: MuPlusLambda = MuPlusLambda(),
    ...
) -> Population:
```
creates default `Population` and `MuPlusLambda` objects if none are passed to the function. Unfortunately, these will be persistent, i.e., subsequent calls to `evolve` seem to reuse the same objects.

To fix this the default argument should be `None` and they should initialized to the defaults in the body of the function if `None` is detected.

jakobj:
When objects are created this way they will be persistent across multiple invocations of the associated function running in the same interpreter session. This can have annoying side effects.

branches from #368 so should be merged after

closes #353

### DIFF
--- a/cgp/hl_api.py
+++ b/cgp/hl_api.py
@@ -9,8 +9,8 @@ from .population import Population
 
 def evolve(
     objective: Callable[[IndividualBase], IndividualBase],
-    pop: Population = Population(),
-    ea: MuPlusLambda = MuPlusLambda(),
+    pop: Population = None,
+    ea: MuPlusLambda = None,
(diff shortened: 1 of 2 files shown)


mschmidt87 (approved):
This change makes a lot of sense, thanks.

mschmidt87:
Is there any reason why this PR is based on #368 ? It looks unrelated to me.

PR #374, opened 2022-11-21
Random local search over a set of permissible values

### LINKED ISSUE #373: Local search randomly over a set of values
Currently our library offers two ways of searching values for parameterised nodes: `Evolution Strategies` and `Gradient based`. Both search values "locally" around the current values. In some use cases one might not want to search locally, but, e.g. over values of different magnitude. One possible way to do this is by randomly sampling possible values for each parameter and updating parameters when fitness increases. Such a search should be added to the local search implementations.

HenrikMettler:
Addresses #373. This is based on an implementation I used in a project with hal-cgp, thus the default values for `permissible-values`. There is two open questions, marked as todo in the code comments. 

For convenience i put a short example at the end. I can extend this to a full example before merging or remove it.

### DIFF
--- a/cgp/local_search/random_search.py
+++ b/cgp/local_search/random_search.py
@@ -1,0 +1,68 @@
+import numpy as np
+from typing import Callable
+
+
+class RandomLocalSearch:
+    def __init__(
+        self,
+        objective: Callable,
+        seed: int,
+        n_steps: int = 10,  # todo: should this be dependent on the number of parameters?
+        permissible_values: np.ndarray = np.logspace(start=0, stop=7, num=8, base=2, dtype=int),
+    ) -> None:
+        self.objective = objective
+        self.seed = seed
+        self.n_steps = n_steps
+        self.permissible_values = permissible_values
+
+    def __call__(self, ind) -> None:
+        rng = np.random.RandomState(self.seed)
+
+        params_values, params_names = ind.parameters_to_numpy_array(only_active_nodes=True)
+
+        if len(params_values) > 0:
+            for _ in range(self.n_steps):
+                # sample a new set of parameter values randomly
+                params_sampled = [
+                    rng.choice(self.permissible_values) for param_value in params_values
+                ]
+                # write the parameters into a clone of individual
+                new_ind = ind.clone()
+                new_ind.update_parameters_from_numpy_array(
+                    params=params_sampled, params_names=params_names
+                )
+                # evaluate fitness
+                self.objective(new_ind)
+                # if fitness improved: replace parameter values and fitness
+                if new_ind.fitness >= ind.fitness:  # todo: should this be >= or > ??
+                    ind.update_parameters_from_numpy_array(
+                        params=params_sampled, params_names=params_names
+                    )
+                    ind.fitness = new_ind.fitness
+
+
+if __name__ == "__main__":
+    import cgp
+
+    def objective(ind):
+        params_values, _ = ind.parameters_to_numpy_array(only_active_nodes=True)
+        ind.fitness = np.sum(params_values)
+        return ind
+
+    seed = 12345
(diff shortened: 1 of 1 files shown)


HenrikMettler:
(CI-check fail seems like an unrelated mypy issue)

cgp/local_search/random_search.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import Callable
+
+
+class RandomLocalSearch:
+    def __init__(
+        self,
+        objective: Callable,
+        seed: int,
+        n_steps: int = 10,  # todo: should this be dependent on the number of parameters?
+        permissible_values: np.ndarray = np.logspace(start=0, stop=7, num=8, base=2, dtype=int),
```
jakobj, line 11: in the current implementation, is it, for example, possible to say "parameters should be uniformly sampled in the interval [0, 1]"? if not, I think it would be a good idea to make the generation of permissible values modular to support such use cases

cgp/local_search/random_search.py
```diff
@@ -0,0 +1,68 @@
+import numpy as np
+from typing import Callable
+
+
+class RandomLocalSearch:
+    def __init__(
+        self,
+        objective: Callable,
+        seed: int,
+        n_steps: int = 10,  # todo: should this be dependent on the number of parameters?
```
jakobj, line 10: addressing your todo: no i think a good, fixed default is best

cgp/local_search/random_search.py
```diff
@@ -0,0 +1,68 @@
...
+        permissible_values: np.ndarray = np.logspace(start=0, stop=7, num=8, base=2, dtype=int),
+    ) -> None:
+        self.objective = objective
+        self.seed = seed
+        self.n_steps = n_steps
+        self.permissible_values = permissible_values
+
+    def __call__(self, ind) -> None:
+        rng = np.random.RandomState(self.seed)
```
jakobj, line 19: this can be moved into the `if len(` block, so that it's not unnecessarily executed for individuals without parameters

cgp/local_search/random_search.py
```diff
@@ -0,0 +1,68 @@
...
+                ]
+                # write the parameters into a clone of individual
+                new_ind = ind.clone()
+                new_ind.update_parameters_from_numpy_array(
+                    params=params_sampled, params_names=params_names
+                )
+                # evaluate fitness
+                self.objective(new_ind)
+                # if fitness improved: replace parameter values and fitness
+                if new_ind.fitness >= ind.fitness:  # todo: should this be >= or > ??
```
jakobj, line 37: what would be the benefit of `>`, what would be the benefit of `>=`?

cgp/local_search/random_search.py
```diff
@@ -0,0 +1,68 @@
...
+                    rng.choice(self.permissible_values) for param_value in params_values
+                ]
+                # write the parameters into a clone of individual
+                new_ind = ind.clone()
+                new_ind.update_parameters_from_numpy_array(
+                    params=params_sampled, params_names=params_names
+                )
+                # evaluate fitness
+                self.objective(new_ind)
+                # if fitness improved: replace parameter values and fitness
```
jakobj, line 36: no need for this comment ;) indeed, also not for the others, I think the code is clear enough.

cgp/local_search/random_search.py
```diff
@@ -0,0 +1,68 @@
...
+                if new_ind.fitness >= ind.fitness:  # todo: should this be >= or > ??
+                    ind.update_parameters_from_numpy_array(
+                        params=params_sampled, params_names=params_names
+                    )
+                    ind.fitness = new_ind.fitness
+
+
+if __name__ == "__main__":
```
jakobj, line 44: the rest should be moved into a test I guess ;)

jakobj (changes requested):
great stuff, thanks for working on this! :rocket: 

I left a few comments, please address them before we merge this

PR #376, opened 2022-12-06
Make sure all individuals (parents and offspring) are assigned a valid, unique index

### LINKED ISSUE #375: Assert runtime invariant: all individuals in `MuPlusLambda::step` need to have a non-None index
The logic in `MuPlusLambda` (and likely in other places) assumes that all individuals in a population have been assigned a valid index. We should thus make sure of this in the beginning of this function using an `assert`.

jakobj:
(as the title suggests)

closes #375

### DIFF
--- a/cgp/ea/mu_plus_lambda.py
+++ b/cgp/ea/mu_plus_lambda.py
@@ -142,6 +142,8 @@ class MuPlusLambda:
         # concatenating the parent population to the offspring
         # population instead of the other way around
         combined = offsprings + pop.parents
+        assert all(ind.idx is not None for ind in combined)
+        assert len(set(ind.idx for ind in combined)) == len(combined)
 
         # we follow a two-step process for selection of new parents:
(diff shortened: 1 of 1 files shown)


cgp/ea/mu_plus_lambda.py
```diff
@@ -142,6 +142,8 @@ def step(
         # concatenating the parent population to the offspring
         # population instead of the other way around
         combined = offsprings + pop.parents
+        assert all(ind.idx is not None for ind in combined)
+        assert len(set(ind.idx for ind in combined)) == len(combined)
```
HenrikMettler, line 146: Maybe below could be easier to understand? (No preference though, both ways should be fine)
```python
assert len(set(ind.idx for ind in combined)) == len([int.idx for ind in combined])
```

HenrikMettler (approved):
Thanks for taking care of this :) Lgtm, a suggestion but maybe that makes the code less readable

PR #391, opened 2026-07-23
feat: Remove pillow from extra requirements to avoid CVE-2026-59200

mschmidt87:


jakobj:
i'm a bit confused. a) does this affect us in any way? i.e., are we reading pdfs from some unknown source? b) why is pillow in our extra requirements to begin with if we could just remove it? sry it's been a while since i thought about this code :D

ps: to be clear, if we don't need it, then of course happy to remove!

